PR #14, opened 2018-02-24
Upgrade to latest generic-array

garious:


newpavlov:
Welcome!

`GenericArray` type is a part of the public API, so update to `generic-array v0.10` must be done with the minor version bump. Semver hack could've probably helped, but it's currently [not used](https://github.com/fizyk20/generic-array/issues) by `generic-array`. `digest v0.8` is not yet published, so you can remove patch version bump, as for other crates let's wait for generic-array to apply semver hack.

garious:
Thanks @newpavlov. Since the latest generic-array doesn't seem to affect any existing functionality in the public API here, I see why you'd want to use the semver hack. Looks like the generic-array author is receptive to the idea too. If that goes in, feel free to close this pull request.

PR #15, opened 2018-03-17
Bump generic-array crate

niklasad1:


garious:
Latest is now 0.11.1. Any chance for an upgrade?

newpavlov:
It will require a coordinated update ideally for all crates, as `GenericArray` is used as a part of public API. It will require some work.

In `digest` case I want to make update `generic-array` together with `digest v0.8`, work on which stalled a bit.

garious:
Thanks @newpavlov. Would you mind expanding a bit on what "all crates" includes and if work stalled because folks are headed another direction?

newpavlov:
Sorry for the late answer! Bumping `generic-array` is a breaking change, as we use `GenericArray` as part of public API. Thus all downstream crates (hashes, block ciphers, MACs, etc.) will have to be updated as well. All in all it's shouldn't be that difficult because crates usually use re-exported `generic-arrays`.

But because it's a breaking change and ideally should be done in a coordinated way (because there is some inter-dependencies between e.g. hashes/MACs, MACs/block ciphers), I've planned to do it as part of `digest` update, which unfortunately is stalled a bit.

newpavlov:
I will close it in favor of #22, while the main focus of the PR is `digest` crate, it also updates other trait crates to `generic-array v0.12`.

PR #37, opened 2019-04-29
upgrade `generic-array` to 0.13

flier:


newpavlov:
Thank you for your effort! But `generic-array` v0.13 uses Rust 2018, so to keep with the [MSRV promise](https://github.com/RustCrypto/traits#minimum-rust-version) we can't do this upgrade without bumping minor versions. I plan to do it eventually, but not right now. So I will close this PR for a time being.

PR #40, opened 2019-07-26
AEAD Trait

jcape:
Hi,

I've previously spoken with @newpavlov via e-mail, and he asked me to create this PR and made some cursory comments on the original, internal-to-MobileCoin, source code for this crate. I've attempted to address the comments and clarify the Nonce object's purpose. My goal here is to get this interface into widespread usage, so I'm amenable to changing whatever needs to be changed to make that a reality.

The crate itself is intended to provide a misuse-resistant, rust-ish way to use RFC5116 AEAD ciphers.

PR #48, opened 2019-08-28
aead: Add `Payload` type

tarcieri:
This implements a `Payload` type suggested by @newpavlov here:

https://github.com/RustCrypto/traits/pull/40#issuecomment-522697851

It uses an `impl From<Payload<'msg, 'aad>>` for all plaintexts and ciphertexts, with the idea that users of the API who are unconcerned with AAD can just pass the plaintext/ciphertext message they intend to encrypt or decrypt.

With a `plaintext: &[u8]`:

```rust
let ciphertext = cipher.encrypt(nonce, plaintext);
let plaintext = cipher.encrypt(nonce, ciphertext).unwrap();
```

Or if you do want to pass AAD:

```rust
let ciphertext = cipher.encrypt(nonce, Payload { msg: plaintext, aad });
let plaintext = cipher.decrypt(nonce, Payload { msg: ciphertext, aad }).unwrap();
```

This makes `msg` and `aad` explicit when used so users don't accidentally pass their intended plaintexts as `aad` and accidentally expose them because AAD is unencrypted.

The `Into<Payload<'_ '_>>` impl on `&[u8]` automatically uses `b""` (i.e. empty byte slice) as the AAD when it's coerced, which is what I think 99% of users will want.

As someone who's concerned about AAD for protocols like Noise (and also other use cases like binding digital signatur keys to ciphertexts), I think for the overwheliming majority of users it's a confusing, superfluous detail. I think this approach neatly leverages Rust's polymorphism to hide this detail in cases where it doesn't matter.

### DIFF
--- a/aead/src/lib.rs
+++ b/aead/src/lib.rs
@@ -7,7 +7,7 @@ extern crate alloc;
 pub use generic_array;
 
 use alloc::vec::Vec;
-use generic_array::{GenericArray, ArrayLength, typenum::Unsigned};
+use generic_array::{typenum::Unsigned, ArrayLength, GenericArray};
 
 #[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
 pub struct Error;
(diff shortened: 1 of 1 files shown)


newpavlov:
Neat! I really like it how it preserves ergonomics for cases without AD.

tarcieri:
If there aren't any objections, can we merge this? (I don't have access to this repo)

newpavlov:
Ah, yes. I will do v0.1.0 release right away.

tarcieri:
@newpavlov I had one more PR I wanted to do first! 😉

newpavlov:
Ok. :) I will do a small commit now with minor edits of Cargo.toml.

XinGOfCloude18 (approved):

XinGOfCloude18 (changes requested):
Plz

PR #49, opened 2019-08-29
aead: Rename StatelessAead -> Aead; Aead -> AeadMut (closes #46)

### LINKED ISSUE #46: aead: Should we have StatefulAead rather than StatelessAead?
Right now we have a stateful `Aead` trait and a `StatelessAead` trait.

It seems like the common case for anything based on a stream cipher (pretty much all AEADs) is `StatelessAead`, and that `StatelessAead` is more useful in that it's `Send` + `Sync`.

Should we rename `StatelessAead` to `Aead`, and rename the current `Aead` trait to `StatefulAead` instead? I think this follows the general Rust convention of giving the immutable version of something the shorter, more convenient name.

tarcieri:
As it seems like most commonly used AEAD algorithms can be implemented as what was previously called `StatelessAead`, this commit renames it to give it the shorter, more convenient name (`Aead`).

Additionally, it renames the prior `Aead` trait to `AeadMut`, as suggested by @burdges in #46.

This seems to follow the general Rust convention of making the immutable version of something shorter and easier to use, and explicitly calling out `Mut` where it exists.

### DIFF
--- a/aead/src/lib.rs
+++ b/aead/src/lib.rs
@@ -12,8 +12,7 @@ use generic_array::{typenum::Unsigned, ArrayLength, GenericArray};
 #[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
 pub struct Error;
 
-/// A trait which can be used to create a new RFC5116 authenticated encryption
-/// scheme.
+/// Instantiate either a stateless (`Aead`) or stateful (`AeadMut`) algorithm.
 pub trait NewAead {
     /// The size of the key array required by this algorithm.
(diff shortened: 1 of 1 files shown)


tarcieri:
@newpavlov alright, this is the last thing I wanted to do before a release 😉

newpavlov:
Looks good!

newpavlov:
Done! `aead v0.1.0` is published.

tarcieri:
@newpavlov awesome, thank you! Will update the `chacha20poly1305` PR to use the released crate

PR #50, opened 2019-08-29
aead: Fix comments on Aead::encrypt/decrypt

tarcieri:
I just noticed that in #49 I swapped the `Aead`/`AeadMut` names, but didn't update the comments accordingly, so `Aead` is documented as saying "Refer to the `Aead::encrypt`/`decrypt` docs. Oops.

This commit swaps the comments to match.

### DIFF
--- a/aead/rustfmt.toml
+++ b/aead/rustfmt.toml
@@ -1,1 +1,0 @@
-merge_imports = true
--- a/aead/src/lib.rs
+++ b/aead/src/lib.rs
@@ -1,4 +1,16 @@
-//! A set of traits designed to support authenticated encryption.
+//! Authenticated Encryption with Associated Data (AEAD) traits
+//!
+//! This crate provides an abstract interface for AEAD ciphers, which guarantee
+//! both confidentiality and integrity, even from a powerful attacker who is
+//! able to execute [chosen-ciphertext attacks]. The resulting security property,
+//! [ciphertext indistinguishability], is considered a basic requirement for
+//! modern cryptographic implementations.
+//!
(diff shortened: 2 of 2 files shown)


newpavlov:
Maybe it's also worth to add some top-level documentation?

newpavlov:
BTW you can write:
>[`AeadMut`] and [`Aead::encrypt()`]
To automatically link types, methods and other stuff.

tarcieri:
@newpavlov pushed up some more comments and tried to use the linkification syntax you suggested, but a local `cargo doc --open` doesn't have them linked 😢

aead/rustfmt.toml
```diff
@@ -1 +0,0 @@
-merge_imports = true
```
tarcieri: This is super annoying for stable Rust users:

```
$ cargo fmt
Warning: can't set `merge_imports = true`, unstable features are only available in nightly channel.
```

newpavlov: I think you can remove this file and formatting can be changed later.

newpavlov:
>local cargo doc --open doesn't have them linked

Hm, looks like this feature is nightly-only at the moment (can't find a link to a tracking issue right now). IIRC docs.rs uses nightly toolchain so links will be properly generated there.

tarcieri:
@newpavlov cool, mind merging then?

tarcieri:
@newpavlov when you get a chance, can you cut another release of the `aead` crate to update the docs.rs documentation?

newpavlov:
@tarcieri Will do.

newpavlov:
I forgot that I've already done it. :)

PR #51, opened 2019-08-29
aead: Add From impls between (&[u8], &[u8]) and Payload

tarcieri:
Random idea I had: I thought there might be scenarios where these conversions are convenient.

They do eliminate the nice labels you get out of a struct though, so they seem like used incorrectly they could be a footgun.

I'm on the fence, but I thought I'd throw them up for a second look.


newpavlov:
I am not sure if we should allow it. Developers are lazy, so usually they will use `encrypt(nonce, (a, b))` instead of `encrypt(nonce, Payload { msg: a, ad: b })` if they can. In this case it's not sufficiently better than the old `encrypt(nonce, a b)` variant.

PR #52, opened 2019-08-30
universal-hash: Initial crate implementation

tarcieri:
This crate contains a generic trait for [universal hash functions] (e.g. GHASH, POLYVAL, Poly1305) whose API is shaped by my experiences implementing both [Poly1305] and [POLYVAL] and trying to build a common abstraction with the things I want, which are the following:

- `Block` as the unified abstraction for all input and output values: performance implementations will immediately want to load these values into an XMM register on x86 architectures. For this reason this implementation enforces that all `Block` types impl `Copy`, which (empirically, by my own measurements) provides better performance than using references.
- Built-in `subtle` comparison support ala `crypto-mac`: let's not beat around the bush here, while a UHF is not a "MAC" in and of itself, we're really always going to be using them to implement MACs (possibly as an internal implementation detail of an AEAD mode that doesn't need to wrap it in the `crypto-mac` trait because it's only consuming it internally), so we may as well make it easy.
- An `input_padded()` API which inputs a byte slice block-by-block and zero pads the last "chunk" if necessary: this is pretty much the main API that any AEAD mode is going to want to use. Every RFC I've read for an AEAD mode built on a UHF-based MAC has a whole section describing a "zero pad to the block size" function, so let's define it once and be done with it.

cc @eternaleye @isislovecruft

[universal hash functions]: https://en.wikipedia.org/wiki/Universal_hashing
[Poly1305]: https://github.com/RustCrypto/MACs/tree/master/poly1305
[POLYVAL]: https://github.com/RustCrypto/MACs/tree/master/polyval

### DIFF
--- a/universal-hash/src/lib.rs
+++ b/universal-hash/src/lib.rs
@@ -15,13 +15,10 @@
 
 #![no_std]
 #![doc(html_logo_url = "https://raw.githubusercontent.com/RustCrypto/meta/master/logo_small.png")]
-#![warn(missing_docs)]
+#![warn(missing_docs, rust_2018_idioms)]
 
 #[cfg(feature = "std")]
 extern crate std;
-
-pub extern crate generic_array;
-extern crate subtle;
 
 use generic_array::{ArrayLength, GenericArray};
 use generic_array::typenum::Unsigned;
--- a/universal-hash/src/lib.rs
+++ b/universal-hash/src/lib.rs
@@ -10,6 +10,8 @@
 //! When combined with a cipher, such as in Galois/Counter Mode or the
 //! Salsa20 family AEAD constructions, they can provide the core functionality
 //! for a Message Authentication Code (MAC).
+//!
+//! [Universal Hash Functions]: https://en.wikipedia.org/wiki/Universal_hashing
 
 #![no_std]
 #![doc(html_logo_url = "https://raw.githubusercontent.com/RustCrypto/meta/master/logo_small.png")]
@@ -22,19 +24,20 @@ pub extern crate generic_array;
 extern crate subtle;
 
 use generic_array::{ArrayLength, GenericArray};
+use generic_array::typenum::Unsigned;
 use subtle::{Choice, ConstantTimeEq};
 
 /// The `UniversalHash` trait defines a generic interface for universal hash
 /// functions.
-pub trait UniversalHash<BlockSize: ArrayLength<u8>>: Clone
-where
-    BlockSize::ArrayType: Copy,
-{
+pub trait UniversalHash: Clone {
+    /// Size of a block (e.g. field element) this universal hash operates on
+    type BlockSize: ArrayLength<u8>;
+
     /// Instantiate a universal hash function with the given key
-    fn new(key: Block<BlockSize>) -> Self;
+    fn new(key: Block<Self::BlockSize>) -> Self;
 
     /// Input a block into the universal hash function
-    fn input_block(&mut self, block: Block<BlockSize>);
+    fn update_block(&mut self, block: Block<Self::BlockSize>);
 
     /// Input data into the universal hash function. If the length of the
     /// data is not a multiple of the block size, the remaining data is
@@ -42,19 +45,20 @@ where
     ///
     /// This approach is frequently used by AEAD modes which use
     /// Message Authentication Codes (MACs) based on universal hashing.
-    fn input_padded(&mut self, data: &[u8]) {
-        let block_size = BlockSize::to_usize();
+    fn update_padded(&mut self, data: &[u8]) {
+        let mut chunks = data.chunks_exact(Self::BlockSize::to_usize());
 
-        for chunk in data.chunks(block_size) {
-            if chunk.len() == block_size {
-                let block_bytes = *GenericArray::from_slice(chunk);
-                self.input_block(block_bytes.into());
-            } else {
-                // Copy the non-aligned chunk into a zero-filled array
-                let mut padded_block = GenericArray::default();
-                padded_block[..chunk.len()].copy_from_slice(chunk);
-                self.input_block(padded_block.into());
-            }
+        for chunk in &mut chunks {
+            let block_bytes = GenericArray::clone_from_slice(chunk);
+            self.update_block(block_bytes.into());
+        }
+
+        let rem = chunks.remainder();
+
+        if !rem.is_empty() {
+            let mut padded_block = GenericArray::default();
+            padded_block[..rem.len()].copy_from_slice(rem);
+            self.update_block(padded_block.into());
         }
     }
 
@@ -62,11 +66,11 @@ where
     fn reset(&mut self);
 
     /// Obtain the output `Block` of a `UniversalHash` function and consume it.
-    fn result(self) -> Block<BlockSize>;
+    fn result(self) -> Block<Self::BlockSize>;
 
     /// Obtain the output `Block` of a `UniversalHash` computation and reset it back
     /// to its initial state.
-    fn result_reset(&mut self) -> Block<BlockSize> {
+    fn result_reset(&mut self) -> Block<Self::BlockSize> {
         let res = self.clone().result();
         self.reset();
         res
@@ -75,7 +79,7 @@ where
     /// Verify the `UniversalHash` of the processed input matches a given output
     /// `Block`. This is useful when constructing Message Authentication Codes (MACs)
     /// from universal hash functions.
-    fn verify(self, output: Block<BlockSize>) -> Result<(), Error> {
+    fn verify(self, output: Block<Self::BlockSize>) -> Result<(), Error> {
         if self.result() == output {
             Ok(())
         } else {
@@ -90,18 +94,14 @@ where
 /// It provides a safe `Eq` implementation that runs in constant time, which
 /// is useful for implementing Message Authentication Codes (MACs) based on
 /// universal hashing.
-#[derive(Copy, Clone)]
-pub struct Block<N: ArrayLength<u8>>
-where
-    N::ArrayType: Copy,
-{
+#[derive(Clone)]
+pub struct Block<N: ArrayLength<u8>> {
     bytes: GenericArray<u8, N>,
 }
 
 impl<N> Block<N>
 where
     N: ArrayLength<u8>,
-    N::ArrayType: Copy,
 {
     /// Create a new `Block`.
     pub fn new(bytes: GenericArray<u8, N>) -> Block<N> {
@@ -112,7 +112,6 @@ where
 impl<N> From<GenericArray<u8, N>> for Block<N>
 where
     N: ArrayLength<u8>,
-    N::ArrayType: Copy,
 {
     fn from(bytes: GenericArray<u8, N>) -> Self {
         Block { bytes }
@@ -122,7 +121,6 @@ where
 impl<N> ConstantTimeEq for Block<N>
 where
     N: ArrayLength<u8>,
-    N::ArrayType: Copy,
 {
     fn ct_eq(&self, other: &Self) -> Choice {
         self.bytes.ct_eq(&other.bytes)
@@ -132,19 +130,13 @@ where
 impl<N> PartialEq for Block<N>
 where
     N: ArrayLength<u8>,
-    N::ArrayType: Copy,
 {
     fn eq(&self, x: &Block<N>) -> bool {
         self.ct_eq(x).unwrap_u8() == 1
     }
 }
 
-impl<N> Eq for Block<N>
-where
-    N: ArrayLength<u8>,
-    N::ArrayType: Copy,
-{
-}
+impl<N: ArrayLength<u8>> Eq for Block<N> {}
 
 /// Error type for when the output `Block` of a `UniversalHash`
 /// is not equal to the expected value.
--- a/universal-hash/src/lib.rs
+++ b/universal-hash/src/lib.rs
@@ -1,0 +1,167 @@
+//! Traits for [Universal Hash Functions].
+//!
+//! Universal hash functions select from a "universal family" of possible
+//! hash functions selected by a key. They are well suited to the purpose
+//! of "one time authenticators" for a sequence of bytestring inputs,
+//! as their construction has a number of desirable properties such as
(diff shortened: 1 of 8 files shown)


tarcieri:
@newpavlov I omitted a `crypto-mac`-like `dev` module for now but that was just because I wanted to push up a first cut at this. I think it would be a very awesome thing to have, and can add it to this PR.

universal-hash/src/lib.rs
```diff
@@ -0,0 +1,167 @@
+//! Traits for [Universal Hash Functions].
```
newpavlov: You forgot to add this link.

universal-hash/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+
+pub extern crate generic_array;
+extern crate subtle;
+
+use generic_array::{ArrayLength, GenericArray};
+use subtle::{Choice, ConstantTimeEq};
+
+/// The `UniversalHash` trait defines a generic interface for universal hash
+/// functions.
+pub trait UniversalHash<BlockSize: ArrayLength<u8>>: Clone
```
newpavlov: Shouldn't be `BlockSize` an associated type instead?

universal-hash/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+    /// Input data into the universal hash function. If the length of the
+    /// data is not a multiple of the block size, the remaining data is
+    /// padded with zeroes up to the `BlockSize`.
+    ///
+    /// This approach is frequently used by AEAD modes which use
+    /// Message Authentication Codes (MACs) based on universal hashing.
+    fn input_padded(&mut self, data: &[u8]) {
+        let block_size = BlockSize::to_usize();
+
+        for chunk in data.chunks(block_size) {
```
newpavlov: You can write a bit more efficient code like this:
```rust
let mut chunks = data.chunks_exact(block_size);
for chunk in &mut chunks {
    let block_bytes = GenericArray::clone_from_slice(chunk);
    self.input_block(block_bytes.into());
}
let rem = chunks.remainder();
if !rem.is_empty() {
    let mut padded_block = GenericArray::default();
    padded_block[..chunk.len()].copy_from_slice(rem);
    self.input_block(padded_block.into());
}
```

universal-hash/src/lib.rs
tarcieri: If you can figure out how to make the bound on the next line work with associated types...

```rust
where
    BlockSize::ArrayType: Copy,
```

I don't think Rust's type system supports associated type bounds yet:

https://github.com/rust-lang/rust/issues/52662

universal-hash/src/lib.rs
tarcieri: Most of that looks good, however...

```
GenericArray::clone_from_slice(chunk)
```

...you're calling `Clone` on a (XMM register-sized) `Copy`  type

universal-hash/src/lib.rs
newpavlov: IIUC you need the `Copy` bound only for de-referncing in `*GenericArray::from_slice(chunk)`. In that case I think we can just replace it with clone (see the other thread).

universal-hash/src/lib.rs
newpavlov: IIUC `clone()` for `Copy` types is as efficient as a simple `memcpy`, so performance-wise `GenericArray::clone_from_slice(chunk)` and `*GenericArray::from_slice(chunk)` should be equivalent.

newpavlov: Plus here you still working with bytes, conversion is done later with `into()`.

tarcieri: I guess I can get rid of the `Copy` bound for now, but when I didn't have it on both Poly1305 and POLYVAL it resulted in more `movups` instructions whereas using a `Copy` type let them accumulate directly into an XMM register.

tarcieri: (note: everything I'm describing was with regular `[u8; 16]`s. Not sure what effect `GenericArray` would have on that)

newpavlov: We also could use `GenericArray::from_slice(chunk)` and change `fn input_block(&mut self, block: Block<BlockSize>);` to `fn input_block(&mut self, block: &GenericArray<u8, BlockSize>);`. BTW maybe it's worth to name it as something like `block_update` in the light of the future renaming of methods?

tarcieri: Changing my other impls from `&[u8; 16]` to `[u8; 16]` resulted in ~5% speedup, which is what I was trying to preserve here.

I used `input_*` for consistency, but if you'd like to just go strait to `update_*` that's fine by me.

newpavlov: Hm, this is strange. From `for` loop you essentially get a single pointer and it shouldn't matter where it will be read. Can you repeat benchmarks but with trait this time? Don't forget to mark `input_block` with `#[inline]`.

>I used input_* for consistency, but if you'd like to just go straight to update_* that's fine by me.

I think it's fine to per-emptively rename it now, it will be less stuff to update later.

tarcieri: I think I'd just rather drop the `Copy` bound, for simplicity and consistency's sake (so we can switch back to an associated type)

tarcieri:
@newpavlov updated with your suggestions

tarcieri:
Welp:

```
error[E0599]: no method named `chunks_exact` found for type `&[u8]` in the current scope
  --> src/lib.rs:49:31
   |
49 |         let mut chunks = data.chunks_exact(Self::BlockSize::to_usize());
   |     
```

@newpavlov can we drop 1.21?

newpavlov:
Yeap, I think we can go straight to 1.32 and 2018 edition and exclude this crate from the 1.21 CI job like we did with `aead`.

tarcieri:
@newpavlov ok, I guess we could go straight to 2018 edition if you made a new `RustCrypto/universal-hashes` repo and we moved `poly1305` and `polyval` out of there.

Will add this crate to the excludes list and do a 2018 edition upgrade

newpavlov:
Done! I've added the new repo under supervision of MACs team, so you should have the necessary access rights.

tarcieri:
@newpavlov awesome, thanks!

I already squatted the `universal-hash` name, but just sent you an owner invite for it (would've added the `traits` team but I don't appear to be on it)

tarcieri:
@newpavlov just pushed a 2018 edition upgrade for `universal-hash`. If the tests pass, I'd say it's good to go.

tarcieri:
And we're green.

@newpavlov look good to you?

universal-hash/src/lib.rs
```diff
@@ -0,0 +1,156 @@
...
+/// functions.
+pub trait UniversalHash: Clone {
+    /// Size of a block (e.g. field element) this universal hash operates on
+    type BlockSize: ArrayLength<u8>;
+
+    /// Instantiate a universal hash function with the given key
+    fn new(key: Block<Self::BlockSize>) -> Self;
+
+    /// Input a block into the universal hash function
+    fn update_block(&mut self, block: Block<Self::BlockSize>);
```
newpavlov: Isn't `Block` is unnecessary here? We don't need const eq in update block, plus IIUC with the current code implementations will not be able to extract the underlying buffer. Plus I think it may be better to pass a reference (i.e. `&GenericArray<u8, Self::BlockSize>`), so code which puts data into XMM register will be able to write `_mm_loadu_si128(block.as_ptr() as *const __m128i)`. Same for the `new` method.

tarcieri: It was originally a more convenient way to represent the `Copy` bound.

I can make both `new(key: ...)` and `update_block(...` take a `GenericArray` instead, and change the `Block` type to be `Output`. Does that sound good to you?

newpavlov: Yes, looks good, also don't forget `verify`. What do you think about using references?

tarcieri: Updated as discussed

universal-hash/src/lib.rs
```diff
@@ -0,0 +1,154 @@
...
+    fn result_reset(&mut self) -> Output<Self::BlockSize> {
+        let res = self.clone().result();
+        self.reset();
+        res
+    }
+
+    /// Verify the `UniversalHash` of the processed input matches a given [`Output`].
+    /// This is useful when constructing Message Authentication Codes (MACs)
+    /// from universal hash functions.
+    fn verify(self, output: Output<Self::BlockSize>) -> Result<(), Error> {
```
newpavlov: You haven't updated `verify` to using `GenericArray` as an input paramter.

universal-hash/src/lib.rs
```diff
@@ -0,0 +1,154 @@
...
+use subtle::{Choice, ConstantTimeEq};
+
+/// The `UniversalHash` trait defines a generic interface for universal hash
+/// functions.
+pub trait UniversalHash: Clone {
+    /// Size of a block (e.g. field element) this universal hash operates on
+    type BlockSize: ArrayLength<u8>;
+
+    /// Instantiate a universal hash function with the given key
+    fn new(key: GenericArray<u8, Self::BlockSize>) -> Self;
```
newpavlov: So what about using `&GenericArray<u8, Self::BlockSize>` here, as well as in `update_block` and `verify`?

tarcieri: Done, however I changed `verify` to:

```rust
fn verify(self, other: impl Into<Output<Self::BlockSize>>) -> Result<(), Error>
```

A `&GenericArray<u8, Self::BlockSize>` will still work, because I also added an `Into` impl for it.

tarcieri:
Okay, we're green again 😉 

@newpavlov everything look good now?

universal-hash/src/lib.rs
```diff
@@ -0,0 +1,162 @@
...
+    fn result_reset(&mut self) -> Output<Self::BlockSize> {
+        let res = self.clone().result();
+        self.reset();
+        res
+    }
+
+    /// Verify the `UniversalHash` of the processed input matches a given [`Output`].
+    /// This is useful when constructing Message Authentication Codes (MACs)
+    /// from universal hash functions.
+    fn verify(self, other: impl Into<Output<Self::BlockSize>>) -> Result<(), Error> {
```
newpavlov: Why do we need this generality here? I think using something like this would be good enough:
```rust
fn verify(self, other: &GenericArray<u8, Self::BlockSize>) -> Result<(), Error> {
    if self.result() == other.into() {
        Ok(())
    } else {
        Err(MacError)
    }
}
```

universal-hash/src/lib.rs
```diff
@@ -0,0 +1,162 @@
...
+}
+
+impl<N> Output<N>
+where
+    N: ArrayLength<u8>,
+{
+    /// Create a new `Block`.
+    pub fn new(bytes: GenericArray<u8, N>) -> Output<N> {
+        Output { bytes }
+    }
```
newpavlov: Note that we don't have any [`code`](https://docs.rs/crypto-mac/0.7.0/crypto_mac/struct.MacResult.html#method.code) method to retrieve result bytes. If I am not missing something it would make this type unusable if we want to transfer or store the resulting tag.

universal-hash/src/lib.rs
```diff
@@ -0,0 +1,162 @@
...
+//! for a Message Authentication Code (MAC).
+//!
+//! [Universal Hash Functions]: https://en.wikipedia.org/wiki/Universal_hashing
+
+#![no_std]
+#![doc(html_logo_url = "https://raw.githubusercontent.com/RustCrypto/meta/master/logo_small.png")]
+#![warn(missing_docs, rust_2018_idioms)]
+
+#[cfg(feature = "std")]
+extern crate std;
```
newpavlov: BTW maybe add `pub extern crate generic_array;` like we do in other trait crates?

newpavlov:
>everything look good now?

Not quite, see the earlier comments. :D

universal-hash/src/lib.rs
tarcieri: Ok I'll get rid of it for now

universal-hash/src/lib.rs
tarcieri: Good catch!

I'm not sure `code` is the appropriate name here, but I do agree there needs to be a way to do this.

How about `into_bytes` and/or an appropriate `Into` impl?

universal-hash/src/lib.rs
newpavlov: Ah, I wrote the other comment before seeing your reply. Let's discuss it there.

universal-hash/src/lib.rs
tarcieri: Oops, I had that but it got lost in the 2018 edition upgrade. Will do a `pub use generic_array`

universal-hash/src/lib.rs
newpavlov: I think a dedicated method will be easier to discover. `into_bytes` sounds good (we probably will need to rename `MacResult` method for consistency later). `Into` can be added as well if you think it will be useful for users.

universal-hash/src/lib.rs
newpavlov: IIRC `pub extern` works on 2018 edition. I think using `pub use` will not be detected by rustdoc as a [re-export](https://docs.rs/crypto-mac/0.7.0/crypto_mac/#reexports).

universal-hash/src/lib.rs
tarcieri: Ok, `Output::into_bytes` added in 83b438f

universal-hash/src/lib.rs
tarcieri: `pub extern` is disallowed by the `rust_2018_idioms` lint:

Screen Shot 2019-08-30 at 11 49 44 AM

`pub use` is detected as a re-export by rustdoc:

Screen Shot 2019-08-30 at 11 49 19 AM

newpavlov: Ah, good to know.

newpavlov:
Ok, everything looks good to me now! :)

tarcieri:
@newpavlov awesome, will open a PR to cut a release

PR #59, opened 2019-10-04
aead: Add in-place API based on `Buffer` trait

tarcieri:
NOTE: This PR includes the commit which adds the detached API from #58. It might make sense to review that first.

This PR adds a `Buffer` trait for `Vec`-like types which impl `AsRef<[u8]>`, `AsMut<[u8]>`, `extend_from_slice()`, and `truncate()`, and optionally impls it for `Vec<u8>` (when the `alloc` feature is enabled), along with the option to use `heapless::Vec` (when the `heapless` feature is enabled), providing a true `#![no_std]`-friendly option.

Uses `impl Buffer` as the argument for `encrypt_in_place` and `decrypt_in_place`, with default implementations that assume a postfix authentication tag. This allows these methods to handle ciphertext assembly and parsing, after which they can invoke the `_detached` APIs.

### DIFF
--- a/aead/Cargo.toml
+++ b/aead/Cargo.toml
@@ -7,12 +7,16 @@ license = "MIT OR Apache-2.0"
 description = "Traits for Authenticated Encryption with Associated Data (AEAD) algorithms"
 documentation = "https://docs.rs/aead"
 repository = "https://github.com/RustCrypto/traits"
-keywords = ["digest", "crypto", "encryption"]
+keywords = ["crypto", "encryption"]
 categories = ["cryptography", "no-std"]
 
 [dependencies]
 generic-array = { version = "0.12", default-features = false }
+heapless = { version = "0.5", optional = true }
 
 [features]
 default = ["alloc"]
 alloc = []
+
+[package.metadata.docs.rs]
+all-features = true
--- a/aead/src/lib.rs
+++ b/aead/src/lib.rs
@@ -18,48 +18,34 @@
 extern crate alloc;
 
 pub use generic_array;
+#[cfg(feature = "heapless")]
+pub use heapless;
 
 #[cfg(feature = "alloc")]
 use alloc::vec::Vec;
 use generic_array::{typenum::Unsigned, ArrayLength, GenericArray};
 
-// Define the default implementation for both `Aead::encrypt()` and
-// `AeadMut::encrypt()`. Uses a macro to gloss over `&self` vs `&mut self`.
-#[cfg(feature = "alloc")]
-macro_rules! encrypt_to_postfix_tagged_vec {
-    ($aead:expr, $nonce:expr, $payload:expr) => {{
-        let payload = $payload.into();
-        let mut buffer = Vec::with_capacity(payload.msg.len() + Self::TagSize::to_usize());
-        buffer.extend_from_slice(payload.msg);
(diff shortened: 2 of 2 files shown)


aead/Cargo.toml
```diff
@@ -7,8 +7,16 @@ license = "MIT OR Apache-2.0"
 description = "Traits for Authenticated Encryption with Associated Data (AEAD) algorithms"
 documentation = "https://docs.rs/aead"
 repository = "https://github.com/RustCrypto/traits"
-keywords = ["digest", "crypto", "encryption"]
+keywords = ["crypto", "encryption"]
 categories = ["cryptography", "no-std"]
 
 [dependencies]
 generic-array = { version = "0.12", default-features = false }
+heapless = { version = "0.5", optional = true }
```
tarcieri: Alternatively we could define our own `heapless::Vec`-like type which wraps a `GenericArray<u8, N>` and does the minimum necessary to impl the `Buffer` trait, which would eliminate this dependency. But `heapless` seems relatively popular in the embedded space.

tarcieri:
@newpavlov mind taking a look at this? (and #58)

newpavlov:
@tarcieri 
Can you resolve the conflicts?

I am a bit hesitant about this PR, looks like it will be a bit weird to work with modes which store tag in prefix. And the `Buffer` trait looks a bit out of place as well. Do you or someone else plan to use API added by this PR? If yes, I will merge it, but I am inclined to remove/rework it in the next minor version.

tarcieri:
@newpavlov 

> Can you resolve the conflicts?

Done

> I am a bit hesitant about this PR, looks like it will be a bit weird to work with modes which store tag in prefix.

I think it will work great with the `offset` parameter proposed in #58, since that can handle shifting the ciphertext/plaintext during encryption/decryption.

In that case, you'd extend the buffer by the tag length with e.g. zeroes for encryption, encrypt with an offset, and return the buffer.

For decryption, copy the tag out to a type, and then do offset in-place decryption, shifting the plaintext back to the beginning of the buffer, and truncating.

> Do you or someone else plan to use API added by this PR? 

Not immediately. It's a proposal and I'm happy to both think about it and try some alternative designs.

Perhaps `bytes` would be helpful here.

tarcieri:
(Note: the test failure looks unrelated)

newpavlov:
Hm, your arguments are reasonable. Let's merge it then and see how this API will play out. I will release v0.1.2 with both PRs right away.

newpavlov:
As for CI failure, looks like we will have to disable no-std build for the time being, since `aead` enables `alloc` by default and there is no good way to disable default features when building workspace...

tarcieri:
@newpavlov awesome! In that case, it will be interesting to shift the current `RustCrypto/aead` crate implementations to use this, as for AEADs which have a typical postfix tag, the default impl should suffice, and it should allow eliminating a lot of boilerplate in those crates for doing the same buffer management.

tarcieri:
I guess another random thought I had about these was splitting them into separate traits with blanket impls, but I think having one trait and some running code which uses it might be more helpful to start with

newpavlov:
`aead v0.1.2` is published!

I think there is still a sizable room for improvements, but it will require a certain amount of deliberation and design work, ideally supported by user experience reports.

>CI

Duh, the CI failure should be fixed by simply removing `--no-default-features`...

tarcieri:
@newpavlov awesome, thanks a lot!

PR #61, opened 2019-10-30
GitHub Actions

### LINKED ISSUE #60: Test GitHub Actions for CI
I noticed a bit of wonkiness in the current CI setup for this project. This seems to be due to the use of Cargo workspaces, which has terrible interactions with features. Can we try separate builds for each crate in a repo? I think there are a bunch of benefits to this:

- Provides more flexibility for testing the individual crates since the complexity associated with multiple features and/or targets won't pollute the entire repository.
- Allows for separate MSRV for each crate.
- Might help with PRs such as https://github.com/RustCrypto/hashes/pull/85 getting merged sooner.

Travis CI is probably not suitable for this since it'll lead to one large `.travis.yml` file with a lot of conditionals and environment variables. With GitHub Actions, you can have multiple workflows in each repo that can be triggered based on path-based filtering. So, a PR that modifies `md5` wouldn't trigger the `sha2` workflow. Also, each workflow file can reside in its respective subdirectory.

Would be great if we could do a trial run of GitHub Actions on the `traits` repo (without removing Travis builds).

akhilles:
Closes #60.

The `workspace.yml` workflow is shared by all crates in the workspace. Workspace MSRV should be the maximum MSRV of all the individual crates.

Some crates might have additional test cases (older MSRV, combinations of features, custom compile flags such as `+aes`), and this is easily supported by adding crate-specific workflows that execute conditionally.

### DIFF
--- a/.github/workflows/block-cipher-trait.yml
+++ b/.github/workflows/block-cipher-trait.yml
@@ -1,0 +1,27 @@
+name: block-cipher-trait
+on:
+  push:
+    paths:
+    - 'block-cipher-trait/**'
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        rust:
+          - stable
+          - beta
+          - nightly
+          - 1.21.0 # MSRV
+    steps:
+    - uses: actions/checkout@v1
+    - uses: actions-rs/toolchain@v1
+      with:
+        profile: minimal
+        toolchain: ${{ matrix.rust }}
+    - run: |
+        cargo check --all-features
+        cargo test --features dev
+        cargo test --features std
+        cargo test --all-features
+      working-directory: block-cipher-trait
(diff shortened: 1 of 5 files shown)


akhilles:
Example CI run: https://github.com/akhilles/traits/runs/281098471

newpavlov:
Thank you! I will try to take a look at it this weekend.

akhilles:
Can this be merged? It won’t affect the existing Travis builds.

tarcieri:
FYI, I've adopted GitHub Actions for the AEADs repo, using a config similar to the one proposed here: https://github.com/RustCrypto/AEADs/tree/master/.github/workflows

I'd suggest going forward with this, although it seems like it will need some tweaking, namely in terms of the number of jobs created.

tarcieri:
Thanks for the contribution! After a bit of work I was able to merge your commits (after rebasing) in #103.

Closing this as effectively merged.

PR #66, opened 2020-01-18
chore: no-change generic-array 0.12 -> 0.13

FauxFaux:
> One of my projects has a bunch of duplicate versions of generic-array, through block-buffer (from sha-2), crypto-mac (through hmac) and digest (through hmac and sha2) -- all of them through usage in postgres-protocol. Here's a start at fixing this up.

I'm using chacha20poly1305 and sha2 directly and end up with *six* dependencies on the old generic-array.

c.f. https://github.com/RustCrypto/utils/pull/39

### DIFF
--- a/aead/Cargo.toml
+++ b/aead/Cargo.toml
@@ -11,7 +11,7 @@ keywords = ["crypto", "encryption"]
 categories = ["cryptography", "no-std"]
 
 [dependencies]
-generic-array = { version = "0.12", default-features = false }
+generic-array = { version = "0.13", default-features = false }
 heapless = { version = "0.5", optional = true }
 
 [features]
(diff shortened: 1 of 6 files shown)


tarcieri:
Guess we'll need to bump MSRV for this. Otherwise I think only @newpavlov has access to release these crates (and might have some opinions on other work that needs to be done first)

djc:
@newpavlov ping? It's been more than a month.

tarcieri:
@djc can you rebase? MSRV on `master` has been bumped to 1.31, which should fix the build error

FauxFaux:
Done.

tarcieri:
Cool, it's green now, so I'll go ahead and merge. Unfortunately I don't have access to do a release, though.

PR #68, opened 2020-02-25
Remove deprecated std::error::Error::description() and bump MSRV from 1.21 to 1.31

linkmauve:
It’s been deprecated since Rust 1.41, and we already have a Display implementation anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ sudo: required
 
 matrix:
   include:
-    - rust: 1.21.0
+    - rust: 1.31.0
       script: cargo test --verbose --all --exclude aead --exclude universal-hash --release
     - rust: 1.36.0
       script: cargo test --verbose --package aead --package universal-hash --release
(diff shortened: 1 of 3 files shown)


linkmauve:
The CI is failing only on Rust 1.21, due to this deprecated function being mandatory back then.

There is another failure on this version, rustc-demangle is too recent for it too: https://travis-ci.org/RustCrypto/traits/jobs/655133561#L254

Should we fix rustc-demangle to compile on such an old version, or should we drop support for Rust 1.21?

tarcieri:
@linkmauve yeah, go ahead and bump MSRV, I guess to the minimum required to merge this PR until I can get clarity from @newpavlov on what our new MSRV should be

linkmauve:
I bumped the MSRV to 1.27, as it seems to be the first version which soft-deprecated `Error::description()`.

linkmauve:
There, bumped it to 1.31, which works for all crates this time, and shouldn’t break in versions in the middle.

tarcieri:
Will merge this for now, and worst case we can revert to an earlier MSRV

linkmauve:
Anything earlier will leave out `crypto-mac` and `aead`, which have been ported to the 2018 edition.  What could be done is to exclude these two crates from this CI pass, but meh.

PR #82, opened 2020-03-09
Mark preview feature as unstable in Cargo.toml

roblabla:
I (and I'm sure a few others) often look at the Cargo.toml to find out what a crate's features are. It might be a good idea to explicitly point out there that the preview features are unstable.

### DIFF
--- a/signature/Cargo.toml
+++ b/signature/Cargo.toml
@@ -32,10 +32,13 @@ sha2 = { version = "0.8", default-features = false }
 
 [features]
 default = ["std"]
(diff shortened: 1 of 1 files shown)


tarcieri:
Hrmm, the build is green on Travis CI, but it seems reporting isn't working or something? (I've been having various other problems with it this morning)

https://travis-ci.org/RustCrypto/traits/builds/660230340?utm_source=github_status&utm_medium=notification

Anyway, this looks good to me, thanks!

PR #85, opened 2020-03-11
Handles variable-length nonce

### LINKED ISSUE #65: aead: support variable-length nonces
Currently, the AEAD-trait only supports fixed-length nonces. However, there are some AEAD algorithms, which allow variable length nonces to be used. I think this is important not only for interoperability with other implementations, but also to not restrict the input of AEAD algorithms arbitrarily. Some cases have already been discussed in RustCrypto/AEADs#62 and https://github.com/RustCrypto/traits/pull/40/files/a5db8f2d8e07593883e7e6aba27eb4e9d54b9460#r307931303.

I could imagine, for type-safety, adding a new trait like `AeadVarlenNonce`, which is only implemented for algorithms supporting it, or adding a method to the existing `Aead` trait, which takes a `&[u8]` as nonce, but may return an `Err` (or `None` or panic) if the algorithm doesn't support varlen nonces.

/cc @tarcieri

masihyeganeh:
I'm trying to fix #65, but I'm not sure I'm on the right track.
This PR changes `nonce` type to `&[u8]` instead of fixed-sized `GenericArray`.
The main code to handle variable-length nonces will be in `RustCrypto/AEADs`.

### DIFF
--- a/aead/src/lib.rs
+++ b/aead/src/lib.rs
@@ -74,8 +74,6 @@ pub trait NewAead {
 /// This trait is intended for use with stateless AEAD algorithms. The
 /// [`AeadMut`] trait provides a stateful interface.
 pub trait Aead {
-    /// The length of a nonce.
-    type NonceSize: ArrayLength<u8>;
     /// The maximum length of the nonce.
     type TagSize: ArrayLength<u8>;
     /// The upper bound amount of additional space required to support a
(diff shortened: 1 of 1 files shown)


tarcieri:
I like the idea of it being possible to use a slice for a nonce (this is a blocker AES-SIV too), but I don't like it being unilateral, because it introduces potential errors when using AEADs with a fixed-size nonce (which is the case 99.9% of the time).

I think it might make sense to make the nonce a generic parameter. See the sketch of that idea I suggested on the original `Aead` trait PR:

https://github.com/RustCrypto/traits/pull/40#issuecomment-516878391

masihyeganeh:
I like your solution better. I just not exactly sure how to do it.
This variable-length nonce feature is holding my project back. I am willing to help implementing it.
Just let me know if I can help you with this.

tarcieri:
It seems https://github.com/RustCrypto/AEADs/pull/126 takes care of supporting any-sized GCM none without changing the trait.

This also seems like a general solution to the problem (i.e. making `Aead::NonceSize` generic) so I think that's actually more desirable than trying to support slices as nonces.

Issue #43, opened 2019-08-13
New crate versions


newpavlov:
Migrating to 2018 edition is a long overdue and since it's a good chance to introduce breaking changes, I propose to use this issue to discuss them. Right now I have the following list of changes:

- [x] Migrate to 2018 edition and bump MSRV to ~~1.36~~ 1.41 (needed for `generic-array` v0.14)
- [x] Update `generic-array` to ~~v0.13~~ v0.14.
- [x] Release `crypto-mac v0.8` which uses `subtle v2`. (see #35)
- [x] Rename `crypto_mac::Mac::input` and `digest::Digest::input` methods to `update`.
- [x] Rename `digest::Input` to `digest::Update`.
- [x] Remove `digest::Reset` trait in favor of reworked `FixedResult` trait. See this [comment](https://github.com/RustCrypto/hashes/issues/86#issuecomment-509045192).
- [x] Deprecate `block-cipher-trait` and use `block-cipher` crate instead #28.
- [x] Introduce a new `NewBlockCipher` trait (#164).
- [x] Add `stream_cipher::FromBlockCipher` trait.
- [x] Add [`impl_write!`](https://github.com/RustCrypto/traits/blob/master/digest/src/lib.rs#L127) macro to `crypto-mac`.
- ~~Find solution for buffer-to-buffer cipher operations #31~~
- ~~Add `try_new` methods~~ (closed as wontfix, see #47)
- [x] Rename `MacResult` type to something like `Code`/`Output` and rename `code` method to `into_bytes` for consistency with the `universal-hash` crate.
- ~~Add macro for implementing `io::Write` to `crypto-mac` similarly to the `digest` crate.~~ (dup of above)
- [x] Bump `opaque-debug` to v0.2
- ~~Support variable length AEAD Nonces #65~~ (EDIT: solved by making ciphers generic over their `NonceSize`)
- [x] Rename `crypto_mac::Mac::result`, `digest::Digest::result`, and `universal_hash::UniversalHash::result` to `::finalize` or `::finish`

I may have forgotten some changes I had in mind. Feel free to suggest other public API changes.

tarcieri:
To me the common terminology is Initialize-Update-Finalize (IUF) and `update` would make sense IMO

newpavlov:
@burdges 
Can you provide examples of other libraries which use `input`? I would also prefer to keep the name to limit breakage, but it looks like it's fairly uncommon name for the method across hashing libraries.

[...]

PR #95, opened 2020-04-12
Uprev generic array to version 0.14

garbageslam:
In 0.14 they added a pretty nice feature, which is that you can
now split not just arrays that you hold by value, but you can
split references to generic arrays, into a pair of references to
generic arrays corresponding to the subslices, where the size
calculations are done correctly at compile-time and the buffer
size information is all preserved.

https://docs.rs/generic-array/0.14.1/src/generic_array/sequence.rs.html#302-320

I have some ECIES code that looks like this, that builds on `aead` trait:

```
    fn decrypt_in_place_detached(
        &self,
        key: &RistrettoPrivate,
        footer: &GenericArray<u8, Self::FooterSize>,
        buffer: &mut [u8],
    ) -> Result<(), Error> {
        let (footer, version_data): (_, &GenericArray<u8, U2>) = footer.split();
```

Alternate version is like

```
    fn decrypt_in_place_detached(
        &self,
        key: &RistrettoPrivate,
        footer: &GenericArray<u8, Self::FooterSize>,
        buffer: &mut [u8],
    ) -> Result<(), Error> {
        let (footer, version_data) = Split::<u8, U48>::split(footer);
```

I think these are both expected to work generic array 0.14. There
are alternatives of course but they are less tidy :)

Since I'm getting the generic array pub export from you, I think
that to use this, I have to convince you that it's worth it to uprev.
LMK what you think! Thanks

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,16 +4,16 @@ sudo: required
 
 matrix:
   include:
-    - rust: 1.31.0
+    - rust: 1.41.0
       script: cargo test --verbose --all --exclude aead --exclude signature --exclude universal-hash --release
-    - rust: 1.36.0
+    - rust: 1.43.0
       script: cargo test --verbose --package aead --package signature --package universal-hash --release
     - rust: stable
       script: cargo test --verbose --all --release
     - rust: nightly
       script: cargo test --verbose --all --release
     # tests if crates can be built with std feature
-    - rust: 1.31.0
+    - rust: 1.41.0
       script: ./build_std.sh
 
     - env: TARGET=i686-unknown-linux-gnu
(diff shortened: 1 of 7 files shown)


tarcieri:
Looks like this has an MSRV bump (which is fine), but the CI config needs to be updated

tarcieri:
Wow, just noticed the MSRV bump is to 1.41.

garbageslam:
sorry I lost track of this -- I will come back to it soon

tarcieri:
I'd say let's go for it

tarcieri:
Thanks!

Per #43 it might be a bit before this gets released.

PR #101, opened 2020-04-26
Forbid unsafe code in all crates

hhirtz:
`#![forbid(unsafe_code)]` attributes make rustc abort compilation if
there are any unsafe blocks in the crate, and exceptions cannot be made
with allow/warn attributes.

Didn't add badges to the readme because signature already has this attribute and no badge had been added.


tarcieri:
Thanks!

PR #120, opened 2020-05-23
aead: split Aead(Mut) from Aead(Mut)InPlace (fixes #70)

### LINKED ISSUE #70: aead: object safety
Right now the [`Aead`](https://docs.rs/aead/0.2.0/aead/trait.Aead.html) (and other) traits are not object safe, because of the `encrypt` method's use of `impl Into<Payload>` and `encrypt_in_place` method's use of `impl Buffer`.

I think I'd like to leave the `impl Into<Payload>` on `encrypt` as-is, but split `Aead` and `AeadMut` into two traits: one object safe, and one not.

Namely I propose we could have `AeadInPlace` and `AeadMutInPlace` which are object-safe. The `impl Buffer` can be replaced with a generic parameter on the trait itself, which would both permit object safety and allow for multiple impls (i.e. for `alloc::vec::Vec` and `heapless::Vec`). We could then have a blanket impl of `Aead` for `AeadInPlace<Vec>`, and `AeadMut` for `AeadMutInPlace<Vec>`.

This would also be helpful for AEAD provider crates who don't want to impl the in-place methods and just have allocating `Vec`-based APIs, which is the case for `ursa`: https://github.com/hyperledger/ursa/pull/91

tarcieri:
Splits the `Aead` and `AeadMut` trait into the following:

- `Aead` and `AeadMut`:gated on the `alloc` feature
- `AeadInPlace` and `AeadMutInPlace`: object-safe, always available

This further has the benefit that implementors who don't want to provide or abstract over the in-place APIs no longer have to. See `ursa` as a [real world example of where this split is useful](https://github.com/hyperledger/ursa/pull/91/files#diff-49cadd47a44ba93f412f87ba4cc2427fR105-R124).

To make the `*InPlace` traits object safe, the one generic parameter they previously used (`impl Buffer`) was switched to `dyn Buffer`. This adds some vtable dispatch overhead for manipulating the buffer, but hopefully it is still cheap compared to the overhead of cryptographic operations on the underlying data. That said, no benchmarks have been performed (yet) on the costs of this change.

### DIFF
--- a/aead/src/lib.rs
+++ b/aead/src/lib.rs
@@ -36,7 +36,10 @@ use generic_array::{typenum::Unsigned, ArrayLength, GenericArray};
 #[cfg(feature = "alloc")]
 use alloc::vec::Vec;
 
-/// Error type
+/// Error type.
+///
+/// This type is deliberately opaque as to avoid potential side-channel
+/// leakage (e.g. padding oracle).
 #[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
 pub struct Error;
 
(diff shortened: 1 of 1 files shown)


tarcieri:
Going to go ahead and merge this, but if anyone has any feedback on it, I'd certainly be interested in hearing it and am happy to do some follow-up passes to revise.

/cc @newpavlov @jcape @burdges @WildCryptoFox

burdges:
I suppose https://github.com/rust-lang/rfcs/pull/2930 could influence the direction this takes in future. Also maybe https://github.com/rust-lang/rfcs/pull/2895

PR #138, opened 2020-05-26
WIP Add "aead::encrypt_fixed_length" and "aead::decrypt_fixed_length"

garbageslam:
I wanted to suggest adding these two APIs which provide alternative
ways to use aead in a no-alloc build environment, without using the
heapless library. Instead it uses `GenericArray::{split, concat}`
to put the tag at the end of the buffer.

I liked using this API in the context when, a struct containing
fixed length buffers is where we need to place the ciphertext.

The mc-crypto-box trait is based closely on this aead trait, and
that's where I got experience using this version of the `encrypt`
function:

https://github.com/mobilecoinofficial/mobilecoin/blob/master/crypto/box/src/traits.rs#L140

I thought I would propose this extension upstream here.

Note: I'm having a little difficulty making the trait-bounds for generic array
work out, I'm going to poke it a little more if you are interested in this.

----

Fwiw I also made this "fixed_buffer" wrapper object to make fixed-length
mutable buffers compatible with the aead buffer trait, I found it was okay,
but I'm not using it anymore. I think it might be useful for someone else,
I'm not sure.

If you are interested in carrying an object like this next to the aead::Buffer
trait, lmk and I will open a PR, otherwise I think I might delete it since we aren't
using it: https://github.com/mobilecoinofficial/mobilecoin/blob/master/crypto/box/src/fixed_buffer.rs#L3

### DIFF
--- a/aead/src/lib.rs
+++ b/aead/src/lib.rs
@@ -30,11 +30,10 @@ pub use generic_array::{self, typenum::consts};
 #[cfg(feature = "heapless")]
 pub use heapless;
 
-use core::fmt;
-use generic_array::{typenum::Unsigned, ArrayLength, GenericArray};
-
 #[cfg(feature = "alloc")]
 use alloc::vec::Vec;
+use core::{fmt, ops::{Add, Sub}};
+use generic_array::{sequence::{Split, Concat}, typenum::{Sum, Diff, Unsigned}, ArrayLength, GenericArray};
 
 /// Error type.
 ///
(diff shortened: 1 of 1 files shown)


tarcieri:
This is an interesting suggestion and use of type-level arithmetic, but a bit problematic given recent changes...

Namely in #120, the `Aead` and `AeadMut` traits were split from `AeadInPlace` and `AeadMutInPlace`. The former have blanket impls for the latter, and also rely on the latter to handle assembly of the ciphertext message for encryption, and parsing it for decryption. A blanket impl means it's not possible to override the implementation, and this implementation necessitates a postfix tag, where some ciphers require a prefix tag or other message elements in other orderings.

There are at least two ciphers (well, really three, or perhaps four) implemented in [RustCrypto/AEADs](https://github.com/RustCrypto/AEADs/tree/master/aes-siv) which rely on prefix as opposed to postfix tags: AES-SIV (and AES-PMAC-SIV), XSalsa20Poly1305 (as well as Curve25519XSalsa20Poly1305, a.k.a. `crypto_box`).

Unfortunately adding these methods to the `Aead(Mut)InPlace` trait would defeat one of the major motivations for the refactor: making the trait object safe.

Perhaps this method could be added via a separate trait?

garbageslam:
thanks for your comments! I will think about this

burdges:
There is seemingly no reason for trait methods here since simple wrapper types suffice, although maybe extension traits work nicer if you still need access to the underlying cipher.  

I'd expect large generic arrays and `concat` to harm build times, so encouraging this interface sounds problematic.  If folks want this, then ideally one should add doc comments on what optimization flags result in `concat` optimizing correctly, although maybe I should trust LLVM more.  ;)  

As a rule, one should use `split_at_mut` to do this with manual optimizations at run time, with `GenericArray::from_slice` providing the unsafe code required by the current traits.

newpavlov:
I think it also would be nice to think about such functionality in the context of future migration to const generics. IIUC at the early stages they will be less powerful than `typenum`, so I am not sure if it will be possible to translate API in this PR.

tarcieri:
Going to close this out. I'd suggest implementing it at the library level.

PR #163, opened 2020-06-02
digest: add `alloc` feature

tarcieri:
Allows `DynDigest` and `*_vec` methods to function in `no_std` + `alloc` environments.


burdges:
I suppose if folks ever do some reasonable array traits then arrayvec could implement them..

PR #179, opened 2020-06-09
block-cipher: add BlockCipherMut

tarcieri:
Adds a stateful equivalent to `BlockCipher` that permits `&mut self` access to the underlying type.

The main use case for this trait is hardware cryptographic accelerators which need to e.g. communitate with a peripheral device via an underlying `&mut` reference.

While it's possible to use some underlying logic to use the existing `BlockCipher` trait in such a scenario, the solutions are somewhat ugly. Here is a real-world example:

https://github.com/iqlusioninc/usbarmory.rs/blob/develop/firmware/usbarmory/src/dcp/aes128.rs#L198-L236

The idea with `BlockCipherMut` would be to alternatively provide `AeadMut`/`AeadMutInPlace` for AEAD modes with an underlying `BlockCipherMut` (when possible).

### DIFF
--- a/block-cipher/src/lib.rs
+++ b/block-cipher/src/lib.rs
@@ -99,3 +99,32 @@ pub trait BlockCipher {
         }
     }
 }
+
+/// Stateful block cipher which permits `&mut self` access.
+///
+/// The main use case for this trait is hardware encryption engines which
+/// require `&mut self` access to an underlying hardware peripheral.
+pub trait BlockCipherMut {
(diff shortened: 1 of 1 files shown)


tarcieri:
I’d agree: I just did it for feature parity, but these use cases are always `ParBlocks=U1`, so it might be nice to omit it for simplicity.

tarcieri:
Updated to remove `ParBlocks`

PR #180, opened 2020-06-09
digest: add `FixedOutputDirty` trait + `finalize_into*`

tarcieri:
Closes https://github.com/RustCrypto/hashes/issues/86

~~Replaces the `Reset` trait with a set of `_reset` methods on all of the various traits, and uses these as the default impl for the methods which consume hashers.~~

~~Also adds a set of methods to `FixedOutput` (`finalize_fixed_into*`) which eliminate copies by accepting an existing buffer as an argument.~~

Adds a `FixedOutputDirty` trait which writes the digest output to a provided byte array, but does not reset the internal state. This is intended for implementations to use in order to ensure that they are not reset in the event the instance is consumed.

Also adds a set of `finalize_into` and `finalize_into_reset` methods to `FixedOutput` whhich also write their input into a provided byte array, and changes the existing `finalize_fixed` (and newly added `finalize_fixed_reset`) methods to have a default implementation which returns a byte array allocated on the stack.

Finally, adds a blanket impl of `FixedOutput` for `FixedOutputDirty` + `Reset` types which handles safely invoking the underlying implementation by either consuming the instance (avoiding a reset) or borrowing the hasher, obtaining the output, and resetting.

### DIFF
--- a/digest/src/dyn_digest.rs
+++ b/digest/src/dyn_digest.rs
@@ -34,7 +34,7 @@ impl<D: Update + FixedOutput + Reset + Clone + 'static> DynDigest for D {
     }
 
     fn finalize_reset(&mut self) -> Box<[u8]> {
-        let res = self.clone().finalize_fixed().to_vec().into_boxed_slice();
+        let res = self.finalize_fixed_reset().to_vec().into_boxed_slice();
         Reset::reset(self);
         res
     }
--- a/digest/src/lib.rs
+++ b/digest/src/lib.rs
@@ -79,8 +79,64 @@ pub trait FixedOutput {
     /// Output size for fixed output digest
     type OutputSize: ArrayLength<u8>;
 
-    /// Retrieve result and consume hasher instance.
-    fn finalize_fixed(self) -> GenericArray<u8, Self::OutputSize>;
+    /// Write result into provided array and consume the hasher instance.
+    fn finalize_into(self, out: &mut GenericArray<u8, Self::OutputSize>);
+
+    /// Write result into provided array and reset the hasher instance.
+    fn finalize_into_reset(&mut self, out: &mut GenericArray<u8, Self::OutputSize>);
+
+    /// Retrieve result and consume the hasher instance.
+    #[inline]
+    fn finalize_fixed(self) -> GenericArray<u8, Self::OutputSize>
+    where
+        Self: Sized,
+    {
+        let mut out = Default::default();
+        self.finalize_into(&mut out);
+        out
+    }
+
+    /// Retrieve result and reset the hasher instance.
+    #[inline]
+    fn finalize_fixed_reset(&mut self) -> GenericArray<u8, Self::OutputSize> {
+        let mut out = Default::default();
+        self.finalize_into_reset(&mut out);
+        out
+    }
+}
+
+/// Trait for fixed-output digest implementations to use to retrieve the
+/// hash output.
+///
+/// Usage of this trait in user code is discouraged. Instead use the
+/// [`FixedOutput::finalize_fixed`] or [`FixedOutput::finalize_fixed_reset`]
+/// methods.
+///
+/// Types which impl this trait along with [`Reset`] will receive a blanket
+/// impl of [`FixedOutput`].
+pub trait FixedOutputDirty {
+    /// Output size for fixed output digest
+    type OutputSize: ArrayLength<u8>;
+
+    /// Retrieve result into provided buffer and leave hasher in a dirty state.
+    ///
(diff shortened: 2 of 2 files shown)


tarcieri:
@newpavlov this is slightly different from what you originally proposed, but I think achieves the same effect. Namely if you wanted something exactly like what you had before, it could be achieved with:

```rust
impl MyDigest {
    /// Private method not intended to be used directly
    fn finalize_fixed_core(&mut self, out: &mut GenericArray<u8, Self::OutputSize>, reset: bool) {
        [...]
    }
}

pub trait FixedOutput {
    type OutputSize: ArrayLength<u8>;

    #[inline]
    fn finalize_fixed(self) -> GenericArray<u8, Self::OutputSize> {
        let mut buf = Default::default();
        self.finalize_fixed_core(&mut buf, false);
        buf
    }

    #[inline]
    fn finalize_fixed_reset(&mut self) -> GenericArray<u8, Self::OutputSize> { .. }
    fn finalize_fixed_into(self, out: &mut GenericArray<u8, Self::OutputSize>) { .. }
    fn finalize_fixed_into_reset(&mut self, out: &mut GenericArray<u8, Self::OutputSize>) { .. }
}
```

digest/src/dyn_digest.rs
```diff
@@ -18,33 +18,28 @@ pub trait DynDigest {
...
-        self.finalize_fixed().to_vec().into_boxed_slice()
+        self.clone()
+            .finalize_fixed_reset()
+            .to_vec()
+            .into_boxed_slice()
     }
 
-    fn reset(&mut self) {
-        Reset::reset(self);
+    fn finalize(mut self: Box<Self>) -> Box<[u8]> {
```
tarcieri, line 41: @newpavlov out of curiosity, why does this return a boxed slice rather than a `Vec<u8>`?

newpavlov: `Vec<u8>` takes 3 words, while `Box<[u8]>` only two.

tarcieri:
That's only if you rely on the default implementation though. If you want to avoid the resets entirely, just define your own implementation (per above) which doesn't invoke the reset code at all.

digest/src/dyn_digest.rs
tarcieri: Interesting!

newpavlov:
This will lead to unnecessary code duplication, which I would like to avoid. How about something like this?
```rust
pub trait FixedOutput {
    // ...

    /// Retrieve result into provided buffer and leave hasher in a dirty state.
    ///  Usage of this method in user code is discouraged, prefer `finalize_fixed`
    /// or `finalize_fixed_reset`.
    fn finalize_into_dirty(&mut self, out: &mut GenericArray<u8, Self::OutputSize>);

    fn finalize_into_reset(&mut self, out: &mut GenericArray<u8, Self::OutputSize>)
    where
        Self: Reset,
    {
        self.finalize_into_dirty(out);
        self.reset();
    }
}

pub trait Reset {
    fn reset(&mut self);
}
```
Yes, users may misuse `finalize_into_dirty` (use a scarier name?), but I don't think it will be a significant issue in practice.

tarcieri:
I think you could achieve something like that with a blanket impl:

pub trait FixedOutputDirty {
    /// Retrieve result into provided buffer and leave hasher in a dirty state.
    /// Usage of this method in user code is discouraged, prefer `finalize_fixed`
    /// or `finalize_fixed_reset`.
    fn finalize_into_dirty(&mut self, out: &mut GenericArray<u8, Self::OutputSize>);
}

impl<D: FixedOutputDirty + Reset> FixedOutput for D {
    fn finalize_into(&mut self, out: &mut GenericArray<u8, Self::OutputSize>)
    where
        Self: Reset,
    {
        self.finalize_into_dirty(out);
    }

    fn finalize_into_reset(&mut self, out: &mut GenericArray<u8, Self::OutputSize>)
    where
        Self: Reset,
    {
        self.finalize_into_dirty(out);
        self.reset();
    }
}

tarcieri:
Also should I shorten `finalize_fixed_into` to `finalize_into`? Seems ok to me.

newpavlov:
The separate trait looks good, making room for `FixedOutput` to be used separately from the dirty trait (e.g. for HW accelerated impls) is certainly nice.

>Also should I shorten finalize_fixed_into to finalize_into?

Yes, sounds good.

tarcieri:
Updated to use a `FixedOutputDirty` trait + blanket impl

newpavlov:
BTW should we do something similar for XOF and `VariableOutput`?

tarcieri:
Sure

PR #181, opened 2020-06-09
Use boxed slices instead of vectors

newpavlov:
For consistency with the `DynDigest` trait.

### DIFF
--- a/digest/src/lib.rs
+++ b/digest/src/lib.rs
@@ -45,7 +45,7 @@ pub use dyn_digest::DynDigest;
 use generic_array::{ArrayLength, GenericArray};
 
 #[cfg(feature = "alloc")]
-use alloc::vec::Vec;
+use alloc::boxed::Box;
 
 /// Trait for updating digest state with input data.
 pub trait Update {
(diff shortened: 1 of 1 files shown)


digest/src/lib.rs
```diff
@@ -157,12 +157,16 @@ pub trait VariableOutput: core::marker::Sized {
...
+    /// Retrieve result into a boxed slice and consume hasher.
+    ///
+    /// `Box<[u8]>` is used instead of `Vec<u8>` to save stack space, since
+    /// they have size of 2 and 3 words respectively.
     #[cfg(feature = "alloc")]
     #[cfg_attr(docsrs, doc(cfg(feature = "alloc")))]
-    fn finalize_vec(self) -> Vec<u8> {
-        let mut buf = Vec::with_capacity(self.output_size());
-        self.finalize_variable(|res| buf.extend_from_slice(res));
+    fn finalize_box(self) -> Box<[u8]> {
```
tarcieri, line 166: Curious if `finalize_boxed` might make more sense here. I've seen both (e.g. `box_clone` seems common, but I've seen `_boxed` as well)

PR #184, opened 2020-06-09
digest: factor out `fixed`, `variable`, and `xof` modules

tarcieri:
Each of these now defines two traits and a blanket impl, so `lib.rs` was getting rather bloated.

This commit factors each of them into their own modules.


digest/src/variable.rs
```diff
@@ -0,0 +1,104 @@
...
+
+    /// Retrieve result into a boxed slice and reset hasher state.
+    ///
+    /// `Box<[u8]>` is used instead of `Vec<u8>` to save stack space, since
+    /// they have size of 2 and 3 words respectively.
+    #[cfg(feature = "alloc")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "alloc")))]
+    fn finalize_boxed_reset(&mut self) -> Box<[u8]> {
+        let n = self.output_size();
+        let mut buf = vec![0u8; n].into_boxed_slice();
```
newpavlov, line 53: Hm, don't we have to explicitly import the `vec` macro? Does CI test build with an enabled `alloc` feature?

tarcieri: > Hm, don't we have to explicitly import the vec macro?

It's being imported from the alloc crate:

https://github.com/RustCrypto/traits/blob/master/digest/src/lib.rs#L23

>  Does CI test build with an enabled `alloc` feature?

Yes, it tests each feature individually as well as `--all-features` (and the build with `--no-default-features` on Cortex-M and WASM)

https://github.com/RustCrypto/traits/blob/master/.github/workflows/digest.yml#L55

newpavlov: Ah, great. Haven't used `#[macro_use]` in a long while. :)

tarcieri: It might be possible to import it as `use alloc::vec` instead but the nice thing about the `#[macro_use]` is it feels the same as when `std` is enabled and it's already in the prelude.

PR #185, opened 2020-06-09
digest: document three levels of organization

tarcieri:
The new `*Dirty` traits add another level of organization to the crate's design. This commit updates the documentation to reflect that.

Rendered:

Screen Shot 2020-06-09 at 12 10 55 PM

### DIFF
--- a/digest/src/lib.rs
+++ b/digest/src/lib.rs
@@ -1,15 +1,23 @@
 //! This crate provides traits which describe functionality of cryptographic hash
 //! functions.
 //!
-//! Traits in this repository can be separated into two levels:
-//! - Low level traits: [`Update`], [`BlockInput`], [`Reset`], [`FixedOutput`],
-//! [`VariableOutput`], [`ExtendableOutput`]. These traits atomically describe
-//! available functionality of hash function implementations.
-//! - Convenience trait: [`Digest`], [`DynDigest`]. They are wrappers around
-//! low level traits for most common hash-function use-cases.
+//! Traits in this repository are organized into high-level convenience traits,
+//! mid-level traits which expose more fine-grained functionality, and
+//! low-level traits intended to only be used by algorithm implementations:
+//!
+//! - **High-level convenience traits**: [`Digest`], [`DynDigest`]. They are wrappers
+//!   around lower-level traits for most common hash-function use-cases.
+//! - **Mid-level traits**: [`Update`], [`BlockInput`], [`Reset`], [`FixedOutput`],
+//!   [`VariableOutput`], [`ExtendableOutput`]. These traits atomically describe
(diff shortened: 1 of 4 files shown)


digest/src/lib.rs
```diff
@@ -1,12 +1,20 @@
...
+//!
+//! - **High-level convenience traits**: [`Digest`], [`DynDigest`]. They are wrappers
+//!   around lower-level traits for most common hash-function use-cases.
+//! - **Mid-level traits**: [`Update`], [`BlockInput`], [`Reset`], [`FixedOutput`],
+//!   [`VariableOutput`], [`ExtendableOutput`]. These traits atomically describe
+//!   available functionality of hash function implementations.
+//! - **Low-level traits**: [`FixedOutputDirty`], [`VariableOutputDirty`],
+//!   [`ExtendableOutputDirty`]. These traits are intended to be implemented by
+//!   low-level algorithm providers only and simplify the amount of work
+//!   implementers need to do, but may panic if used incorrectly and therefore
```
newpavlov, line 16: I don't think implementors of those traits will panic.

digest/src/lib.rs
```diff
@@ -1,12 +1,20 @@
...
+//!   shouldn't be used in application-level code.
 //!
 //! Additionally hash functions implement traits from `std`: `Default`, `Clone`,
 //! `Write`. (the latter depends on enabled-by-default `std` crate feature)
```
newpavlov, line 20: Strictly speaking, `std` feature is enabled by downstream crates, not by this crate itself. Also I would change `std` to "standard library", since `Default` and `Clone` are defined in core. :)

digest/src/lib.rs
tarcieri: Aah, it's in some of the other docs, but I can remove that.

That said, as I'm updating some of the `RustCrypto/hashes` crates I'm literally dealing with some code where it'd be quite advantageous not to panic, so that makes sense.

digest/src/lib.rs
tarcieri: Hey you wrote all that 😉

newpavlov: I know, but it's a good chance to fix it. :)

tarcieri: Updated to: "(the latter depends on the downstream crate having a `std` feature)"

digest/src/lib.rs
tarcieri: Removed any mention of panicking and described the method's contract better on each of the traits

digest/src/lib.rs
newpavlov: I've update this paragraph a bit, hopefully it become a bit more clear.

PR #191, opened 2020-06-11
aead: add `NewAead::new_varkey` method

tarcieri:
Adds a method which accepts a slice for a key, similar to the methods in the `block-cipher` and `stream-cipher` crates.

### DIFF
--- a/aead/src/lib.rs
+++ b/aead/src/lib.rs
@@ -67,8 +67,22 @@ pub trait NewAead {
     /// The size of the key array required by this algorithm.
     type KeySize: ArrayLength<u8>;
 
-    /// Construct a new stateful instance for the given key.
+    /// Create a new AEAD instance with the given key.
     fn new(key: &Key<Self>) -> Self;
+
+    /// Create new AEAD instance from key with variable size.
+    ///
+    /// Default implementation will accept only keys with length equal to `KeySize`.
+    fn new_varkey(key: &[u8]) -> Result<Self, Error>
+    where
+        Self: Sized,
+    {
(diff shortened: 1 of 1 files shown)


tarcieri:
@newpavlov one thing I noticed was the method names are inconsistent among the crates.

- `block-cipher` uses `new_varkey`
- `stream-cipher` uses `new_var`

I'd suggest either standardizing on `new_varkey` or, to bikeshed, something like `from_slice` (as a breaking change across all 3 crates, for some future hypothetical release)

burdges:
I'd suggest `fn new_varkey(key: impl Borrow<[u8]>) -> Result<Self, Error>` so that fixed size array work.

newpavlov:
Stream cipher initialization function takes both key and nonce, while block cipher takes only key. For example, `BlockMode` also uses `new_var`. I guess we can rename `new_varkey` to `new_var` in the next breaking release and maybe add `new_var` as an alias until then.

tarcieri:
@burdges we've used `impl AsRef<[u8]>` elsewhere. I think it's more exhaustive in terms of its coercions, which could be a good thing or a bad thing...

> Stream cipher initialization function takes both key and nonce, while block cipher takes only key. 

Ok, makes sense

tarcieri:
For consistency with the other `new_var` / `new_varkey` APIs I'd suggest just taking a slice for now, and revisiting taking `impl Borrow<[u8]>` or `impl AsRef<[u8]>` across all three crates the next time we make a breaking change

tarcieri:
This was released in v0.3.1:

https://docs.rs/aead/0.3.1/aead/trait.NewAead.html

PR #194, opened 2020-06-13
Add dev module to the aead crate

newpavlov:
I plan to use it for [`ccm`](RustCrypto/AEADs#174) and maybe later migrate `aes-gcm` tests as well (170 KB for a file with test code does not look great :smile: ).

I also could add a benchmark macro based on your Criterion-based benches.

### DIFF
--- a/aead/src/dev.rs
+++ b/aead/src/dev.rs
@@ -51,7 +51,11 @@ macro_rules! new_test {
             let data = include_bytes!(concat!("data/", $test_name, ".blb"));
             for (i, row) in Blob6Iterator::new(data).unwrap().enumerate() {
                 let [key, nonce, aad, pt, ct, status] = row.unwrap();
-                let pass = status[0] != 0;
+                let pass = match status[0] {
+                    0 => false,
+                    1 => true,
+                    _ => panic!("invalid value for pass flag"),
+                };
                 if let Err(reason) = run_test(key, nonce, aad, pt, ct, pass) {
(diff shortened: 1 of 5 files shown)


aead/src/dev.rs
```diff
@@ -0,0 +1,62 @@
...
+                    .map_err(|_| "encryption failure")?;
+                if res != pt {
+                    return Err("encrypted data is different from target ciphertext");
+                }
+                let res = cipher
+                    .decrypt(nonce, Payload { aad: aad, msg: ct })
+                    .map_err(|_| "decryption failure");
+                if res != pt {
+                    return Err("decrypted data is different from target plaintext");
+                }
```
tarcieri, line 38: Another thing most of the current AEAD tests do is tweak the ciphertext and ensure that inauthentic ciphertext is detected (i.e. that the MAC is being verified)

newpavlov: Yeah, I was also thinking about adding a modifier which would switch generated function between "all tests should return `Ok`" and "all tests should return `Err`". Alternatively we could add a sixth field, which would indicated expected test status.

Another potential improvement is allow reuse of blobs without duplicating them, it would be especially useful for keys. But it would require introducing backwards incompatible changes to `blobby`.

newpavlov:
If you don't have objections against `blobby` depending on `alloc`, I will replace the git dependency after `blobby v0.3` publication.

PR #197, opened 2020-06-16
README.md: add inline logo

tarcieri:
[Rendered](https://github.com/RustCrypto/traits/blob/12b5819c0d8d90eb77938302389ca0112958fca5/README.md)


tarcieri:
@newpavlov WDYT?

tarcieri:
>  I wonder how other projects handle logos in README (I couldn't find an example after a cursory search).

Generally when I add logos to projects they're large, vector-based imagines that take up the whole headline, e.g.:

https://github.com/iqlusioninc/signatory

> But I like the idea of using git.io URL shortener, how about addiing/replacing #![doc(html_logo_url = "..")] with a shortened URL?

Sure

PR #198, opened 2020-06-20
`cryptography` facade crate

tarcieri:
I've been given ownership of the [`cryptography`](https://crates.io/crates/cryptography) crate! 🎉 

This PR contains my suggested usage of it. If this sounds useful, I'll add @RustCrypto/traits as an owner.

---

Adds a `cryptography` crate which re-exports compatible versions of the other trait crates as a facade.

This should simplify cases where it's necessary to combine several cryptographic algorithms, by ensuring you have a single import to get compatible versions.

### DIFF
--- a/.github/workflows/cryptography.yml
+++ b/.github/workflows/cryptography.yml
@@ -1,0 +1,55 @@
+name: cryptography
+
+on:
+  pull_request:
+    paths:
+      - "cryptography/**"
+      - "Cargo.*"
+  push:
+    branches: master
+
+defaults:
+  run:
+    working-directory: cryptography
+
+env:
+  CARGO_INCREMENTAL: 0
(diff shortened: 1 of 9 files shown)


tarcieri:
Great! Let me flesh out the PR a little bit more.

newpavlov:
If you'll decide to add algorithms as well, I think it would be better to move this PR to the meta repository.

tarcieri:
Hrmm, that's an interesting point. As a meta-crate, perhaps it does belong in meta (even if it only contains the traits)

newpavlov:
If you plan to keep it traits-only, I have no problem in keeping it in this repository. I guess it should make updates a bit easier.

tarcieri:
Maybe we can start with that and experiment with moving it to `meta` later

PR #209, opened 2020-07-03
Add NonceSize to FromBlockCipher

newpavlov:
This change is needed for RustCrypto/block-ciphers#134. GOST modes are somewhat peculiar  as they allow additional flexibility, thus breaking the assumption used in the `FromBlockCipher` trait. Additionally this PR improves the blanket impl by overwriting the `new_var` method.

Unfortunately it's a breaking change, so a minor release will be needed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -45,7 +45,7 @@ dependencies = [
 name = "block-cipher"
 version = "0.7.1"
 dependencies = [
- "blobby 0.1.2",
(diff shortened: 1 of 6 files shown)


newpavlov:
I think, yes. :)

newpavlov:
I've bumped `stream-cipher` to v0.5.0, and because `cryptography` uses `path`, I had to bump it to v0.2 as well.

PR #210, opened 2020-07-07
Add blanket trait impls for references

newpavlov:
Closes RustCrypto/MACs#47

IIUC strictly speaking these changes are not backwards compatible, since downstream crates may define their own trait implementations for references. I wonder if we can overlook it and release these changes under patch releases.

Unfortunately it's not possible to write `impl<Alg: BlockCipherMut> BlockCipherMut for &mut Alg { .. }`, since it potentially conflicts with `impl<Alg: BlockCipher> BlockCipherMut for Alg { .. }`. Same goes for `StreamCipher`, which conflicts with the `SyncStreamCipher` impl. I think it's impossible to solve this problem without negative trait bounds, which are not even on horizon. Specialization may help, but I failed to make it work.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -45,7 +45,7 @@ dependencies = [
 name = "block-cipher"
 version = "0.8.0"
 dependencies = [
- "blobby 0.2.0",
+ "blobby 0.3.0",
  "generic-array 0.14.2",
 ]
 
@@ -254,7 +254,7 @@ checksum = "dba1a27d3efae4351c8051072d619e3ade2820635c3958d826bfea39d59b54c8"
 name = "stream-cipher"
 version = "0.6.0"
 dependencies = [
- "blobby 0.2.0",
+ "blobby 0.3.0",
  "block-cipher",
  "generic-array 0.14.2",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -43,7 +43,7 @@ dependencies = [
 
 [[package]]
 name = "block-cipher"
-version = "0.7.1"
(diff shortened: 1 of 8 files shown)


burdges:
I advocated for rand doing this way back, which I still support since it makes traits more ergonomic, but it brings some complexity too, and more breakage than expected.  I'd be cautious here since the symmetric crypto traits are more complex than rand's traits.  And asymmetric crypto traits become more complex still, making it unlikely this works there.

Ideally, rust would fix [E0666](https://doc.rust-lang.org/stable/error-index.html#E0666) so that `impl Borrow<impl Trait>` worked, and provide enough lazy normalization so that associated type bounds work correctly: https://github.com/rust-lang/rfcs/pull/2289#issuecomment-409337644

cc @hdevalence

newpavlov:
@burdges 
An analogous blanket impl for `RngCore` does [exist](https://github.com/rust-random/rand/blob/rand_core-0.5.1/rand_core/src/lib.rs#L388-L408) in the latest `rand_core` version and AFAIK works without issues. Can you remind issues to be aware of?

burdges:
https://github.com/dalek-cryptography/curve25519-dalek/issues/232
https://github.com/dalek-cryptography/curve25519-dalek/pull/219

newpavlov:
Ah, indeed. Then to be safe we should do a breaking release after all.

newpavlov:
@tarcieri 
If you don't have objections, I'll go ahead with merging this PR and published the new versions of `block-cipher` and `stream-cipher`.

PR #217, opened 2020-07-15
crypto-mac v0.9

newpavlov:
- Add `FromBlockCipher` trait and respective `NewMac` blanket implementation
- Bump `blobby` to v0.3
- Remove `impl_write!` macro, since it's useless for generic types

I couldn't make code generic over `From<BlockCipher>` trait due to the "unconstrained type parameter" error, so I had to introduce a new non-polymorphic trait instead. Any ideas how this problem can be worked around?


tarcieri:
> I couldn't make code generic over From<BlockCipher> trait due to the "unconstrained type parameter" error, so I had to introduce a new non-polymorphic trait instead. Any ideas how this problem can be worked around?

What's the concrete error? Do you have a full example?

newpavlov:
I've tried the following code:
```rust
impl<T, C> NewMac for T
where
    T: From<C>,
    C: BlockCipher + NewBlockCipher,
{
    type KeySize = C::KeySize;

    fn new(key: &Key<Self>) -> Self {
        C::new(key).into()
    }

    fn new_varkey(key: &[u8]) -> Result<Self, InvalidKeyLength> {
        C::new_varkey(key).map_err(|_| InvalidKeyLength)?.into()
    }
}
```
And I get this compilation error:
```
type, or predicates
   --> crypto-mac/src/lib.rs:124:9
    |
124 | impl<T, C> NewMac for T
    |         ^ unconstrained type parameter

error: aborting due to previous error
```
Which in hindsight should've been expected. Right now I have no idea how to properly structure code and if it's even possible in principle.

tarcieri:
@newpavlov aah yeah, that looks similar to what I was running into with `signature`. I ended up solving it with custom derive.

newpavlov:
So what do you think? Should we go ahead with the current code for the time being?

PR #220, opened 2020-07-25
elliptic-curve: add `Generate` trait

tarcieri:
Trait for randomly generating types using a `rand_core::CryptoRng`.

Potentially useful in trait bounds that need to generate e.g. `Scalar` types.

### DIFF
--- a/elliptic-curve/src/lib.rs
+++ b/elliptic-curve/src/lib.rs
@@ -21,14 +21,8 @@
 #[cfg(feature = "std")]
 extern crate std;
 
-#[cfg(feature = "rand_core")]
-pub use rand_core;
-
 pub mod error;
 pub mod secret_key;
-
-pub use generic_array::{self, typenum::consts};
-pub use subtle;
 
 // TODO(tarcieri): other curve forms
 #[cfg(feature = "weierstrass")]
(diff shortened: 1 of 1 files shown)


tarcieri:
`Scalar` is the main use case I have in mind for this. If there aren't others that are useful, it could potentially be constrained to that.

elliptic-curve/src/lib.rs
```diff
@@ -21,21 +21,30 @@
...
+use rand_core::{CryptoRng, RngCore};
 
 /// Byte array containing a serialized scalar value (i.e. an integer)
 pub type ScalarBytes<Size> = generic_array::GenericArray<u8, Size>;
+
+/// Trait for randomly generating a value
+#[cfg(feature = "rand_core")]
+pub trait Generate {
+    /// Generate a random element of this type using the provided [`CryptoRng`]
+    fn generate(rng: &mut (impl CryptoRng + RngCore)) -> Self;
```
newpavlov, line 49: `RngCore` is implemented for mutable RNG references, so we can use `rng: impl CryptoRng + RngCore` here.

tarcieri: Ooh, nice

PR #223, opened 2020-07-27
elliptic-curve: extract toplevel `Curve` trait

tarcieri:
This is a fairly substantial refactoring of the `elliptic-curve` crate which extracts a toplevel `elliptic_curve::Curve` trait and redefines all types to use it.

The `elliptic_curve::weierstrass::Curve` trait is maintained, but relegated to a marker trait. That said, it's still used as the marker trait for all traits/types defined in the `elliptic_curve::weierstrass` module.

Notably the main thing this facilitates is making `SecretKey` generic around a `C: elliptic_curve::Curve` generic type. The previous impetus was to avoid directly associating types dependent on scalars with specific elliptic curves (rather the curve's order), but this is ultimately unhelpful as it precludes accounting for trait impls on the curve in trait bounds.

With the new approach, we can use trait bounds to conditionally define methods on `SecretKey`, e.g. bounding a `SecretKey::generate` impl on curves which impl `Arithmetic where Self::Scalar: Generate`. This would eliminate the need for a special `GenerateSecretKey` trait.

One other notable change is the associated type for computing type sizes for a particular curve has been renamed to `ElementSize`, rather than the previous `ScalarSize`. This hopefully captures that this size is for all elements related to a particular curve, i.e. both the base and scalar fields.


tarcieri:
Note: this is effectively complete, however I'm going to attempt to migrate the existing curve implementations to use it locally and make sure there isn't anything I've missed.

tarcieri:
I did the updates to get the `ecdsa` crate building on top of this in https://github.com/RustCrypto/signatures/pull/105 and also have all of the current curve implementations building locally.

In both cases thanks to the existing type aliases, it was all fairly straightforward.

Based on that, and unifying all of the types/traits as being generic around a single `elliptic_curve::Curve` trait, I'm going to go ahead and merge this as I think it's a net improvement.

PR #225, opened 2020-07-27
elliptic-curve: fix warnings

tarcieri:
Fixes some unused import warnings overlooked in #224


tarcieri:
Weird, it seems like GitHub Actions is acting up today. The web UI reported success to me for #224, but then I saw later that some of the tests failed.

PR #228, opened 2020-07-28
elliptic-curve: add `Invert` trait

tarcieri:
Add a trait for performing scalar/field inversions.

The main motivation for this is to allow for things like `BlindedScalar` type which use random blinding for either efficiency reasons (to allow a vartime inversion) and/or as a side-channel defense.

### DIFF
--- a/elliptic-curve/src/lib.rs
+++ b/elliptic-curve/src/lib.rs
@@ -22,6 +22,7 @@
 extern crate std;
 
 pub mod error;
+pub mod ops;
 pub mod secret_key;
 
 // TODO(tarcieri): other curve forms
--- a/elliptic-curve/src/ops.rs
+++ b/elliptic-curve/src/ops.rs
@@ -1,0 +1,12 @@
+//! Traits for arithmetic operations on elliptic curve field elements
+
+use subtle::CtOption;
+
+/// Perform an inversion on a field element (i.e. base field element or scalar)
+pub trait Invert {
+    /// Field element type
+    type Output;
+
+    /// Invert a field element.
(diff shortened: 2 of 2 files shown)


tarcieri:
@nickray FYI, my intended use for this is to remove `masking_scalar` from the `SignPrimitive` API.

Instead we can have something like a `BlindedScalar` type constructed from a `Scalar` and a `CryptoRng` which supports an `Into<Scalar>` conversion, but whose `Invert::invert()` impl can use the vartime inversion.

I think if we prototype that in the `p256` crate, we could potentially hoist it into this crate and have a generic `BlindedScalar` implementation that works with any curve.

nickray:
The intended use is an interesting approach - I like it!

So far I never used any RNG traits (felt too implicit, I like to see exactly where entropy is consumed), but would be interested to see how you'd do this! Oh and my last attempt at field traits failed too 😅.

nickray:
I wonder how for practical purposes this CtOption approach compares to restricting the domain of this trait to just invertible elements?

tarcieri:
I didn't put a whole lot of thought into this trait and for me it's mostly it's a means to an end for implementing https://github.com/RustCrypto/elliptic-curves/pull/99

It's mostly a trait extraction of what exists in the `p256` and `k256` codebases, for both `Scalar` and `FieldElement`. I only impl'd it on the former, since that's all I care about for https://github.com/RustCrypto/elliptic-curves/pull/99

I'll make a note of it in https://github.com/RustCrypto/elliptic-curves/issues/22 as I'm sure there's some prior art here I've overlooked (or potentially another crate we can source the trait from).

PR #239, opened 2020-07-31
elliptic-curve: change MulBase trait to be impl'd on point types

tarcieri:
This makes the `MulBase` trait a static method for point types, which takes an associated scalar type as an associated parameter.

This allows multiple impls on both `AffinePoint` and `ProjectivePoint`.

### DIFF
--- a/elliptic-curve/src/ops.rs
+++ b/elliptic-curve/src/ops.rs
@@ -11,11 +11,13 @@ pub trait Invert {
     fn invert(&self) -> CtOption<Self::Output>;
 }
 
-/// Fixed-base scalar multiplication
-pub trait MulBase {
-    /// Affine point type
-    type Output;
+/// Fixed-base scalar multiplication.
+///
+/// This trait is intended to be implemented on a point type.
+pub trait MulBase: Sized {
+    /// Scalar type
+    type Scalar;
 
-    /// Multiply this value by the generator point for the elliptic curve
-    fn mul_base(&self) -> CtOption<Self::Output>;
+    /// Multiply scalar by the generator point for the elliptic curve
+    fn mul_base(scalar: &Self::Scalar) -> CtOption<Self>;
 }
--- a/elliptic-curve/src/lib.rs
+++ b/elliptic-curve/src/lib.rs
@@ -66,13 +66,10 @@ pub trait Curve: Clone + Debug + Default + Eq + Ord + Send + Sync {
 /// Elliptic curve with curve arithmetic support
 pub trait Arithmetic: Curve {
     /// Scalar type for a given curve
-    type Scalar: ConditionallySelectable
-        + Default
-        + secret_key::FromSecretKey<Self>
-        + ops::MulBase<Output = Self::AffinePoint>;
+    type Scalar: ConditionallySelectable + Default + secret_key::FromSecretKey<Self>;
 
     /// Affine point type for a given curve
-    type AffinePoint: ConditionallySelectable;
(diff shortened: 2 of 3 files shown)


elliptic-curve/src/ops.rs
```diff
@@ -11,11 +11,13 @@ pub trait Invert {
...
+///
+/// This trait is intended to be implemented on a point type.
+pub trait MulBase: Sized {
+    /// Scalar type
+    type Scalar;
 
-    /// Multiply this value by the generator point for the elliptic curve
-    fn mul_base(&self) -> CtOption<Self::Output>;
+    /// Multiply scalar by the generator point for the elliptic curve
+    fn mul_base(scalar: &Self::Scalar) -> CtOption<Self>;
```
fjarri, line 22: Hm, why is it returning an `Option`? Is it because `AffinePoint` doesn't have a state reflecting a point at infinity?

tarcieri: Yes. It makes it somewhat annoying on ProjectivePoint, unfortunately.

Open to alternatives, but I can't think of any.

fjarri: I see two possible options:

1. Make `AffinePoints` support infinity points. `libsecp256k1` does it, at least.

2. Have two traits, with `MulBase { mul_base() -> Self }` and `TryMulBase { try_mul_base() -> CtOption<Self> }`, with the default implementation in the latter returning the wrapped result of the former if it is implemented (I think it can be done in Rust, can it?)

tarcieri: If you'd like to implement the former, go for it. Otherwise option 2 seems straightforward enough, and yes blanket impls can work so `TryMulBase` is always impl'd for `MulBase`.

Another way to potentially make it infallible is to have it take a `NonZeroScalar` argument.

fjarri: 1 would probably require doing so for every curve in `elliptic-curves`? Not sure if I'm up for that right now.

If there was a `Point` trait of some kind, `MulBase` could actually in turn have a default implementation if  `Generator` and `Mul` are available (by the way, why is it `mul_base()`, but `generator()`?).

> Another way to potentially make it infallible is to have it take a NonZeroScalar argument.

I like that, but there's no such type yet, is there? A `NonZeroPoint` may be convenient as well.

Going further, one could distinguish between "infinity-enabled" and not affine points by having an `Identity` trait for points. But perhaps I'm overengineering here.

tarcieri: > by the way, why is it mul_base(), but generator()

If we added a `Generator` trait, and bounds for `Mul`, we could get rid of `MulBase`. Unfortunately a `Mul` impl on `AffinePoint` would run into the same problem.

`MulBase` is a stopgap.

> > NonZeroScalar argument.
>
> I like that, but there's no such type yet, is there?

Nope, however we could potentially make one that's a generic wrapper around `Scalar` which could be used for any curve.

fjarri: It could be resolved with an `Identity` trait. `Generator + Identity + Mul` -> `MulBase`, `Generator + Mul` -> `TryMulBase`.

With `NonZeroScalar` and `NonInfinityPoint` types, `TryMulBase` could have two methods: `mul_base(Scalar) -> Option<Point>` and `mul_base_nonzero(NonZeroScalar) -> NonInfinityPoint`.

Anyway, really, why not `mul_generator()` to be consistent?

tarcieri: Sure

PR #240, opened 2020-07-31
elliptic-curve: basic OID support

tarcieri:
Adds a (very simple, const-and-no_std-friendly) `ObjectIdentifier` type and an `Identifier` trait for attaching an OID to a curve.

No validation of the nodes is presently performed (and since it's impl'd as a `const fn`, it'd require a MSRV bump to do so). Also requires the OID be specified as an array literal: no parsing is supported for either string or binary-encoded OIDs.

Only string serialization is supported for now (via `Display`). A `no_std`-friendly binary serializer will be somewhat tricky.

### DIFF
--- a/elliptic-curve/src/lib.rs
+++ b/elliptic-curve/src/lib.rs
@@ -22,6 +22,7 @@
 extern crate std;
 
 pub mod error;
+pub mod oid;
 pub mod ops;
 pub mod secret_key;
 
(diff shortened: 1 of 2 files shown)


tarcieri:
I took a look at the [`oid` crate](https://crates.io/crates/oid) for this, unfortunately it has a hard dependency on liballoc and no support for const initialization.

newpavlov:
Ideally we should specify OID for other algorithms as well, so if `oid` is not a good fit, maybe create a separate crate for it?

tarcieri:
Yeah, good point, we could potentially split this out into a separate crate somewhere, although I can't think of a good name.

newpavlov:
`crypto-oid`?

tarcieri:
`const-oid` or `oid-consts` might describe the use case a little more specifically

PR #241, opened 2020-07-31
elliptic-curve: NonZeroScalar + Generator (replaces MulBase)

tarcieri:
Adds a `NonZeroScalar` type generic over curves along with a `Generator` trait which together with the `Mul` trait from `core` can replace the `MulBase` trait.

`NonZeroScalar` is able to use the bounds on `Arithmetic::Scalar` and therefore usable anywhere a `Scalar` otherwise would, but with the guarantee that multiplication by an `AffinePoint` will not result in the point at infinity, for which no `AffinePoint` representation is possible given current trait bounds/implementations.

cc @fjarri

### DIFF
--- a/elliptic-curve/src/scalar.rs
+++ b/elliptic-curve/src/scalar.rs
@@ -1,0 +1,51 @@
+//! Scalar types
+
+use crate::{Arithmetic, Curve};
+use subtle::{ConstantTimeEq, CtOption};
+
+#[cfg(feature = "zeroize")]
+use zeroize::Zeroize;
+
+/// Non-zero scalar type.
+///
+/// This type ensures that its value is not zero, ala `core::num::NonZero*`.
+/// To do this, the generic `S` type must impl both `Default` and
+/// `ConstantTimeEq`, with the requirement that `S::default()` returns 0.
+///
+/// In the context of ECC, it's useful for ensuring that scalar multiplication
+/// cannot result in the point at infinity.
+pub struct NonZeroScalar<C: Curve + Arithmetic> {
+    scalar: C::Scalar,
+}
+
+impl<C> NonZeroScalar<C>
+where
+    C: Curve + Arithmetic,
+{
+    /// Create a [`NonZeroScalar`] from a scalar, performing a constant-time
+    /// check that it's non-zero.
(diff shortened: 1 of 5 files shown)


elliptic-curve/src/scalar.rs
```diff
@@ -0,0 +1,48 @@
...
+pub struct NonZeroScalar<S: Default + ConstantTimeEq>(S);
+
+impl<S> NonZeroScalar<S>
+where
+    S: Default + ConstantTimeEq,
+{
+    /// Create a [`NonZeroScalar`] from a scalar, performing a constant-time
+    /// check that it's non-zero.
+    pub fn new(scalar: S) -> CtOption<Self> {
+        let is_some = scalar.ct_eq(&S::default());
```
fjarri, line 26: It's a bit weird using `default()` and not `zero()`/`is_zero()`, but I guess it's the best that can be done at the moment. Should probably be refactored when there's a fully-functional `Scalar` trait (or a collection of traits) is available.

tarcieri: Yeah, it's the only extant trait to define this in terms of right now.

However, I wouldn't say it's "weird" per say, at least for usage of `Default` in Rust: zero is the `Default` value for all numerical types.

fjarri: I suppose it's ok for simple types, but this approach may fail if `Scalar` ever implements lazy reduction arithmetic similar to that in `FieldElement` (and it will fail for the current `AffinePoint`/`ProjectivePoint`), because in that case you could have something that is zero/identity and yet isn't `default()`.

tarcieri:
Pushed up a slightly modified version of this PR, and also added `Generator` to it and removed `MulBase`. I have `k256` wired up with all of this locally.

I changed `NonZeroScalar` to be generic over `Curve + Arithmetic`, which allows it to reuse the trait bounds for `Arithmetic::Scalar`. Perhaps a better solution here would be a `Scalar` trait, but I think that's something to explore in a separate PR.

I think I'm going to move forward with this for now as it also solves the problem of generic variable-base scalar mult in addition to fixed-base, which is useful for a high-level generic ECDHE API. That's not to say I consider this to be the final say here, just incremental progress in exposing more functionality.

PR #243, opened 2020-08-03
elliptic-curve: add weierstrass::PublicKey::compress

tarcieri:
Generic point compression support for SEC-1 encoded public keys

### DIFF
--- a/elliptic-curve/src/weierstrass/point.rs
+++ b/elliptic-curve/src/weierstrass/point.rs
@@ -171,6 +171,16 @@ where
     pub fn into_bytes(self) -> GenericArray<u8, UncompressedPointSize<C>> {
         self.bytes
     }
+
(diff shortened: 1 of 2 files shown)


elliptic-curve/src/weierstrass/point.rs
```diff
@@ -171,6 +171,16 @@ where
     pub fn into_bytes(self) -> GenericArray<u8, UncompressedPointSize<C>> {
         self.bytes
     }
+
+    /// Get the x-coordinate of this curve point
+    pub(crate) fn x(&self) -> &ScalarBytes<C> {
```
tarcieri, line 176: Something of a misuse of the `Scalar` part of `ScalarBytes` here... perhaps it should be renamed to `ElementBytes` to reflect the `Curve::ElementSize` rename (from `Curve::ScalarSize`)

PR #251, opened 2020-08-06
elliptic-curve: Elliptic Curve Diffie-Hellman support

tarcieri:
Adds an `ecdh` module gated under an `ecdh` Cargo feature containing a generic high-level interface for performing ephemeral Elliptic Curve Diffie-Hellman key exchanges.

The API is inspired by the one in `x25519-dalek`.

### DIFF
--- a/elliptic-curve/Cargo.toml
+++ b/elliptic-curve/Cargo.toml
@@ -23,7 +23,9 @@ subtle = { version = "2.2.2", default-features = false }
 zeroize = { version = "1", optional = true,  default-features = false }
 
 [features]
-default = []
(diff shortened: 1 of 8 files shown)


newpavlov:
Would it make sense to keep it in a separate crate? We could ask @manuels if he's willing to transfer ownership of the `ecdh` crate (it hasn't been updated in 5 years).

tarcieri:
@newpavlov I'm not sure it really make sense to split this out into a separate crate yet. Maybe after the generic curve arithmetic traits and key types have settled down a bit.

PR #253, opened 2020-08-10
Use SVG logo

newpavlov:


signature/src/lib.rs
```diff
@@ -153,8 +153,10 @@
 
 #![no_std]
 #![cfg_attr(docsrs, feature(doc_cfg))]
+
 #![doc(
-    html_logo_url = "https://raw.githubusercontent.com/RustCrypto/meta/master/ferris_signer.png",
```
newpavlov, line 157: @tarcieri 
Should I keep the `ferris_signer` logo or are you fine with the standard one?

tarcieri: Could go either way. A standard logo is fine.

PR #260, opened 2020-08-14
stream-cipher: add SyncStreamCipherSeek::try_seek method

newpavlov:
Closes: RustCrypto/stream-ciphers#9

Unfortunately I forgot about this issue during the work on `stream-cipher` v0.6.

I wonder if it'll be worth to go even further and introduce something like this:
```rust
pub trait SyncStreamCipherSeek<T: num_traits::Unsigned = u64> {
    fn get_pos(&self) -> T {
        self.try_get_pos().unwrap()
    }
    fn try_get_pos(&self) -> Result<T, OverflowError>;
    fn seek(&mut self, pos: T) {
        self.try_seek(pos).unwrap()
    }
    fn try_seek(&mut self, pos: T) -> Result<(), LoopError>;
}
```

cc @kazcw

### DIFF
--- a/stream-cipher/src/lib.rs
+++ b/stream-cipher/src/lib.rs
@@ -255,6 +255,4 @@ macro_rules! impl_seek_num {
     };
 }
 
-// note: the trait is implemented for i32 to allow expressions
-// like `cipher.seek(10)`
 impl_seek_num! { u8 i8 u16 i16 u32 i32 u64 i64 u128 i128 }
--- a/stream-cipher/src/lib.rs
+++ b/stream-cipher/src/lib.rs
@@ -95,11 +95,11 @@ pub trait SyncStreamCipher {
     fn try_apply_keystream(&mut self, data: &mut [u8]) -> Result<(), LoopError>;
 }
 
-/// Trait for additionall functionality available for seekable stream ciphers.
+/// Trait for seekable stream ciphers.
 ///
-/// Methods of this trait are generic over the following numeric types: `u8`,
-/// `u16`, `u32`, `u64`, and `u128`. This is done via a "sealed" `SeekNum`
-/// trait. By default methods
+/// Methods of this trait are generic over the [`NumSeek`] trait, which is
+/// implemented for the following numeric types: `u8`, `u16`, `u32`, `u64`,
+/// and `u128`. By default methods will use `u64`.
 pub trait SyncStreamCipherSeek<T: SeekNum = u64> {
     /// Try to get current keystream position
     ///
--- a/stream-cipher/src/lib.rs
+++ b/stream-cipher/src/lib.rs
@@ -22,12 +22,13 @@ pub mod dev;
 
 mod errors;
 
-pub use crate::errors::{InvalidKeyNonceLength, LoopError};
+pub use errors::{InvalidKeyNonceLength, LoopError, OverflowError};
 pub use generic_array::{self, typenum::consts};
 
 #[cfg(feature = "block-cipher")]
 pub use block_cipher;
 
+use core::convert::{TryFrom, TryInto};
 use generic_array::typenum::Unsigned;
 use generic_array::{ArrayLength, GenericArray};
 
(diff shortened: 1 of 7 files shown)


stream-cipher/src/lib.rs
```diff
@@ -97,10 +97,21 @@ pub trait SyncStreamCipher {
 /// Synchronous stream cipher seeking trait.
 pub trait SyncStreamCipherSeek {
     /// Return current position of a keystream in bytes from the beginning.
-    fn current_pos(&self) -> u64;
+    fn get_current_pos(&self) -> u64;
```
tarcieri, line 100: Why rename this?

newpavlov: I was thinking about adding `try_*` variant for it and `try_current_pos` was a bit strange for my taste. If we'll decide that it's not needed, I probably will revert this renaming.

newpavlov:
@tarcieri 
I've made the seek trait generic over numeric types. Can you take a look? The approach assumes that cipher implementations have two counters: one for blocks and one for cursor position inside a cached block. The chain of `TryFrom/TryInto` bounds on `SeekNum` is a bit ugly, but it was the easiest solution I could find which keeps implementations code clean.

stream-cipher/src/lib.rs
```diff
@@ -185,3 +210,57 @@ where
...
+    + TryInto<u128>
+    + TryFrom<u128>
+{
+    /// Try to get position for block number `block`, byte position inside
+    /// block `byte`, and block size `bs`.
+    fn from_block_byte<T: SeekNum>(block: T, byte: u8, bs: u8) -> Result<Self, OverflowError>;
+
+    /// Try to get block number and bytes position for given block size `bs`.
+    fn to_block_byte<T: SeekNum>(self, bs: u8) -> Result<(T, u8), OverflowError>;
+}
```
tarcieri, line 241: Hmmmm... 😉

My gut reaction to this is there's a lot of complexity here which feels like there's not a valid use case to justify it.

I'm trying to think of cases where it'd be helpful to have something other than a 32-bit or 64-bit number.

In terms of block counts, why would you ever want to use `u8` or `u128`, for example?

I guess it's complete/consistent.

stream-cipher/src/lib.rs
```diff
@@ -94,13 +95,38 @@ pub trait SyncStreamCipher {
...
+    ///
+    /// Returns [`LoopError`] if provided position value is bigger than
+    /// keystream leangth
+    fn try_seek<T: SeekNum>(&mut self, pos: T) -> Result<(), LoopError>;
+
+    /// Get current keystream position
+    ///
+    /// # Panics
+    /// If position can not be represented by type `T`
+    fn get_pos<T: SeekNum>(&self) -> T {
```
tarcieri, line 119: Bikeshed, but: I'd prefer `current_pos` or `pos` here. [It's generally considered not idiomatic to name getter methods `get_*` in Rust](https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter)

stream-cipher/src/lib.rs
newpavlov: I agree that it could be trimmed, but I left it for completeness/symmetricity sake.

>why would you ever want to use u8 or u128

Well, OpenSSL variant of the CTR mode uses 128-bit counters. :)

newpavlov:
I wonder if method for "seek from the current position" would be useful...

stream-cipher/src/lib.rs
newpavlov: Also it would be nice to use a more compact formatting (i.e. use several bounds per line), but I couldn't find a relevant `rustfmt` option for it.

tarcieri: Yeah, the variant of AES-CTR used by AES-CMAC-SIV also uses a 128-bit counter, but that's for determining the initial position.

Data volume limits for most ciphers alone would limit where you can safely seek in the keystream after that.

newpavlov: Yeah, but I also seen (private) code which constructs offset based on bits from several sources (i.e. something like 20 bits from `id1`, 30 bits from `id2`, etc.). It could be done by generating nonces instead, but oh well...

I've added signed integers for completeness sake, I think additional code should not noticeably influence compilation times and hopefully one day we will find a way how to make this code more compact.

newpavlov:
@tarcieri 
If you don't have any objections, I am going to merge this PR in the current form and migrate stream ciphers to v0.7.

PR #264, opened 2020-08-19
[WIP] Refactor `PublicKey` into `sec1::EncodedPoint`

tarcieri:
Previously the `PublicKey` was a "wire type" for SEC1 encoded curve points which by design didn't validate the point is on the curve (so as to avoid mandating a curve arithmetic implementation for the type to be useful).

However, this has a pretty big drawback, in that when we do have a curve arithmetic implementation, there are many fallible operations which could be infallible if we knew the point is valid a priori.

Additionally, "thanks" to `typenum`-based arithmetic, it has pretty atrocious trait bounds, which are needed anywhere we presently deal with a public key.

This commit moves this encoded point type to `sec1::EncodedPoint`, which both much more precisely describes what it represents, but also moves it out-of-the-way so we can add a better `PublicKey` type (e.g. one which is or wraps an `AffinePoint` in crates which have a curve arithmetic implementation).


tarcieri:
Opening this as a draft, both to get early review, and because it's a SemVer breaking change and I think there are still some non-breaking, purely additive changes that can be made to the `elliptic-curve` crate before we bump minor versions again.

elliptic-curve/src/sec1.rs
```diff
@@ -0,0 +1,510 @@
...
+    <C::ElementSize as Add>::Output: Add<U1>,
+    CompressedPointSize<C>: ArrayLength<u8>,
+    UncompressedPointSize<C>: ArrayLength<u8>,
+{
+    /// Compressed Weierstrass elliptic curve point
+    Compressed(CompressedPoint<C>),
+
+    /// Uncompressed Weierstrass elliptic curve point
+    Uncompressed(UncompressedPoint<C>),
+}
```
tarcieri, line 50: Something we could do here which would be a significant simplification is get rid of `CompressedPoint` and `UncompressedPoint` (and `CompressedPointSize` and `UncompressedPointSize`) entirely and have this type be a struct with a buffer which is large enough to contain either a compressed-or-uncompressed point (which can still be disambiguated by the leading tag byte).

I might take a stab at that as part of this PR.

tarcieri:
I pushed up an additional commit (7a3a0ce8a6ebb5c81e036facd782ceaee609dd66) which I think simplifies things greatly by eliminating the `CompressedPoint` and `UncompressedPoint` types.

fjarri:
Can't say I'm a big fan of a mutable `EncodedPoint`. I would prefer to have the correctness of its state to be enforced by the compiler.

Also, what would be the opposite of `compress()`? A `FromEncodedPoint` implementation for `AffinePoint`?

tarcieri:
> Can't say I'm a big fan of a `mutable` EncodedPoint.

I take it that's in regard to `compress()`?

> Also, what would be the opposite of `compress()`?

I'd also like to conditionally impl `decompress()` (it came up right away when people were trying to use `k256` and match test vectors), and yes, the impl could be bounded by a `FromEncodedPoint` impl on `Arithmetic::AffinePoint`, although we may need an additional `ToEncodedPoint` trait which takes a `compress: bool` parameter to replace the previous `Into` bounds.

fjarri:
> I take it that's in regard to `compress()`?

Yes, and in general for `EncodedPoint` to have two possible states. If the type conditions for an `enum` were too complicated, what about having two separate types (uncompressed/compressed encoded point)? This will also allow us to return a `GenericArray` from `to_bytes()` (no `alloc` needed).

tarcieri:
That's exactly how the current `CompressedPoint` and `UncompressedPoint` types presently work.

newpavlov:
I am not familiar enough with `elliptic-curve` codebase, so I can't really give a constructive feedback at the moment, but couldn't we solve the logic duplication problem by introducing newtype wrappers?

tarcieri:
As someone who's a fan of newtype wrappers, concretely I'm not sure how they'd help here.

Concretely I'd break the problem down here as either validating or constructing an encoded document:

- The parser validates the document and then stores it verbatim
- The serializer constructs the document

The previous implementation used separate types for when the document is compressed or uncompressed, and an enum to provide a common type for representing both forms. But really in pretty much all cases the enum is all anything ever interacted with.

However, despite the enum being pretty much the only thing that mattered, the complexity of the compressed-vs-uncompressed type representations was leaking out everywhere in the trait bounds.

tarcieri:
@fjarri I made `compress()` return a new point, and with that eliminated all `&mut` methods other than `zeroize()`.

I also added a bevy of tests which provide ~95% code coverage on `sec1.rs` (sidebar: it'd be nice to set up codecov on this repo). Even with that, this PR still manages to shave off ~90 LOC versus the previous implementation, which had no tests(!)

I ended up liking this enough that I marked this as "ready for review" as I think it has significantly higher code quality while also simplifying the trait bounds significantly. Earlier I said I thought we might want to hold off on breaking changes, but I changed my mind: this cleans up the trait bounds enough I think I'd like to move forward with it, along with defining a replacement public key type (e.g. associated type e.g. `AffinePoint` or newtype wrapper thereof)

fjarri:
> That's exactly how the current CompressedPoint and UncompressedPoint types presently work (prior to this PR).

Ah, that's right. So the main reason for the approach in this PR is the simplification of type bounds? I must say, I still prefer the solution with two separate types, especially since `decompress()` is not trivial. And if `UncompressedPublicKey`/`CompressedPublicKey` were separate types and not enum variants, perhaps the bounds could have been kept simple? 

But anyway, I do not feel strongly enough about this to insist. It's your call.

> @fjarri I made compress() return a new point, and with that eliminated all &mut methods other than zeroize().

I suppose the only issue with this approach might be in some kind of an embedded environment where you really have to fight for every byte - but I don't have much experience in that area.

PR #280, opened 2020-08-31
elliptic-curve: remove `digest` crate path directive

tarcieri:
This makes it incompatible with `signature` when fetched via git, which does not have such a declaration (and cannot, as it'd break the dev-dependency on `sha2`).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -128,7 +128,7 @@ name = "elliptic-curve"
 version = "0.5.0"
 dependencies = [
  "const-oid",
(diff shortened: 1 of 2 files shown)


newpavlov:
Maybe we also should remove `path` from other crates as well? For example, it's a bit annoying to have to update `cryptography` after bumping minor version of another trait crate.

tarcieri:
Normally I like it, but it precludes these sort of circular dev-dependencies. Regardless, perhaps we should be consistent whatever we do.

PR #281, opened 2020-09-02
elliptic-curve: import arithmetic helper functions

tarcieri:
These are helper functions for performing arithmetic on field element "limbs", e.g. implementing carry chains.

The particular versions in this file come from the `k256`, which started as a fork of the `p256` crate, and notably provide 32-bit and 64-bit versions of the same functions.

These are potentially helpful to have in the `elliptic-curve` crate for several reasons, most notably for testing things which are generic across elliptic curves without having to provide a full curve arithmetic implementation.

Also having things in one place is nice, and potentially we can use these to provide a generic baseline implementation of certain types of curve arithmetic for curves that don't provide their own arithmetic.

cc @str4d who I had talked to about this at one point.

### DIFF
--- a/elliptic-curve/src/lib.rs
+++ b/elliptic-curve/src/lib.rs
@@ -30,6 +30,7 @@ pub mod ops;
 pub mod point;
 pub mod scalar;
 pub mod secret_key;
+pub mod util;
 
 #[cfg(feature = "ecdh")]
 #[cfg_attr(docsrs, doc(cfg(feature = "ecdh")))]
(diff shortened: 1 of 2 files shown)


tarcieri:
> most notably for testing things which are generic across elliptic curves 

Note: I specifically intend to use these to test the generic RFC6979 implementation. The spec mandates that all scalars are reduced, but also performs rejection sampling on HMAC_DRBG outputs, so testing it requires the ability to ensure that a particular output doesn't overflow the curve's order.

PR #285, opened 2020-09-03
elliptic-curve: vendor core `group` traits; add `ff` dependency

tarcieri:
Vendors the core set of group traits from the `group` crate:

https://github.com/zkcrypto/group

Using the upstream crate is presently blocked on `no_std` support:

https://github.com/zkcrypto/group/pull/9

However, aside from a suggested change to the RNG API (zkcrypto/group#10) the traits used in this PR are verbatim from the upstream crate, so migrating to it should be easy.

This PR also removes the `rand` and `weierstrass` features from the `elliptic-curve` crate, making `rand_core` a hard requirement, since it seems at least `rand_core` will probably be a hard requirement for `group`, and requiring an RNG as a hard dependency makes sense.

cc @str4d @ebfull

### DIFF
--- a/elliptic-curve/src/group.rs
+++ b/elliptic-curve/src/group.rs
@@ -1,0 +1,114 @@
+//! Algebraic group traits vendored from the `group` crate
+
+// Copyrights in the "group" library are retained by their contributors. No
+// copyright assignment is required to contribute to the "group" library.
+//
+// The "group" library is licensed under either of
+//
+// * Apache License, Version 2.0, (see ./LICENSE-APACHE or http://www.apache.org/licenses/LICENSE-2.0)
+// * MIT license (see ./LICENSE-MIT or http://opensource.org/licenses/MIT)
+//
+// at your option.
+//
+// Unless you explicitly state otherwise, any contribution intentionally
+// submitted for inclusion in the work by you, as defined in the Apache-2.0
+// license, shall be dual licensed as above, without any additional terms or
+// conditions.
+
+// TODO(tarcieri): switch to the upstream crate. Blocked on `no_std` support:
+// <https://github.com/zkcrypto/group/pull/9>
+
+use core::fmt;
+use core::iter::Sum;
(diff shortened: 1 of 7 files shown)


elliptic-curve/src/group.rs
```diff
@@ -0,0 +1,114 @@
...
+    + Sum
+    + for<'a> Sum<&'a Self>
+    + Neg<Output = Self>
+    + GroupOps
+    + GroupOpsOwned
+    + ScalarMul<<Self as Group>::Scalar>
+    + ScalarMulOwned<<Self as Group>::Scalar>
+{
+    /// Scalars modulo the order of this group's scalar field.
+    type Scalar: PrimeField;
```
fjarri, line 73: I feel like `Field` would be more accurate here; or, if you have `PrimeField`, the trait should be called `PrimeOrderGroup`. But since this is copied from `group` (and will eventually be replaced by its use), I should probably raise this question there.

tarcieri: As it were, [`PrimeField`](https://docs.rs/ff/0.7.0/ff/trait.PrimeField.html) is from the [`ff`](https://github.com/zkcrypto/ff) crate (which also has a [`Field`](https://docs.rs/ff/0.7.0/ff/trait.Field.html) trait).

tarcieri:
Looks like the upstream PR got merged, so I can try using the upstream crate instead: https://github.com/zkcrypto/group/pull/9

elliptic-curve/src/group.rs
fjarri: I'm just saying, it seems logical to me to have `Group::Scalar: Field` and `PrimeGroup::Scalar: PrimeField`, but evidently the authors of `group` followed some different logic, perhaps better suited for ECC specifics.

tarcieri: Aah, I see what you're saying.

I can't speak to why it is the way it is either, but if you're curious, here's `CofactorGroup`: https://github.com/zkcrypto/group/blob/master/src/cofactor.rs

PR #288, opened 2020-09-04
elliptic-curve: rename `ElementBytes` => `FEBytes`

tarcieri:
Some type name bikeshedding...

It's not immediately clear that "Element" in `ElementBytes` refers to a field element.

This is a shorter alternative, although it uses an acronym for field element and therefore may be less clear.


tarcieri:
Going to leave this one open a bit to gauge what people think.

elliptic-curve/src/lib.rs
```diff
@@ -62,8 +62,9 @@ use generic_array::{typenum::Unsigned, ArrayLength, GenericArray};
 use rand_core::{CryptoRng, RngCore};
 use subtle::{ConditionallySelectable, ConstantTimeEq, CtOption};
 
-/// Byte array containing a serialized scalar value (i.e. an integer)
-pub type ElementBytes<C> = GenericArray<u8, <C as Curve>::FieldSize>;
+/// Byte array containing a serialized field element, i.e. an element
+/// of the given curve's base or scalar fields.
+pub type FEBytes<C> = GenericArray<u8, <C as Curve>::FieldSize>;
```
tarcieri, line 67: @str4d I've noticed the `ff` and `group` crates use a "Repr" naming convention here... do you think something like `FERepr` would be better?

tarcieri:
Closing in favor of #296

PR #291, opened 2020-09-08
elliptic-curve: add `ff` and `group` bounds on Arithmetic trait

tarcieri:
Adds the following:

- `Arithmetic::ProjectivePoint` associated type with `group::Curve` and `group::Group` bounds
- `Arithmetic::Scalar` is now bounded by `ff::PrimeField`


tarcieri:
NOTE: I think there are a lot of traits that can be cleaned up and consolidated once these bounds are in place.

I consider this commit an incremental step to start moving over to the `ff` and `group` traits.

PR #300, opened 2020-09-10
elliptic-curve: refactor ProjectiveArithmetic trait

tarcieri:
Renames the `Arithmetic` trait to `ProjectiveArithmetic` and changes it to have only a single associated type: `ProjectivePoint`.

Adds `AffinePoint<C>` and `Scalar<C>` type aliases which are able to look up the corresponding associated type chain (along with `ProjectivePoint<C>` for consistency, although this is available as `<C as ProjectiveArithmetic>::ProjectivePoint`).

### DIFF
--- a/elliptic-curve/src/ecdh.rs
+++ b/elliptic-curve/src/ecdh.rs
@@ -1,7 +1,7 @@
 //! Elliptic Curve Diffie-Hellman (Ephemeral) Support.
 //!
 //! This module contains a generic ECDH implementation which is usable with
-//! any elliptic curve which implements the [`Arithmetic`] trait (presently
+//! any elliptic curve which implements the [`ProjectiveArithmetic`] trait (presently
 //! the `k256` and `p256` crates)
 //!
 //! # Usage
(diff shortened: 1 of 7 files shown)


tarcieri:
Marking this as ready for review.

This was a pretty painful PR that introduced some ugly trait bounds, but I was able to remove several superfluous traits from the `elliptic-curve` crate (including `Arithmetic`, `FromFieldBytes`, and `Generate`), leaning pretty much entirely on the ones in the `ff` and `group` crates.

I've gone ahead and opened corresponding draft PRs that integrate these changes into downstream crates:
- https://github.com/RustCrypto/signatures/pull/164
- https://github.com/RustCrypto/elliptic-curves/pull/179

I think this PR is a decent intermediate step towards an `ff`/`group` migration. My main concerns with it are:

- Some ugly trait bounds
- Some operations that were previously explicitly constant time aren't anymore. This mostly relates to various kinds of "repr" parsing, which is something I'd like to circle back on in terms of better integrating the `ff`/`group` serialization traits.

Regarding the ugly trait bounds, I opened one issue on `ff` with a small suggestion:

https://github.com/zkcrypto/ff/pull/43

PR #309, opened 2020-09-21
elliptic-curve: fix `sec1::EncodedPoint::decompress`

tarcieri:
This method was previously expecting `Decompress` and `ToEncodedPoint` impls on `Scalar`, rather than `AffinePoint`.

This updates it to use the correct types.

### DIFF
--- a/elliptic-curve/src/sec1.rs
+++ b/elliptic-curve/src/sec1.rs
@@ -21,14 +21,14 @@ use alloc::boxed::Box;
 
 #[cfg(feature = "arithmetic")]
 use crate::{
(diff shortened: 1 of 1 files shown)


tarcieri:
This is technically a breaking change as it alters the trait bounds, however the original implementation was buggy, so there's nothing to "break" here.

I would suggest yanking `elliptic-curve` v0.6.0 and publishing another release with this fix, once it's confirmed it addresses the problem.

tarcieri:
Note: this method now has downstream tests in the `k256` and `p256` crates: https://github.com/RustCrypto/elliptic-curves/pull/194

PR #317, opened 2020-09-28
cryptography: add `std` feature

tarcieri:
Adds an off-by-default `std` feature which activates the `std` features of all of the other dependencies.

This is a somewhat lousy way of doing this, and I wish that there were a better approach, so much that I posted a Pre-Pre-RFC about it:

https://internals.rust-lang.org/t/pre-pre-rfc-weak-cargo-feature-activation/13141

Edit: seems there's an open issue about this already:

https://github.com/rust-lang/cargo/issues/3494

For now though, this is immediately useful to me.


newpavlov (approved):
Yeah, something like weak features would be really great to have.

Issue #304, opened 2020-09-13
Async traits


tarcieri:
Several crates in this repo could benefit from having `async` equivalents. There are several use cases for `async`, such as communicating with network services which implement a particular cryptographic primitive (Cloud KMS, NetHSM), or things like cryptographic tokens, SmartCards, non-networked HSMs (e.g. PCI cards, USB devices), or embedded cryptographic accelerators.

Unfortunately Rust doesn't natively support async traits yet. The most commonly used solution at the moment is @dtolnay's [async-trait](https://github.com/dtolnay/async-trait) crate. This provides an ergonomic syntax which should hopefully be easy to transition to hypothetical future natively supported async traits, however at the moment it requires `std` and uses `Box` to store `Future`s. There's some [discussion about improving this happening on rust-internals right now](https://internals.rust-lang.org/t/async-traits-the-less-dynamic-allocations-edition/13048).

I think it'd be useful to start defining some async traits, even if they presently depend on `std`. Here's a tracking list of crates that could benefit from async traits:

- [ ] `signature`: communicating with Cloud KMS/NetHSM services

tarcieri:
I would really love to find *some* solution for allowing an embedded cryptographic accelerator to process work in the background while the primary processor/MCU is performing other operations.

Down-the-road I think `async` would be great here: you could submit work asynchronously to a cryptographic accelerator, receiving a `Future` you can poll for completion. Unfortunately `async-trait` seems like a nonstarter in the embedded space, so perhaps it's worth investigating some sort of stopgap API which lets you submit work and then poll/busywait for completion.

This would be quite helpful for algorithms like AES-GCM and AES-GCM-SIV (decryption, at least), by allowing encryption to happen in parallel on an accelerator/coprocessor while authentication is handled on the main processor/MCU.

Dirbaio:
In nightly it is already possible to have async traits with `feature(generic_associated_types)`. ([example](https://github.com/akiles/nrf-softdevice/blob/52329d22febd6f19a63ab398de1dbc2a228ee3d9/async-flash/src/lib.rs#L17)). This does fully static dispatch and no heap allocations. You can even implement the traits with `async` blocks (instead of manually-written futures) with `feature(type_alias_impl_trait)` ([example](https://github.com/akiles/nrf-softdevice/blob/52329d22febd6f19a63ab398de1dbc2a228ee3d9/nrf-softdevice/src/flash.rs#L31))

[...]

PR #320, opened 2020-10-05
async-signature: initial crate

tarcieri:
Adds an `async-signature` crate with `AsyncSigner` and `AsyncDigestSigner` traits, based on the `async-trait` crate.

Though we'd discussed not using an `async-trait`-based approach (#304), there are some immediate needs for these traits for `std` users, namely providing a common `async` API to things like Cloud KMS services and NetHSMs.

I think we can still pursue nightly-only versions of these traits in the `signature` crate itself (ones which will work in `no_std` environments).

In the meantime, these traits are self-contained within their own, relatively small crate to unblock making some initial progress with things like Cloud KMS abstractions.

### DIFF
--- a/signature/async/src/lib.rs
+++ b/signature/async/src/lib.rs
@@ -1,0 +1,93 @@
+//! RustCrypto: `async-signature` crate.
+//!
+//! This is an experimental crate containing `async` versions of select traits
+//! from the [`signature`] crate, namely [`AsyncSigner`] and when the `digest`
+//! feature is enabled, [`AsyncDigestSigner`].
+//!
+//! Traits are implemented using [`async-trait`], which rewrites the traits to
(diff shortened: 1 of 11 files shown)


signature/async/src/lib.rs
```diff
@@ -0,0 +1,93 @@
+//! RustCrypto: `async-signature` crate.
+//!
+//! This is an experimental crate containing `async` versions of select traits
+//! from the [`signature`] crate, namely [`AsyncSigner`] and when the `digest`
+//! feature is enabled, [`AsyncDigestSigner`].
```
tarcieri, line 5: Note: this crate is presently only providing signing traits.

I didn't see the immediate use case for async signature verification, but if someone has one I can add `AsyncVerifier`/`AsyncDigestVerifier`.

PR #333, opened 2020-10-14
stream-cipher: remove NewBlockCipher bound on FromBlockCipher

tarcieri:
Note: breaking change.

As far as I can tell this was an oversight: `FromBlockCipher` never needs to instantiate a new block cipher (isn't that the point?), making this bound needlessly limiting.

The rationale for it in the first place appears to be for the blanket impl of `NewStreamCipher` for `C: FromBlockCipher` which needs it, but we can add the bound to the blanket impl, rather than (needlessly) sticking it on the trait.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -124,7 +124,7 @@ dependencies = [
 
 [[package]]
 name = "cryptography"
-version = "0.4.1"
+version = "0.5.0-pre"
 dependencies = [
  "aead",
  "block-cipher 0.8.0",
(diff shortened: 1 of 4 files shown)


tarcieri:
@newpavlov I'd like to merge this for now as it's badly blocking https://github.com/RustCrypto/stream-ciphers/pull/170 and more specifically migrating `aes-gcm` and `aes-gcm-siv` to use it (since it has a "viral" `NewBlockCipher` bound, which prevents using the `impl<B: BlockCipher> BlockCipher for &B` blanket impl).

Will wait for your input before cutting another release, though. Worst case we can revert it.

newpavlov:
Looks good! I think it was indeed an oversight on my part.

PR #334, opened 2020-10-14
stream-cipher: add FromBlockCipherMut trait

tarcieri:
Support for initializing a stream cipher from a [`BlockCipherMut`](https://docs.rs/block-cipher/0.8.0/block_cipher/trait.BlockCipherMut.html).

This is a corollary to [`FromBlockCipher`](https://docs.rs/stream-cipher/0.7.1/stream_cipher/trait.FromBlockCipher.html), but for cases where `&mut` access is needed to the inner block cipher instance (e.g. cryptographic accelerator hardware peripherals).

Using chained blanket impls, it's possible to allow a blanket impl of `NewStreamCipher` for types which impl either `FromBlockCipherMut` or `FromBlockCipher`.

### DIFF
--- a/stream-cipher/src/lib.rs
+++ b/stream-cipher/src/lib.rs
@@ -35,7 +35,7 @@ use generic_array::typenum::Unsigned;
 use generic_array::{ArrayLength, GenericArray};
 
 #[cfg(feature = "block-cipher")]
-use block_cipher::{BlockCipher, NewBlockCipher};
+use block_cipher::{BlockCipher, BlockCipherMut, NewBlockCipher};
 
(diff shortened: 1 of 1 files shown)


tarcieri:
Want to try this out with `Ctr32BE` and `Ctr32LE`, adding `Ctr32BEMut` and `Ctr32LEMut` wrappers, which will finally enable wiring up AES-GCM with a hardware cryptographic accelerator that needs `&mut` access.

@newpavlov will make sure to get your review/approval before cutting a (non-pre)release with this change. We can always revert it.

PR #337, opened 2020-10-15
Unify `block-cipher` and `stream-cipher` into `cipher`

tarcieri:
We are now owners of the `cipher` crate:

https://crates.io/crates/cipher

This commit unifies the existing `block-cipher` and `stream-cipher` crates into the `cipher` crate, placing their code under the `cipher::block` and `cipher::stream` modules respectively.

Ideally this means the crate is a drop-in replacement, and all end users need to do is `s/block_cipher/cipher::block/` and
`s/stream_cipher/cipher::stream` to upgrade.


tarcieri:
@newpavlov also if this looks good to you, can you add @RustCrypto/traits as owner of the `cipher` crate?

newpavlov (approved):
>if this looks good to you, can you add @RustCrypto/traits as owner of the cipher crate?

Done!

tarcieri:
Awesome, thanks!

PR #350, opened 2020-10-30
cipher: revamp macro names

tarcieri:
The old macro names come from the `block-cipher` and `stream-cipher` crates, where the crate names disambiguated them.

However, now that they've been merged into the same crate, and being re-exported from the toplevel as macros are, it's now unclear whether a macro is for a block cipher or a stream cipher.

This commit adds `block_cipher*` and `stream_cipher*` to the start of each macro name to disambiguate them.

It also maintains the existing macros as wrappers for the new names for backwards compatibility, but with a `#[deprecated]` attribute added.

### DIFF
--- a/cipher/README.md
+++ b/cipher/README.md
@@ -10,7 +10,7 @@
 Traits which define the functionality of [block ciphers] and [stream ciphers].
 
 See [RustCrypto/block-ciphers] and [RustCrypto/stream-ciphers] for algorithm
-implementations which use this trait.
+implementations which use these traits.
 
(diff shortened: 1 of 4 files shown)


newpavlov (approved):
Looks good! Though I think these changes can wait until v0.3, but feel free to go ahead if you want to update the dependent crates now.

tarcieri:
@newpavlov I'd like to hold off on a v0.3 release for awhile as we've had quite a few releases lately, but yeah, it'd be nice to get the dependent crates updated in advance

Issue #332, opened 2020-10-13
cipher: block cipher trait inefficiencies


jack-signal:
tl;dr The block cipher trait exposes parallelism (this is good!). The block cipher trait only exposes parallelism for specific widths (this is bad!)

The aes-soft and aesni crates both implement an 8x wide AES operation, due to bitslicing and pipelining respectively. However the interface mandates that you batch process in exactly the width of the underlying implementation. This means that if you have say 4 or 6 blocks which could be processed at once, your options are to either process the blocks serially, or to set up some extra dummy blocks which are encrypted and then thrown away. It turns out, at least for aes-soft on my machine, always processing 8 blocks and using dummy inputs is faster, even when processing just 2 blocks.

This design also restricts the possible implementation approaches. There is no reason that, for example, AES-NI couldn't have several loops, one unrolled 8x and another 2x, followed by a final 1x to handle a trailing block. But since callers will only ever provide input in multiple of 8 (or else between 1 and 7 serial blocks) there is no possibility for intermediate unrolling.

In theory everyone could just perform this batching in higher level crates which use this trait. In practice effectively nobody will, and as a result everything built on the block cipher traits is not as efficient as it otherwise could be. The trait should instead accept any number of blocks, and process them as efficiently as is possible, along with advertising the preferred parallelism which would allow higher level algorithms to tune their buffer sizes properly.

tarcieri:
Yes, @newpavlov and I have discussed this and generally agree with you, I think. See also: #289 

It would be more efficient on modern Intel architectures to process blocks 4-at-a-time.

One general reason which requires implementations to specify the granularity of parallelism is it may be hardcoded directly into the implementation. This is particularly true of AVX2/AVX512 implementations, especially when combined with a universal hash.

> The trait should instead accept any number of blocks, and process them as efficiently as is possible

In the `universal-hash` crate we have the [`update_padded` method](https://docs.rs/universal-hash/0.4.0/universal_hash/trait.UniversalHash.html#method.update_padded) for this, but a similar method for `block-cipher` which works on a slice of blocks sounds like a reasonable idea.

I think there still needs to be some sort of associated type for the ideal buffer size for the number of blocks to act on in parallel, but there could be a simple method with a default implementation which can act on arbitrary numbers of blocks and map them to underlying parallel buffer processing.

tarcieri:
@jack-signal take a look at #351 and see if it addresses some of your concerns

tarcieri:
[...]

PR #351, opened 2020-10-30
cipher: add BlockCipher::{encrypt_slice, decrypt_slice}

tarcieri:
Partially addresses #332.

Adds methods which work over an arbitrarily sized slice of blocks to encrypt/decrypt.

Provides a default implementation, but can be potentially further optimized by individual implementations.

### DIFF
--- a/cipher/src/block.rs
+++ b/cipher/src/block.rs
@@ -20,6 +20,7 @@ pub use errors::InvalidKeyLength;
 // TODO(tarcieri): remove these re-exports in favor of the toplevel one
 pub use generic_array::{self, typenum::consts};
 
+use core::convert::TryInto;
 use generic_array::{typenum::Unsigned, ArrayLength, GenericArray};
 
 /// Key for an algorithm that implements [`NewBlockCipher`].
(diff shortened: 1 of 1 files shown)


cipher/src/block.rs
```diff
@@ -79,6 +80,20 @@ pub trait BlockCipher {
         }
     }
 
+    /// Encrypt a slice of blocks, leveraging parallelism when available.
+    #[inline]
+    fn encrypt_slice(&self, blocks: &mut [Block<Self>]) {
+        for chunk in blocks.chunks_mut(Self::ParBlocks::to_usize()) {
```
newpavlov, line 86: I think the following code should be a bit more efficient:
```rust
let pb = Self::ParBlocks::to_usize();
if pb > 1 {
    let mut iter = blocks.chunks_exact_mut(pb);
    for chunk in &mut iter {
        self.encrypt_blocks(chunk.try_into().unwrap())
    }
    blocks = iter.into_remainder();
}
for block in blocks {
    self.encrypt_block(block);
}
```

tarcieri: @newpavlov updated the PR to use your suggested implementation.

Agree that ideally the slice-based methods would be called `encrypt_blocks`/`decrypt_blocks`. We can cross that bridge when we get there, I guess.

PR #352, opened 2020-11-01
cipher: add encrypt/decrypt-only block cipher traits

### LINKED ISSUE #349: cipher: traits for encrypt-only (and decrypt-only) block ciphers?
All of the AES-based AEADs we've implemented are based on AES-CTR (as implemented in various forms in the `ctr` crate).

CTR notably only needs the encryption component of AES, but right now if we use `aes-soft` in combination with `ctr`, we still pay an eager key schedule setup penalty for decryption, even if that code is never used (I think... it'd be nice if if LLVM were smart enough to elide it that'd, and also reduce code size).

Perhaps we should have traits like `crypto::block::Encrypt` and `crypto::block::Decrypt`, which could be combined into `BlockCipher`? For example, `BlockCipher` could have a blanket impls for both `Encrypt` and `Decrypt`, so you can use a `BlockCipher` anywhere that `Encrypt` or `Decrypt` are required, and we could find ways to make it easy to construct a `BlockCipher` out of separate `Encrypt` and `Decrypt` impls for cases like `aes-soft` where the two are really independent.

I think ideally in cases where we're using AES-CTR, we can avoid the key schedule setup penalty for decryption, and ideally all of the decryption-related code as well.

tarcieri:
Closes #349

Adds `BlockEncrypt` and `BlockDecrypt` traits which can be used in cases where e.g. only the encryption portion of a block cipher is used, as in CTR mode.

This PR does not otherwise attempt to add things like a blanket impl.

### DIFF
--- a/cipher/src/block.rs
+++ b/cipher/src/block.rs
@@ -186,11 +186,15 @@ impl<Alg: BlockCipher> BlockCipher for &Alg {
     }
 }
 
-/// Encrypt-only functionality for block ciphers
-pub trait Encrypt {
+/// Marker trait for block size
+// TODO(tarcieri): rename this to `BlockCipher` in the next breaking release
+pub trait BlockSizeMarker {
     /// Size of the block in bytes
     type BlockSize: ArrayLength<u8>;
+}
 
+/// Encrypt-only functionality for block ciphers
+pub trait BlockEncrypt: BlockSizeMarker {
     /// Number of blocks which can be processed in parallel by
     /// cipher implementation
     type ParBlocks: ArrayLength<GenericArray<u8, Self::BlockSize>>;
@@ -234,10 +238,7 @@ pub trait Encrypt {
 }
 
 /// Decrypt-only functionality for block ciphers
-pub trait Decrypt {
-    /// Size of the block in bytes
-    type BlockSize: ArrayLength<u8>;
-
+pub trait BlockDecrypt: BlockSizeMarker {
     /// Number of blocks which can be processed in parallel by
     /// cipher implementation
     type ParBlocks: ArrayLength<GenericArray<u8, Self::BlockSize>>;
--- a/cipher/src/block.rs
+++ b/cipher/src/block.rs
@@ -53,8 +53,7 @@ pub trait NewBlockCipher: Sized {
     }
 }
 
-/// The trait which defines in-place encryption and decryption
-/// over single block or several blocks in parallel.
+/// Trait which marks a type as being a block cipher.
 pub trait BlockCipher {
     /// Size of the block in bytes
     type BlockSize: ArrayLength<u8>;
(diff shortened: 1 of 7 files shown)


tarcieri:
Note: I used the new suggested method names here: `*_block`, `*_par_blocks`, `*_blocks`.

Perhaps it would make more sense to leave them the same for consistency.

cipher/src/block.rs
```diff
@@ -185,3 +185,97 @@ impl<Alg: BlockCipher> BlockCipher for &Alg {
...
+            for chunk in &mut iter {
+                self.encrypt_par_blocks(chunk.try_into().unwrap())
+            }
+
+            blocks = iter.into_remainder();
+        }
+
+        for block in blocks {
+            self.encrypt_block(block);
+        }
```
tarcieri, line 232: This is copypasta from the `BlockCipher` trait.

It may make more sense to extract some sort of common method which encapsulates this behavior.

cipher/src/block.rs
```diff
@@ -185,3 +185,97 @@ impl<Alg: BlockCipher> BlockCipher for &Alg {
...
+            for chunk in &mut iter {
+                self.decrypt_blocks(chunk.try_into().unwrap())
+            }
+
+            blocks = iter.into_remainder();
+        }
+
+        for block in blocks {
+            self.decrypt_block(block);
+        }
```
tarcieri, line 279: More copypasta here.

tarcieri:
After a few attempts at eliminating the redundancy, I think it might make more sense to transform the `BlockCipher` trait (via a breaking change) into a marker trait for the case where a type impls both `Encrypt` and `Decrypt` with compatible `BlockSize`/`ParBlocks`.

newpavlov (commented):
Aren't `Encrypt`/`Decrypt` too generic? It could be worth to use something like `BlockEncrypt` instead of `block::Encrypt`. Such naming would also allow us to simplify the crate structure by removing `stream` and `block` modules.

cipher/src/block.rs
```diff
@@ -185,3 +185,97 @@ impl<Alg: BlockCipher> BlockCipher for &Alg {
         Alg::decrypt_blocks(self, blocks);
     }
 }
+
+/// Encrypt-only functionality for block ciphers
+pub trait Encrypt {
+    /// Size of the block in bytes
+    type BlockSize: ArrayLength<u8>;
```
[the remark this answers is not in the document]
tarcieri: Yeah, that sounds good

tarcieri:
>  It could be worth to use something like BlockEncrypt instead of block::Encrypt. Such naming would also allow us to simplify the crate structure by removing stream and block modules.

That's a potential alternative, yes. I think it makes sense for the code to still be factored into two modules, but we could re-export everything at the toplevel rather than using modules.

tarcieri:
In 5096a41 I renamed the traits to `BlockEncrypt` and `BlockDecrypt`, and added a `BlockSizeMarker` trait for carrying the `BlockSize` associated type/constant.

newpavlov:
>I think it makes sense for the code to still be factored into two modules

I agree, I meant how API will look on outside.

tarcieri:
@newpavlov does this look ok now, or would you rather hold off until `cipher` v0.3?

newpavlov:
I think it will be better to release these changes in v0.3.

tarcieri:
@newpavlov I've gone ahead and made the discussed breaking changes in 2c61ca2, including splitting `BlockCipherMut` into `BlockEncryptMut` and `BlockDecryptMut` which require the given type impls `BlockCipher`.

This had the added benefit of eliminating the need for a `stream::FromBlockCipherMut` trait.

PR #353, opened 2020-11-01
[WIP] exploring a `BlockCipher` marker trait (not for merge)

tarcieri:
Attempt at refactoring `BlockCipher` into a trait which marks a type as being capable of both `Encrypt` and `Decrypt`.

As far as I can tell this requires the unfortunate change of making the block size and parallel block parameters generic.


tarcieri:
Closing this in favor of [@newpavlov's suggestion from #352](https://github.com/RustCrypto/traits/pull/352#discussion_r515645768).

PR #363, opened 2020-11-02
Made PublicKey a newtype that does point validation upon deserialization

rozbb:
This change, coupled with some changes in the `elliptic_curve` trait, should fix https://github.com/RustCrypto/elliptic-curves/issues/236

I know it's kinda messy, but this was the minimal solution I could come up with. It resembles `ecdsa::VerifyKey`

### DIFF
--- a/elliptic-curve/src/ecdh.rs
+++ b/elliptic-curve/src/ecdh.rs
@@ -24,11 +24,16 @@ use crate::{
     consts::U1,
     generic_array::ArrayLength,
     scalar::NonZeroScalar,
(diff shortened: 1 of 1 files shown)


tarcieri:
This looks good, but I'll hold of on merging for a bit as it's a breaking change.

I might consider making one in a couple weeks (when Rust 1.48 is out, which seems like a reasonable enough time to bump MSRV to 1.46).

tarcieri:
Decided to make a breaking change anyway (#365) so I'll go ahead and merge this. Might be a bit before I cut another release and update `k256`/`p256`/`p384` though.

PR #366, opened 2020-11-03
elliptic-curve: extract toplevel PublicKey; add functionality

tarcieri:
This PR refactors the `ecdh::PublicKey` type introduced in #363 into a toplevel `public_key::PublicKey` type which is generally useful anywhere a public key is required.

It additionally adds a number of additional impls including `AsRef`, `Deref`, and various `From`/`TryFrom` conversions.

It also promotes the `from_encoded_point` method into a proper impl of the `FromEncodedPoint` trait, leveraging a `TryFrom` impl instead as a more convenient API.

cc @rozbb

### DIFF
--- a/elliptic-curve/src/ecdh.rs
+++ b/elliptic-curve/src/ecdh.rs
@@ -22,85 +22,20 @@
 
 use crate::{
     consts::U1,
-    generic_array::ArrayLength,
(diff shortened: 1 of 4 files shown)


rozbb:
Looks good!

PR #369, opened 2020-11-11
Rename `cryptography` to `crypto`

tarcieri:
We are now the owners of the `crypto` crate! 🎉

My suggested use for it is a shorter name for `cryptography`.

I don't think there's a non-confusing way to use both `crypto` and `cryptography` otherwise, so IMO this is the best way to use it.


newpavlov (approved):
>I don't think there's a non-confusing way to use both `crypto` and `cryptography` otherwise

Well, one possible approach is to use `crypto` for "recommended" algorithms and `cryptography` for all implemented ones. :)

tarcieri:
Cool. We can revisit other usages of `cryptography` in the future.

Issue #262, opened 2020-08-18
digest: could VariableOutput be replaced entirely with FixedOutput?


tarcieri:
I believe the current [`digest::VariableOutput`](https://docs.rs/digest/0.9.0/digest/trait.VariableOutput.html) trait could be replaced entirely by [`FixedOutput`](https://docs.rs/digest/0.9.0/digest/trait.FixedOutput.html) and making the type which impls the trait generic around a type-level (`typenum`) parameter that defines the `OutputSize`.

Some precedent for this can be found in how the `AesGcm` crate is generic around a `NonceSize`:

https://github.com/RustCrypto/AEADs/blob/18a6128/aes-gcm/src/lib.rs#L169-L173

Digests which are only valid for specific sizes could define a marker trait for which sizes are considered valid, and then use that marker trait in the `where` bounds for when they impl `digest::FixedOutput`(Dirty).

This approach would make construction of a variable output digest instance infallible with output sizes checked by the type system, and also reduce the total number of traits.

newpavlov:
Yes, it could be done and IIRC it was the initial approach. But it has a significant drawback: you will not be able to select output size at runtime without creating a huge enum, though I am not sure if anyone needs this capability in practice. I guess we could use this approach and see if anyone complains.

tarcieri:
Aah yeah, that's a good point, although I think any time I would want to support variable sizes picked at runtime, there's a low cardinality of them

tarcieri:
@newpavlov it'd definitely be good to improve and/or replace `VariableOutput` before cutting a `digest` v0.10 is released.

As a semi-related concern, it's not possible to implement Argon2 using the `blake2` crate.

Argon2 requires a BLAKE2 output size of 1024-bytes, however the `blake2` crate currently caps it at 64.

It seems the reason for the cap is to allocate an output buffer in advance, but that got me wondering: if we did keep `VariableOutput` around, why can't you just provide it a buffer of your desired size to write into at output time, as opposed to having to choose one in advance?

newpavlov:
It's already removed in favor of fixed-size traits.

>Argon2 requires a BLAKE2(b) output size of 1024-bytes, however the blake2 crate currently caps it at 64.

IIRC BLAKE2(b) specification states that max output of the variable variant is 64 bytes. The 1024 byte variant is an Argon2 specific modification, which is built on top of the fixed 64-byte variant. I think it can be implemented directly in an argon2 crate, without exposing it in `blake2`.

[...]

PR #380, opened 2020-12-04
digest: add block-level API

newpavlov:
This PR exposes block-level API and adds a wrapper type which implements the higher-level traits. It allows to remove some of boilerplate from the implementation crates, plus we no longer need the macro for implementing `io::Read`. Also it will allow a more efficient HMAC implementation.

Additionally PR removes the variable length API in favor of making type generic over output length as suggested in #262.


newpavlov:
@tarcieri 
I wonder if we can remove the non-core dirty traits.

tarcieri (approved):
Seems like a nice improvement!

jplatte:
What's the status of this? It seems like it's the only major change holding back releases of digest and blake2 (https://github.com/RustCrypto/hashes/pull/198).

newpavlov:
@jplatte 
I wanted to implement block stream cipher trait for `XofReaderCore` to solve RustCrypto/hashes#181, but unfortunately I couldn't find a good block API for stream ciphers (see #444 for the most recent issue). I probably will simply omit it for now, so `digest v0.10` will not be blocked on release of `cipher v0.3`.

PR #381, opened 2020-12-04
elliptic-curve: PKCS#8 support for `SecretKey`

tarcieri:
Adds initial support for parsing PKCS#8-encoded ECC secret keys.

This is presently somewhat difficult to test within this crate (which should be addressed regardless) as it relies on generic parameters of a curve implementation, however this has been tested concretely against the `p256` crate which has tests that exercise it in this PR:

https://github.com/RustCrypto/elliptic-curves/pull/243

### DIFF
--- a/elliptic-curve/src/lib.rs
+++ b/elliptic-curve/src/lib.rs
@@ -71,8 +71,8 @@ pub use group::{self, Group};
 #[cfg(feature = "digest")]
 pub use digest::{self, Digest};
 
-#[cfg(feature = "oid")]
-pub use oid;
+#[cfg(feature = "pkcs8")]
+pub use pkcs8;
 
 #[cfg(feature = "zeroize")]
 pub use secret_key::SecretKey;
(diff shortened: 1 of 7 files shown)


elliptic-curve/src/lib.rs
```diff
@@ -115,10 +122,15 @@ pub trait FromDigest<C: Curve> {
...
-pub trait Identifier: Curve {
+/// Associate an [`ObjectIdentifier`][`pkcs8::ObjectIdentifier`] (OID) with an
+/// elliptic curve algorithm implementation.
+///
+/// This is used as as the `parameters` of an `AlgorithmIdentifier` as
+/// described in RFC 5280 Section 4.1.1.2:
+/// <https://tools.ietf.org/html/rfc5280#section-4.1.1.2>
+#[cfg(feature = "pkcs8")]
+#[cfg_attr(docsrs, doc(cfg(feature = "pkcs8")))]
+pub trait AlgorithmParameters: Curve {
```
tarcieri, line 133: Not sure about this name. I chose it to distinguish it from `pkcs8::AlgorithmIdentifier`, and because this OID is ultimately stored in the `parameters` field of that struct, but while that appropriately names the field in the ASN.1 schema, the phrase `AlgorithmParameters` doesn't exactly scream "this is the trait you use to associate an OID with a curve implementation" to me.

PR #400, opened 2020-12-16
elliptic-curve: make `ToEncodedPoint` fallible

### LINKED ISSUE #399: elliptic-curve: ToEncodedPoint should be fallible
The point at infinity has no SEC1 bytestring representation. If an `AffinePoint` is the point at infinity, this method should fail. Currently, it will just output whatever garbage is in the x,y coordinates of the point.

https://github.com/RustCrypto/elliptic-curves/blob/bd7c9ea91a2e6bdb2f3fd71290708459b8aabc5f/p256/src/arithmetic/affine.rs#L150-L154

This might be difficult to fix, since `ToEncodedPoint::to_encoded_point` is an infallible method. Options are:

1. Make the point at infinity non-representable again in `AffinePoint` (i.e., remove the `infinity` field)
2. Change `ToEncodedPoint` to have an associated `ErrorType` which can be set to `!`
3. Ignore this and just make a big warning on the method

tarcieri:
Closes #399.

As discussed in the aforementioned issue, this changes `ToEncodedPoint` to be fallible, returning an `Option<EncodedPoint<C>>` with the expectation that `None` is returned for the additive identity
(a.k.a. point-at-infinity)

Brief background: originally it was not expected for `AffinePoint` to be able to encode the additive identity, however supporting it allowed for infallible conversions between `AffinePoint` <-> `ProjectivePoint`.

Technically, according to SEC1 Section 2.3.3, the identity can be represented by the `Elliptic-Curve-Point-to-Octet-String` encoding by using a single NULL byte (i.e. `[0u8]`).

However, this would have the disadvantage of complicating `EncodedPoint`, making APIs that are presently infallible fallible, and may be unexpected by end users of the traits.

If there's a legitimate use case for supporting SEC1-encoded identity points, we can revisit this tradeoff, adding support for encoding the identity element to `EncodedPoint`, and making `ToEncodedPoint` infallible once again.

cc @rozbb @fjarri

### DIFF
--- a/elliptic-curve/src/public_key.rs
+++ b/elliptic-curve/src/public_key.rs
@@ -152,7 +152,7 @@ where
     }
 }
 
-impl<C> From<PublicKey<C>> for EncodedPoint<C>
+impl<C> TryFrom<PublicKey<C>> for EncodedPoint<C>
 where
     C: Curve + ProjectiveArithmetic + point::Compression,
     FieldBytes<C>: From<Scalar<C>> + for<'r> From<&'r Scalar<C>>,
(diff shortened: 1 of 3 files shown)


elliptic-curve/src/public_key.rs
```diff
@@ -177,8 +179,10 @@ where
     UntaggedPointSize<C>: Add<U1> + ArrayLength<u8>,
     UncompressedPointSize<C>: ArrayLength<u8>,
 {
-    fn from(public_key: &PublicKey<C>) -> EncodedPoint<C> {
-        public_key.to_encoded_point(C::COMPRESS_POINTS)
+    type Error = Error;
+
+    fn try_from(public_key: &PublicKey<C>) -> Result<EncodedPoint<C>, Error> {
+        public_key.to_encoded_point(C::COMPRESS_POINTS).ok_or(Error)
     }
```
tarcieri, line 186: These conversions could potentially remain infallible if we had `PublicKey` enforce the invariant that the inner point cannot be the additive identity.

That would require making `PublicKey::from_affine` fallible, which might make sense, and would nicely match the similar invariants of `SecretKey`, which wraps a `NonZeroScalar`.

All that said, I'd like to follow up on that in a separate PR.

tarcieri:
Per continued discussion on #399, I think I'll retry this with a slightly different approach.

PR #416, opened 2020-12-18
Make PublicKey::from_encoded_point go through PublicKey::from_affine

### LINKED ISSUE #415: PublicKeys can be zero
Re the recent changes, I tried this code in `p256` and it failed (i.e., it didn't panic)

```rust
#[should_panic]
#[test]
fn zero_pubkey() {
    PublicKey::from_sec1_bytes(&[0]).unwrap();
}
```

This is because `PublicKey::from_encoded_point` doesn't use the identity-sensitive `PublicKey::from_affine` method. I'm sending a PR right now to fix this

rozbb:
Fixes #415

### DIFF
--- a/elliptic-curve/src/public_key.rs
+++ b/elliptic-curve/src/public_key.rs
@@ -208,7 +208,8 @@ where
 {
     /// Initialize [`PublicKey`] from an [`EncodedPoint`]
     fn from_encoded_point(encoded_point: &EncodedPoint<C>) -> Option<Self> {
-        AffinePoint::<C>::from_encoded_point(encoded_point).map(|point| Self { point })
+        AffinePoint::<C>::from_encoded_point(encoded_point)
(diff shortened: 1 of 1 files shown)


rozbb:
It's hard to add a regression test for this, because there are no concrete curves we can use to run a test

tarcieri:
It'd be fantastic if you could add a regression test for this as well

rozbb:
One way to allow for a regression test is to pull in a curve like `p256` just for this test. Might be overkill though, and it's a bit circular

tarcieri:
@rozbb yeah this is something of an ongoing problem... no worries!

I have a "mock curve" in the `ecdsa` crate but not in this one.

Perhaps I should move it to this crate (e.g. in a `dev` module) and then the `ecdsa` crate could reuse it.

PR #419, opened 2020-12-19
elliptic-curve: `dev` module with `MockCurve` type

tarcieri:
There is presently no good way to test traits and types in this crate which depend on a concrete curve implementation.

This commit adds a `dev` module gated on an eponymous crate feature which provides a `MockCurve` type. This type does not provide a full curve arithmetic implementation (most methods are stubbed out using `unimplemented!`), but rather implements the full set of traits used in the bounds of traits defined in the `elliptic-curve` crate.

This type is actually an extraction from the `ecdsa` trait, where similar testing problems were previously encountered. However, now that there's a `dev` module, the `ecdsa` crate can now use it!

### DIFF
--- a/.github/workflows/elliptic-curve.yml
+++ b/.github/workflows/elliptic-curve.yml
@@ -37,6 +37,7 @@ jobs:
           override: true
       - run: cargo build --no-default-features --release --target ${{ matrix.target }}
       - run: cargo build --no-default-features --release --target ${{ matrix.target }} --features arithmetic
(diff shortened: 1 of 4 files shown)


tarcieri:
See #416 for an example of where this would've been helpful. It wasn't possible to write a regression test for that bug because it would require a mock elliptic curve implementation.

newpavlov:
But you can import an elliptic curve implementation crate as a dev-dependency, no? Or do you think it will be an overkill?

tarcieri:
@newpavlov you can, but that creates a circular dependency

PR #394, opened 2020-12-13
cipher: integrate `block-modes`

newpavlov:
A number of changes for `cipher v0.3`.

Edit: many of these changes were already merged in #435 -@tarcieri

PR #435, opened 2020-12-26
cipher: v0.3 changes (without block-modes integration)

tarcieri:
This PR contains the proposed API changes from #394, but removes the WIP work to also integrate the `block-modes` crate.

The reason is this work is a blocker on getting downstream crates updated to the new API, and the `block-modes` integration work is otherwise purely additive so it can easily be done separately.


cipher/src/block.rs
```diff
@@ -32,13 +28,13 @@ pub trait NewBlockCipher: Sized {
...
     /// Create new block cipher instance from key with fixed size.
-    fn new(key: &Key<Self>) -> Self;
+    fn new(key: &BlockCipherKey<Self>) -> Self;
 
     /// Create new block cipher instance from key with variable size.
     ///
     /// Default implementation will accept only keys with length equal to
     /// `KeySize`, but some ciphers can accept range of key lengths.
-    fn new_varkey(key: &[u8]) -> Result<Self, InvalidLength> {
+    fn new_var(key: &[u8]) -> Result<Self, InvalidLength> {
```
tarcieri, line 37: As long as we're bikeshedding this method name, I've thought it would be clearer if it had `slice` in the name, e.g. `new_from_slice` or `from_slice`.

tarcieri:
Going to go ahead and land this. Hopefully that's ok!

newpavlov:
>Hopefully that's ok!

No problem! Sorry, I couldn't finish the PR myself.

cipher/src/block.rs
newpavlov: `new_from_slice` sounds good! `from_slice` is shorter, but I think will be more confusing, since it could be mistaken for a specialized version of the `From` trait. Also usually `from*` methods imply simple conversion, which is not the case here.

i wonder if for stream cipher and block modes we should use plural `new_from_slices`, since they accept two slices.

tarcieri: @newpavlov sounds good

PR #437, opened 2020-12-29
password-hash crate

### LINKED ISSUE #263: Crate for passwords hashing trait(s)
The main motivation for introducing this crate is to have a single implementation of MCF and PHC formats (see RustCrypto/password-hashes#34).

I've reserved `password-hash` for this purpose.

tarcieri:
Closes #263.

Adds a crate containing a (WIP!) no_std-friendly heapless/stack-allocated implementation of the PHC string format specification.

As the specification notes, it's designed with constraints which enable implementations based entirely on stack-allocated buffers. In order to maximize portability, this implementation relies entirely on stack allocated buffers.


tarcieri:
In broad strokes: I think the goal of this crate should be to subsume and unify  the `include_simple` features of the  https://github.com/RustCrypto/password-hashes crates.

Beyond that, I think each of these crates can continue to define static functions with their specific params, but a `password-hash` crate can provide a common type for representing parsed PHC strings and acting on them generically via TBD traits.

password-hash/src/ident.rs
```diff
@@ -0,0 +1,261 @@
...
+///
+/// [1]: https://github.com/P-H-C/phc-string-format/blob/master/phc-sf-spec.md
+#[derive(Copy, Clone, Eq, Hash, PartialEq, PartialOrd, Ord)]
+pub struct Ident {
+    /// Byte array containing an ASCII-encoded string.
+    bytes: [u8; MAX_LENGTH],
+
+    /// Length of the string in ASCII characters (i.e. bytes).
+    length: u8,
+}
```
tarcieri, line 48: Note: this implementation contains some fairly repetitive `arrayvec`/`tinyvec`/`heapless::Vec`-like types which would probably benefit from using some sort of fix-sized `Vec`-like abstraction.

The decision not to use any of these is deliberate: avoiding any provides a baseline for experimentation as to which, if any, we should use.

With `min_const_generics` on the horizon it seemed best not to be too eager to pick, or to loop in `generic-array` quite yet.

The needs for such a type are quite minimal (these types are mostly immutable/allocate-once) with only a fixed size overhead, so a reasonable first step in a `min_const_generics` feature is to define a type internal to the library for representing them, and if that turns out to be too cumbersome, considering a dependency.

password-hash/src/b64.rs
```diff
@@ -0,0 +1,321 @@
+//! "B64" encoding.
+//!
+//! Subset of the standard Base64 encoding (RFC 4648, section 4) which omits
+//! padding (`=`) as well as extra whitespace, as described in the PHC string
+//! format specification:
+//! <https://github.com/P-H-C/phc-string-format/blob/master/phc-sf-spec.md#b64>
+//!
+//! Supports the Base64 character subset: `[A-Z]`, `[a-z]`, `[0-9]`, `+`, `/`
```
tarcieri, line 8: @newpavlov I just noticed you own the `b64` crate.

It might make sense to split this functionality out into its own crate (which could be used, for example, to implement PEM encoding in the `pkcs8` crate)

tarcieri:
Marking this as ready for review. It's a bit involved for a first pass, but I think it provides quite a bit of shared functionality and interoperability support.

I wound up defining a `PasswordHashingFunction` trait (perhaps a bit long for a name, but certainly clear!) with a single `hash_password` function, so the touch points with algorithm crates is relatively small.

There's also a `PasswordHash` struct that maps to the PHC string format, and some generic inherent methods on it:

- `PasswordHash::generate`: create a new password hash using the supplied algorithm
- `PasswordHash::verify_password`: (perhaps could just be `::verify`) takes a slice of "allowed" `PasswordHashingFunction`s as a trait object, and attempts to use each of them to verify the password, returning `Ok(())` if one does.

As the above are inherent methods, you don't even need the `PasswordHashingFunction` trait in scope to use them, which is nice.

This PR includes a small (optional feature gated) algorithm registry, which includes Argon2, bcrypt, PBKDF2, scrypt, SHA-crypt, and MD5 crypt. It may be overkill, but password hashing algorithms are such a mess I think it's helpful to have some central organization and documentation.

There's still quite a bit that could be added, most notably a `rand_core` integration for generating cryptographically random salts, but at 3,600 lines I feel this PR is probably already enough to review (although I may circle back on RNG support, time permitting).

newpavlov (commented):
I am not sure I like the proposed approach. It's quite weird that a trait crate has knowledge about concrete algorithms. Maybe I am missing something, but I would expect algorithm names and parameters to be handled in implementation crates.

I have a very rough draft [here](https://gist.github.com/newpavlov/2b206a52a1af6f3340cfcf2293f7ac3a), which demonstrates an alternative approach. It has the core `PasswordHasher` trait and everything else is built on top of it (probably everything else should be feature-gated).

There are several open questions to which I don't have clear answers now:
- Can we assume that salt is always a binary b64 encoded string?
- Naming of convenience methods which would work with `String` and handle salt generation using `OsRng`/`ThreadRng`/custom RNG.
- Maybe we should reuse PHC buffer for decoding salt and hash? Encoded values are always bigger than original values, so if `PhcString` types will wrap `&mut [u8]` instead of `&[u8]`, we can decode values in place, thus removing the need for creation of temporary buffers and arbitrary restrictions on their length.
- How to make it easier support of several algorithms at once? I think a good solution could be to write a macro which will generate functions from a list of algorithms. Since usually you have a fixed list of supported algorithms, i don't think it's worth to bother with trait objects. Also this is the main reason why I added the `phc_verify_password_noalg` method, since it will allow to first extract algorithm ID, find a target algorithm and do computations without re-checking the algorithm ID again.

password-hash/src/lib.rs
```diff
@@ -0,0 +1,269 @@
...
+        phf.hash_password(None, password.as_ref(), salt, params)
+    }
+
+    /// Verify this password hash using the specified set of supported
+    /// [`PasswordHashingFunction`] objects.
+    pub fn verify_password(
+        &self,
+        phfs: &[&dyn PasswordHashingFunction],
+        password: impl AsRef<[u8]>,
+    ) -> Result<(), VerifyError> {
```
tarcieri, line 155: @newpavlov regarding this comment:

> How to make it easier support of several algorithms at once? 

That's implemented by this method. Prospectively it could work like:

```rust
password_hash.verify_password(&[Argon2, Scrypt], password)
```

tarcieri:
Thanks for taking a look. A few general responses, and the rest I'll try to address inline

## Owned vs borrowed data

It's definitely something to consider, especially for `Ident`. I recently wrote the `pkcs8` crate and extracted the `der` crate based almost entirely on borrowed data.

It would certainly be a lot more memory efficient to write this one the same way.

## Algorithm registry

Regarding an algorithm registry, it's certainly debatable. We haven't done anything like that before in any of the trait crates so I know it seems a little weird, but I think there are some good reasons for having it.

For pretty much all other algorithms there are well-defined algorithm taxonomies based on e.g. OIDs, which can be implemented without having any sort of central algorithm registry.

If you do want to ditch it entirely, I'd suggest at least factoring the information I've collected about the various variants of Argon2, bcrypt, PBKDF2, and scrypt into the respective crates. Password hashes are a mess and I think it's important to capture that information somewhere.

All of that said, while password hashing algorithm taxonomies are a mess, they're at least a low cardinality one. Unlike all of the other algorithm families, I don't think it's that much work to actually capture the entire useful registry of password hashing algorithms, and in particular the ones which are applicable to the PHC string format. I'd say this PR is probably most of the way there already.

Given the above, that it's already implemented, and that it's gated under a cargo feature, is there a compelling reason *not* to have a password hashing algorithm registry?

password-hash/src/ident.rs
```diff
@@ -0,0 +1,253 @@
...
+    /// panicking it will cause a compile error.
+    ///
+    /// For fallible non-panicking parsing of an [`Ident`], use the [`FromStr`]
+    /// impl on this type instead, e.g. `s.parse::<Ident>()`.
+    pub const fn new(s: &str) -> Self {
+        let input = s.as_bytes();
+
+        /// Constant panicking assertion.
+        // TODO(tarcieri): use const panic when stable.
+        // See: https://github.com/rust-lang/rust/issues/51999
```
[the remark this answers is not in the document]
tarcieri: You can't unwrap in a `const fn`, period, until const panic is stable. It won't compile.

That said, this implementation includes a fallible alternative in the form of this `FromStr` implementation, which the tests heavily leverage via `"rawident".parse()`:

https://github.com/RustCrypto/traits/pull/437/files/1e4189ed0c4e19d35e51aa4f312d3041827c1736#diff-2e34ac1d42c5a9cae1b61ea25cf7de3ae4599712eb0d28c42953751310b44439R134-R166

newpavlov:
I simply do not see any use for the bunch of enums which you have added. Different variants are usually represented by different types, which would implement the `PasswordHasherAlgId`, thus effectively creating a decentralized registry. I think users will not have any use for information that two algorithms belong to the same family. Algorithm names used in PHC (`ident` in your code) are effectively OID, but more anarchistic. In practice I think it's safe to assume that we will not get algorithm name collisions.

It could be useful to have an informal registry in the form of a document, which would contain list of different password hashing algorithms, their names, and list of their parameters with recommended values, but I don't see why we should have it in the trait crate code.

It also matter of the core API. I think it's worth to keep core as simple as possible and have the PHC stuff as an optional extension. In your code the core trait pulls the `Algorithm` type, which looks like a really redundant thing to do.

tarcieri:
The other notable differentiator for password hashes, beyond the low cardinality, is the abundance of ways they can be integrated which wouldn't necessarily involve the https://github.com/rustcrypto/password-hashes implementations.

Namely, password hashes are a common feature of OS runtimes/security frameworks, and there are a number of libraries for password hashing it might be interesting to build FFI bindings to. In all of these sorts of cases a binding will have to deal with password hashes for multiple algorithms.

For people integrating with those APIs who want to provide a common interface, I think it'd be handy to have an algorithm registry in a central place, as opposed to having to pull in an additional crate for every algorithm they want to support just to define an algorithm specific enum/constants.

> It also matter of the core API. I think it's worth to keep core as simple as possible and have the PHC stuff as an optional extension. In your code the core trait pulls the Algorithm type, which looks like a really redundant thing to do.

Fair point. That could be decoupled easily enough by making the `algorithm` field of `PasswordHash` an `Ident` rather than an `Algorithm`, and having `impl TryFrom<Ident> for Algorithm`.

With that sort of structure, the entire `algorithm` module including the `Algorithm` type could be feature-gated and kept out of the core API.

newpavlov:
>I think it'd be handy to have an algorithm registry in a central place, as opposed to having to pull in an additional crate for every algorithm they want to support just to define an algorithm specific enum/constants.

I don't think such FFI crate has to pull other crates `elliptic-curves`-style (I don't see much use for the arithmetic-less elliptic curve crates as well and think that interoperability should be handled via OID, not via concrete types, but it's a different topic). It would simply define separate types/functions per each supported algorithm and implement the appropriate traits for it, i.e. I don't think it's worth to pull a crate just for one small constant and it's alright to simply copy-paste it.

I guess we could define a higher level PHC trait which will cover a set of supported algorithms, but it probably will be better to do it when someone will attempt to write such FFI crate and we be easier for us to evolve the crate.

I don't have much time to work on `password-hashes` right now, so I do not oppose merging the current version (after simplifying the core API). But I probably will attempt to rework it later using the approach outlined in the gist.

>That could be decoupled easily enough by making the algorithm field of PasswordHash an Ident rather than an Algorithm, and having impl TryFrom<Ident> for Algorithm.

For me the core trait should be implemented for a concrete algorithm (i.e. for a single `ident`), since it's more consistent with other trait crates, e.g. the `digest` traits are implemented separately for SHA-256 and SHA-512, not via a single implementation which allows to switch between them, same goes for the AES variants.

tarcieri:
> For me the core trait should be implemented for a concrete algorithm (i.e. for a single ident), since it's more consistent with other trait crates, e.g. the digest traits are implemented separately for SHA-256 and SHA-512, not via a single implementation which allows to switch between them, same goes for the AES variants.

Let me pull up the concrete traits for comparison:

## This PR

```rust
pub trait PasswordHashingFunction {
    fn hash_password(
        &self,
        alg: Option<Algorithm>,
        password: &[u8],
        salt: Salt,
        params: Params,
    ) -> Result<PasswordHash, PhfError>;
}
```

## Your `pwd_hash.rs`

https://gist.github.com/newpavlov/2b206a52a1af6f3340cfcf2293f7ac3a

```rust
pub trait PasswordHasher {
    type Params: Default;
    const DEFAULT_HASH_SIZE: usize;
    const DEFAULT_SALT_SIZE: usize;

    fn hash_password(
        password: &[u8],
        salt: &[u8],
        params: &Self::Params,
        out: &mut [u8],
    ) -> Result<(), Error>;
}
``

## Notes

- They're pretty similar.I like the `PasswordHasher` name.
- This PR uses a universal `password_hash::Params` type whereas yours uses an associated type.
- This PR supports supplying an algorithm hint
- This PR doesn't allow customizing the default hash/salt size, but rather was following the PHC string spec's recommendations of 16-byte salts and 32-byte outputs

There's a common theme to most of these differences: this PR's `PasswordHashingFunction` is designed to work as a trait object, which is what makes this API possible:

```rust
password_hash.verify_password(&[Argon2, Scrypt], password)
```

It sounds like you were intending to use macros to solve the same problem, but I think the same problem can be elegantly solved without resorting to macros.

newpavlov:
As I've said earlier, I think it's important to give users a full access to underlying algorithms without tying to PHC restrictions. Such core API will serve as an escape hatch for potential edge cases. Do not forget that it's possible to build an object-safe trait on top of a non-object-safe core API (a la `DynDigest`). Note that in the `PhcPasswordHasher` trait most methods are object-safe, the only non-safe method is `phc_hash_password_explicit`.

The universal `Params` type is more wasteful than associated type (especially if application uses only one PHF algorithm) and less flexible (though the latter is probably not that important).

Regarding trait objects vs a macro-based solution. Trait objects will result in minor inefficiencies, e.g. you will get double indirection while the macro-based one will contain none (well, there will be branches, but they are better than pointer jumps), also IIUC it results in a more efficient parsing of PHC string.

tarcieri:
> As I've said earlier, I think it's important to give users a full access to underlying algorithms without tying to PHC restrictions.

I agree, but I think that's best done through algorithm-specific inherent methods. A `PasswordHash` type is more useful in the context of cross-algorithm abstractions.

> Do not forget that it's possible to build an object-safe trait on top of a non-object-safe core API 

Also don't forget you can go the other way 😉 If we can develop a nice object safe API, there's still always the possibility of exploring a non-object-safe API which provides finer grained control.

> Regarding trait objects vs a macro-based solution. Trait objects will result in minor inefficiencies, e.g. you will get double indirection while the macro-based one will contain none (well, there will be branches, but they are better than pointer jumps), also IIUC it results in a more efficient parsing of PHC string.

Compared to the massive amounts computation performed by practically any password hashing algorithm, vtable dispatch is vanishingly small and certainly not something I'm concerned with.

I think a trait object-based approach is more maintainable and elegant than a macro, and can provide smaller code size.

tarcieri:
I went ahead and removed the registry since it seems like the most controversial thing in this PR, and also renamed `PasswordHashingFunction` to `PasswordHasher`.

Looking at converting the `Ident` (and also `Value`) struct(s) to contain a borrowed string with an associated lifetime, this does have one big drawback: it means it's no longer possible to `impl FromStr` on these types, since `FromStr` doesn't allow the input string to have a lifetime.

I can replace these with `TryFrom` impls, but this still means `str::parse` won't work, which as it were the test suite for the current implementation uses in spades... somewhat annoying.

tarcieri:
@newpavlov mind taking another look and telling me if (most of) your concerns were addressed?

tarcieri:
Regarding making `PasswordHash` generic around an algorithm-specific parameters type, one thing that strikes me as possible is renaming the current `struct Params` to something like `GenericParams`, and changing `PasswordHash` to be something like:

```rust
pub struct PasswordHash<'a, P: Params = GenericParams> {
    ...
}
```

This would allow `PasswordHash` to still operate independently of any particular password hashing algorithm, but still allow algorithms to define their own `Params` type which can be instantiated from e.g. an `Iterator` over `Pair`.

password-hash/src/ident.rs
tarcieri: Done in 739402d

password-hash/src/salt.rs
```diff
@@ -0,0 +1,224 @@
...
+/// Based on this guidance, this type enforces an upper bound of 48-bytes
+/// as a reasonable maximum, and recommends using 16-bytes.
+///
+/// [1]: https://en.wikipedia.org/wiki/Rainbow_table
+/// [2]: https://github.com/P-H-C/phc-string-format/blob/master/phc-sf-spec.md#function-duties
+/// [3]: https://github.com/P-H-C/phc-string-format/blob/master/phc-sf-spec.md#argon2-encoding
+#[derive(Copy, Clone, Eq, PartialEq)]
+pub struct Salt {
+    /// Byte array containing a salt value.
+    bytes: [u8; MAX_LENGTH],
```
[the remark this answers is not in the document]
tarcieri: I changed `Ident` to use a lifetime-backed string in 739402d. Same for the `Value` type for `Params` in 8bb6813.

I'm a bit more hesitant about `Salt` and `Output`. I haven't done it yet, but I'd like to add something to the effect of `Salt::random` which isn't really possible without an owned buffer.

tarcieri:
Regarding making `Salt` (and potentially `Output`) borrowed rather than owned values, unlike `Ident`/`Params` they require an intermediate decoding step from "B64" to be useful to a `PasswordHasher`.

tarcieri:
Seems like any lingering concerns can be fixed in follow-up PRs, so I'm going to go ahead and merge this.

PR #439, opened 2020-12-30
cipher: revert SeekNum::from_block_byte change

tarcieri:
This reverts the implementation of `SeekNum::from_block_byte` which was merged as part of #435.

I'm not exactly sure what the issue is and it's somewhat difficult to debug in that it's code that involves both generics and macros causing an error as part of any failure in a long chain of checked arithmetic, which I'm trying to debug from the context of a concrete stream cipher impl (both `chacha20` and `salsa20`) where it's operating over a generic type.

The error manifests as `OverflowError`:

https://github.com/RustCrypto/stream-ciphers/pull/205#issuecomment-752314229

This commit reverts to the previous implementation, which is at least much simpler.

### DIFF
--- a/cipher/src/stream.rs
+++ b/cipher/src/stream.rs
@@ -114,17 +114,8 @@ macro_rules! impl_seek_num {
             impl SeekNum for $t {
                 fn from_block_byte<T: TryInto<Self>>(block: T, byte: u8, bs: u8) -> Result<Self, OverflowError> {
                     debug_assert!(byte < bs);
-                    let pos = block
-                        .try_into()
-                        .ok()
-                        .and_then(|v| match byte == 0 {
-                            true => Some(v),
-                            // if `byte` is not zero, then block counter was incremented
-                            false => v.checked_sub(1),
-                        })
-                        .and_then(|v| v.checked_mul(bs as Self))
-                        .and_then(|v| v.checked_add(byte as Self))
-                        .ok_or(OverflowError)?;
+                    let block = block.try_into().map_err(|_| OverflowError)?;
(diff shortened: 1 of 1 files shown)


tarcieri:
I've confirmed this fixes the test failures in the `chacha20` and `salsa20` crates locally.

tarcieri:
@newpavlov this appears to be a bug in the new implementation of this logic, although I'm not sure what specifically the problem was

newpavlov:
I will try to look into it.

newpavlov:
Ah, I've changed how `block-buffer` handles counters to simplify downstream implementations. Let's say we have two counters `(block_counter, buffer_counter)`, assuming we process one byte at a time previously we had sequence `(0, 0)`, `(0, 1)`, `(0, 2)`, etc. In the new version we have the following sequence instead: `(0, 0)`, `(1, 1)`, `(1, 2)`, etc. It allows to merge block counter increment and generation of a new block.

tarcieri:
@newpavlov is there something that needs to change in `chacha20` and `salsa20`? I was confused why their test suites were failing with `OverflowError`:

https://github.com/RustCrypto/stream-ciphers/pull/205#issuecomment-752314229

newpavlov:
Probably, yes. I will need some time to refresh my memory of how they structured. When working on the `cipher` PR i was mostly working with the `ctr` crate, so I am not sure how compatible the introduced changes with `chacha20` and other stream ciphers.

tarcieri:
Both `chacha20` and `salsa20` contain what was originally copypasta from the `ctr` crate, so it likely needs to be updated.

Hopefully they can lean more heavily on `block-buffer` going forward to simplify these updates.

PR #447, opened 2021-01-04
aead: add `alloc`-based APIs to `stream` module

tarcieri:
Adds a corresponding set of allocating APIs to `StreamPrimitive`, `Decryptor`, and `Encryptor`.

### DIFF
--- a/aead/src/stream.rs
+++ b/aead/src/stream.rs
@@ -24,6 +24,9 @@ use generic_array::{
     ArrayLength, GenericArray,
 };
 
+#[cfg(feature = "alloc")]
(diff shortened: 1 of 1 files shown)


tarcieri:
This PR contains a bit more copy-paste than I'd like, which generally speaks to some composability problems with the current trait design (both in the `stream` module and the `aead` module in general).

That said, I'd like to at least spike out the functionality, and then pursue addressing the above by exploring some breaking changes to the `aead` crate in general.

tarcieri:
Merging this, but will post a follow-up issue on some potential (breaking) ways to refactor and clean up the `aead` crate.

Issue #450, opened 2021-01-05
password-hash: MCF support


tarcieri:
Now that #437 is landed, we should have fairly robust, generic support for the [PHC string format](https://github.com/P-H-C/phc-string-format/blob/master/phc-sf-spec.md).

However, the [Modular Crypt Format (MCF)](https://passlib.readthedocs.io/en/stable/modular_crypt_format.html) is much more common, albeit underspecified. While the PHC format makes more sense to focus on as a forward-thinking default, it would still be good to be able to parse and serialize MCF strings as well.

A couple options:

1. Have separate `PhcHash` and `McfHash` types.
2. Make `PasswordHash` capable of representing the PHC or MCF formats.

I think regardless of either of these approaches, we should treat them separately, which is to say that there should at least be separate methods for working with MCF hashes (if not types).

Beyond that, the main distinctions between the two seem to be laxer rules for MCF (e.g. it can have a trailing `$` on the end which is ignored), and how the "params" are parsed.

Namely, MCF hashes often accept a single, unlabeled parameter as the second segment. Here's an example using bcrypt:

```
$2a$10$N9qo8uLOickgx2ZMRZoMyeIjZAgcfl7p92ldGxad68LJZdL17lhWy
\__/\/ \____________________/\_____________________________/
 Alg Cost      Salt                        Hash
```

Note the `$10` "cost" parameter.

That can be parsed easily enough as a `Value`, which means we could change `PasswordHash` to have a `params` field something like:

```rust
pub enum Params {
    Phc(PhcParams),
    Mcf(Value)
}
```

This would disambiguate between PHC and MCF hashes and allow supporting both formats without too much work.

tarcieri:
To make things even trickier, it seems that the reference implementation of Argon2 now produces hashes like the following:

https://github.com/P-H-C/phc-string-format/issues/1

```
$argon2i$v=19$m=512,t=2,p=2$5VtWOO3cGWYQHEMaYGbsfQ$AcmqasQgW/wI6wAHAMk4aQ
```

Note the `$v=19` parameters separate from the `$m=512,t=2,p=2`

tarcieri:
Another thing to consider are salts. The implementation in #437 always attempted to encode/decode them as B64, however re-reading the [PHC spec](https://github.com/P-H-C/phc-string-format/blob/master/phc-sf-spec.md), it says:

[...]

PR #451, opened 2021-01-05
password-hash: remove B64 restrictions on Salt

tarcieri:
Per [this comment](https://github.com/RustCrypto/traits/issues/450#issuecomment-754750427), changes the `Salt` type to be a simple newtype of `ValueStr`.

Since the `Salt` value is now borrowed, this unfortunately removes the ability to generate a random salt into a buffer, so this PR also removes the `rand_core` dependency.

We can potentially revisit adding a type like `SaltBuf` to use for an owned salt value.

### DIFF
--- a/password-hash/src/errors.rs
+++ b/password-hash/src/errors.rs
@@ -250,8 +250,8 @@ impl std::error::Error for OutputError {}
 /// Salt-related errors.
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
 pub enum SaltError {
-    /// "B64" encoding error.
-    B64(B64Error),
+    /// Parse errors.
+    Parse(ParseError),
 
     /// Salt too short (min 4-bytes).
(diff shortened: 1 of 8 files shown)


password-hash/src/errors.rs
```diff
@@ -250,8 +250,8 @@ impl std::error::Error for OutputError {}
 /// Salt-related errors.
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
 pub enum SaltError {
-    /// "B64" encoding error.
-    B64(B64Error),
+    /// Parse errors.
+    Parse(ParseError),
 
     /// Salt too short (min 4-bytes).
     TooShort,
```
tarcieri, line 257: We could probably get rid of `SaltError` and combine it with `ParseError`

tarcieri:
Landing this. Will follow up on error type simplification.

PR #476, opened 2021-01-15
password-hash: add associated Params type to PasswordHasher

tarcieri:
Adds an associated type for algorithm parameters, then handles automatically converting to them when using higher-level APIs.

### DIFF
--- a/password-hash/src/lib.rs
+++ b/password-hash/src/lib.rs
@@ -41,22 +41,6 @@ extern crate alloc;
 #[cfg(feature = "std")]
 extern crate std;
 
-use core::{
-    convert::{TryFrom, TryInto},
-    fmt,
-};
-
-pub use crate::{
-    errors::{HashError, HasherError, VerifyError},
-    ident::Ident,
-    output::Output,
-    params::Params,
-    salt::Salt,
-    value::{Decimal, Value, ValueStr},
-};
-
-pub use b64ct as b64;
-
 pub mod errors;
 
 mod ident;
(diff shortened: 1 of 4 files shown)


password-hash/src/params.rs
```diff
@@ -39,14 +39,14 @@ const MAX_LENGTH: usize = 8;
 ///
 /// [1]: https://github.com/P-H-C/phc-string-format/blob/master/phc-sf-spec.md#specification
 #[derive(Clone, Default, Eq, PartialEq)]
-pub struct Params<'a> {
+pub struct ParamsBuf<'a> {
```
tarcieri, line 42: Note: this name change is motivated both by having a different name versus `T::Params` and some additional refactoring I'd like to do.

password-hash/src/lib.rs
```diff
@@ -41,35 +41,41 @@ extern crate alloc;
...
 /// Separator character used in password hashes (e.g. `$6$...`).
 const PASSWORD_HASH_SEPARATOR: char = '$';
 
 /// Trait for password hashing functions.
 pub trait PasswordHasher {
+    /// Algorithm-specific parameters.
+    type Params: Clone
+        + Debug
+        + for<'a> TryFrom<&'a ParamsBuf<'a>, Error = HasherError>
+        + for<'a> TryInto<ParamsBuf<'a>, Error = HasherError>;
```
tarcieri, line 77: I played with a few different combinations of these bounds and this was the best I could come up with.

I think the one other thing to potentially consider is removing the `Error` bound on the `TryInto` impl, which would allow for an `Into` impl to satisfy this bound.

PR #502, opened 2021-01-29
digest: variable output traits

newpavlov:
The mid-level `VariableOutput` trait is the same as in `digest v0.9`, but on lower-level it's based on the `VariableOutputCore`. Types which implement this core trait can be used in two ways:
- By specifying output size at compile time via `CtVariableCoreWrapper`. This wrapper implements the fixed core traits, so for example `Groestl384` will be defined as `UpdateCoreWrapper<CtVariableCoreWrapperGroestlBigCore, U48>>`.
- By selecting output size at run time via `RtVariableCoreWrapper`. This wrapper implements the `VariableOutput` trait.

This approach could be seen as a bit of generalization overkill, but I think it neatly describes structure of variable hash functions and allows all the flexibility we could want.

Also I have restructured the code a bit. Now the core traits and wrappers reside in the `core_api` module, thus top-level docs are no longer cluttered with them.

### DIFF
--- a/digest/src/core_api/ct_variable.rs
+++ b/digest/src/core_api/ct_variable.rs
@@ -1,0 +1,74 @@
+use super::{AlgorithmName, FixedOutputCore, UpdateCore, VariableOutputCore};
+use block_buffer::BlockBuffer;
+use core::{fmt, marker::PhantomData};
+use generic_array::typenum::type_operators::IsLessOrEqual;
+use generic_array::{ArrayLength, GenericArray};
+
+/// Wrapper around [`VariableOutputCore`] which selects output size
+/// at compile time.
+#[derive(Clone)]
+pub struct CtVariableWrapper<T, OutSize>
+where
+    T: VariableOutputCore,
+    OutSize: ArrayLength<u8> + IsLessOrEqual<T::MaxOutputSize>,
+{
+    inner: T,
+    _out: PhantomData<OutSize>,
+}
+
+impl<T, OutSize> UpdateCore for CtVariableWrapper<T, OutSize>
+where
+    T: VariableOutputCore,
+    OutSize: ArrayLength<u8> + IsLessOrEqual<T::MaxOutputSize>,
+{
+    type BlockSize = T::BlockSize;
+
+    #[inline]
+    fn update_blocks(&mut self, blocks: &[GenericArray<u8, Self::BlockSize>]) {
+        self.inner.update_blocks(blocks);
+    }
+}
+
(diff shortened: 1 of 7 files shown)


digest/src/core_api/ct_variable.rs
```diff
@@ -0,0 +1,74 @@
...
+use generic_array::typenum::type_operators::IsLessOrEqual;
+use generic_array::{ArrayLength, GenericArray};
+
+/// Wrapper around [`VariableOutputCore`] which selects output size
+/// at compile time.
+#[derive(Clone)]
+pub struct CtVariableWrapper<T, OutSize>
+where
+    T: VariableOutputCore,
+    OutSize: ArrayLength<u8> + IsLessOrEqual<T::MaxOutputSize>,
```
newpavlov, line 13: Unfortunately `IsLessOrEqual` will be not an easy bound to specify using current design of const generics.

digest/src/lib.rs
```diff
@@ -173,3 +171,67 @@ pub trait ExtendableOutput {
         buf
     }
 }
+
+/// Trait for variable output size hash functions.
+pub trait VariableOutput: Sized {
```
tarcieri, line 176: I think it'd be nice if this trait had a method like `Digest::digest`, something like:

```rust
fn digest_variable(input: &[u8], output: &mut [u8]) -> Result<(), InvalidOutputSize> {
    let hasher = Self::new(output.len())?;
    hasher.update(input);
    hasher.finalize_variable(|out| output.copy_from-slice(out));
    Ok(())
}
```

...which could be used as a one-shot API

newpavlov: Done!

tarcieri: Sweet, thank you!

tarcieri (approved):

digest/src/core_api/ct_variable.rs
newpavlov: Welp, looks `IsLessOrEqual` bound does nothing, at least when used with type aliases...

newpavlov: Hm, looks like I was using it incorrectly. The bound should look like this:
```rust
OutSize: ArrayLength<u8> + IsLessOrEqual<T::MaxOutputSize>,
LeEq<OutSIze, T::MaxOutputSize>: NonZero
```

newpavlov: The bound is now fixed, but type aliases still do not check bounds, so compilation error will not happen until an incorrect alias will be used in code. Not ideal, but we can not do anything about it.

Issue #489, opened 2021-01-25
Add Nonce, Tag sizes to Aead Traits


theory:
I've been experimenting with the AEAD implementations, such as chacha20poly1305, by creating a vector of the appropriate size and putting the nonce, ciphertext, and tag into it. However, it's tricky to do this if one does not know the nonce or tag size until runtime, for example when giving the user the choice of AEADs. Would the project be receptive to adding a trait such as this to the implemented AEADs?

``` rust
trait SealSize {
    /// The length of a nonce.
    type NonceSize: ArrayLength<u8>;

    /// The length of the tag.
    type TagSize: ArrayLength<u8>;

    fn nonce_size() -> usize {
        Self::NonceSize::to_usize()
    }
    fn tag_size() -> usize {
        Self::TagSize::to_usize()
    }
}
```

That would enable one to create the appropriate-sized buffers in, say a blanket implementation such as:

``` rust
impl<D: Aead + SealSize> Seal for D {
    fn seal(&self, msg: &[u8], aad: &[u8]) -> Result<Vec<u8>> {
        // Generate a nonce.
        let mut nonce = vec![0u8; Self::nonce_size()];
        rand::thread_rng().fill(&mut nonce[..]);
        // ...
    }
}
```

newpavlov:
We have a similar [method](https://docs.rs/digest/0.9.0/digest/trait.Digest.html#tymethod.output_size) in the `Digest` trait, so personally I am not against addition of such methods to the existing traits (though I prefer to use `Self::NonceSize::USIZE` instead). But note that such methods probably will removed after migration to const generics.

It's also may be worth to consider addition of an object-safe trait similar to [`DynDigest`](https://docs.rs/digest/0.9.0/digest/trait.DynDigest.html).

theory:
Is there a design document for const generics? Curious what that looks like.

newpavlov:
Are you asking about [RFC 2000](https://github.com/rust-lang/rfcs/blob/master/text/2000-const-generics.md)? Its MVP version has been already stabilized and will be available in Rust 1.51, but unfortunately it's not sufficient for our needs (most notable due to this [issue](https://github.com/rust-lang/rust/issues/60551)), so migration will be farther in future.

tarcieri:
I was planning a refactor of the `aead` traits which accomplishes some of these goals for the next release.

[...]

PR #508, opened 2021-01-30
aead: extract AeadCore trait

tarcieri:
Closes https://github.com/RustCrypto/traits/issues/489 (potentially, cc @theory).

Extracts a single trait for defining nonce, tag, and overhead sizes instead of repetitively defining those associated types on every other trait.

This allows for fully generic `Nonce<A>` and `Tag<A>` type aliases.


tarcieri:
This doesn't include helper methods which return `usize`, but those could potentially be included.

theory:
Looks like it removes a ton of repetitiveness, nice!

PR #518, opened 2021-02-01
password-hash: add encoding support to PasswordHash

tarcieri:
More work towards being able to use `PasswordHash` with SHA-crypt, which uses a PHC-like hash/params, but with `crypt(3)` encoding.

This change is also trying to be purely additive to avoid having to cut new minor releases of everything.

### DIFF
--- a/password-hash/src/lib.rs
+++ b/password-hash/src/lib.rs
@@ -249,6 +249,11 @@ pub struct PasswordHash<'a> {
 impl<'a> PasswordHash<'a> {
     /// Parse a password hash from a string in the PHC string format.
     pub fn new(s: &'a str) -> Result<Self, HashError> {
+        Self::parse(s, Encoding::B64)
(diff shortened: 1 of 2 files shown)


password-hash/src/lib.rs
```diff
@@ -249,6 +249,11 @@ pub struct PasswordHash<'a> {
 impl<'a> PasswordHash<'a> {
     /// Parse a password hash from a string in the PHC string format.
     pub fn new(s: &'a str) -> Result<Self, HashError> {
+        Self::parse(s, Encoding::B64)
+    }
+
+    /// Parse a password hash from the given [`Encoding`].
+    pub fn parse(s: &'a str, encoding: Encoding) -> Result<Self, HashError> {
```
tarcieri, line 256: Perhaps this should be `decode` for parity with some of the other methods?

PR #524, opened 2021-02-04
.img: add diagram of Rogaway's STREAM construction

tarcieri:
For use in `aead` crate rustdoc


newpavlov:
Maybe we should create repository for media files?

tarcieri:
Now I'm reminded I was using https://github.com/RustCrypto/meta/tree/master/img for this before, but if that isn't the sanctioned place to put images, perhaps there should be one

newpavlov:
I also put the logos into the meta repository, but in hindsight we probably should have created a separate repo for media files.

newpavlov:
I have created https://github.com/RustCrypto/media, but we probably should keep the existing files in the meta repository for some time as well so existing links will not be broken.

tarcieri:
In the PR that links this image, I put the commit hash in the image URL to prevent breakages when images get moved around. I think at least some of the other images might be linked that way.

Issue #444, opened 2020-12-30
Improving cipher parallelism


tarcieri:
Taking a step back from https://github.com/RustCrypto/traits/pull/354, I thought it'd be good to look how and where ILP and SIMD parallelism is currently used across the project as a whole, and how that could be improved.

The only place we presently have any sort of parallelism abstraction at the trait-level is `BlockCipher::ParBlocks`.  Otherwise various crates leverage e.g. SIMD internally. Regarding `BlockCipher::ParBlocks` specifically, the only crate that leverages it is the `aes` crate.

The following crates have SIMD backends:

## Ciphers

- `aes`
- `chacha20`

## UHFs/"MACs"

- `polyval`
- `poly1305`

## AEADs

In AEADs, we'd like to glue the above crates together in fairly fixed combinations in order to leverage ILP, passing SIMD buffers from ciphers to UHFs for authentication:

- `aes-gcm`/`aes-gcm-siv`: `aes` + `ghash`/`polyval`
- `chacha20poly1305`: `chacha20` + `poly1305`

(also `aes-siv` and `pmac`, but this is less of a priority)

In either of these cases there's a single specific buffer type I think it'd be nice for both the cipher implementation and UHF to support in common:

- `aes-gcm`/`aes-gcm-siv`: "i128x8" i.e. `[__m128i; 8]` on x86/x86_64
- `chacha20poly1305`: "i256x4" i.e. `[__m256i; 4]` on x86/x86_64

## Concrete proposal

My suggestion is to get rid of `BlockCipher::ParBlocks` and replace it with more general SIMD types and traits designed to work with them, namely:

- Add a new `utils` crate e.g. `simd-buffers` which provides "i128x8" and  "i256x4" SIMD buffer types which are backed by `__m128i`/`__m256i` on x86/x86_64 and otherwise provide a portable implementation. These types don't need to implement any sort of arithmetic, just provide wrappers for passing data between SIMD implementations.
- Add traits to `cipher` and `universal-hash` which operate on SIMD buffers.
- Use SIMD buffers types in the implementations of `aes-gcm`, `aes-gcm-siv`, and `chacha20poly1305`

## `cipher` API suggestion

I'd suggest adding traits to `cipher` which use the SIMD buffer types which are useful for both block ciphers and stream ciphers.

I also think it might make sense to use a generic parameter rather than an associated type to permit support for multiple buffer types (e.g. on newer CPUs, "i128x4" might be a better option for AES, but we can support both):

```rust
[...]

PR #528, opened 2021-02-05
[WIP] cipher/universal-hash: generic buffer traits

### LINKED ISSUE #159: block-cipher: Block alignment
Some hardware implementations of AES need particular alignment of blocks to operate directly over AHB (e.g. NXP LPC55's HASHCRYPT wants 32-bit alignment, I can imagine this is a typical case for ARM microcontrollers). If a large chunk of data needs to be en/de-crypted that's aligned, it would be neat to pass this information through on the type level to avoid stack copies for each block. The `aligned` crate would be one approach to do it.

Splitting this out from https://github.com/RustCrypto/traits/issues/43 for possible discussion.

tarcieri:
Adds traits which are generic around a buffer type, allowing more precise control of inputs to encryption and universal hash algorithms.

Potentially resolves #159 and #444.

### DIFF
--- a/cipher/src/common.rs
+++ b/cipher/src/common.rs
@@ -1,6 +1,9 @@
 //! Functionality common to block ciphers and stream ciphers
 
-use crate::{errors::InvalidLength, BlockCipher, NewBlockCipher};
+use crate::{
+    errors::{BufferError, InvalidLength},
+    BlockCipher, NewBlockCipher,
+};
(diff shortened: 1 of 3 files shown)


tarcieri:
This is a more minimalist take on what I was trying to do in https://github.com/RustCrypto/utils/pull/221

The trait names are perhaps a little bit confusing, but I'd like to explore the general API shape as a possible way of preventing needless stores/loads when shuffling data between block ciphers, stream ciphers, and universal hash functions.

PR #531, opened 2021-02-10
password-hash: add docs.rs metadata to Cargo.toml

abonander:
Based on the following `#[cfg_attr]` it looks like this was the intention but the required lines in Cargo.toml were omitted: https://github.com/RustCrypto/traits/blob/073edacf50e04c257053b6c2ef2fea8276a241e1/password-hash/src/salt.rs#L185


tarcieri:
Awesome, thanks!

PR #537, opened 2021-02-11
Add crypto-common crate

### LINKED ISSUE #259: aead: add FromBlockCipher trait?
Unfortunately we can't rely on `From<BlockCipher>`, see #217.

### LINKED ISSUE #503: digest: compatibility with lazy buffers
Currently `block-buffer` and `digest` do not support lazy buffering, which is needed for some algorithms like Skein. Instead of padding those algorithms "tweak" compression function with a flag denoting last block. Lack of support is intentional, since it allows to support `BlockBuffer` invariant (cursor is always smaller than block size) used by padding functions.

One possible way forward is to generalize `digest::core_api::UpdateCoreWrapper` over block buffer types, i.e. `block-buffer` will contain two buffer types: `BlockBuffer` and `LazyBlockBuffer`, generalized by a trait.

### LINKED ISSUE #536: Crate for common traits?
For a long time I thought about creating a crate with some common traits shared by several algorithm families. Here is a rough list of traits which could be defined in it:
- `FromKey`: replaces `NewAead`, `NewMac`, `NewUniversalHash`, `NewBlockCipher`.
- `FromKeyNonce`: replaces `NewCipher`. Could be used for AEAD?
- `Update`/`UpdateCore`: replaces `digest::Update/UpdateCore`, partially `Mac` and `UniversalHash`.
- `Reset`: replaces `digest::Reset`, partially `Mac` and `UniversalHash`. Also can be a potential shortcut for `StreamCipherSeek::seek(0)`.
- `FinalizeFixed`: replaces `digest::FixedOutput`, partially `Mac` and `UniversalHash`. `Mac` and `UniversalHash` will be extension traits for `FinalizeFixed` with additional family-specific helper methods.

Higher level crates will simply re-import required traits, so existence of this crate should be almost invisible for users.

Addition of such crate will help reduce a bit the total amount of traits which we define. The main motivation for introducing it now is trait interactions in `blake2`. The main disadvantage of this approach is that we add yet another crate to our already somewhat complex crate hierarchy. Also even though the proposed traits are quite simple and probably will be rarely changed, breaking change in this crate will ripple through downstream trait crates.

newpavlov:
Closes #536 
Closes #259
Closes #503


crypto-common/src/lib.rs
```diff
@@ -0,0 +1,169 @@
...
+    ) -> (
+        GenericArray<u8, Self::KeySize>,
+        GenericArray<u8, Self::NonceSize>,
+    ) {
+        (Self::generate_key(&mut rng), Self::generate_nonce(&mut rng))
+    }
+}
+
+/// Trait for types which can be created from key.
+pub trait FromKey: Sized {
```
tarcieri, line 86: Nice

tarcieri (approved):

newpavlov:
Argh, because `FromKeyNonce` is now defined in another crate we can not write the following impl in `cipher`:
```rust
impl<C> FromKeyNonce for C
where
    C: FromBlockCipherNonce,
    C::BlockCipher: FromKey,
{ .. }
```
I hope such implementations (when generic type is bound by a trait defined in the current crate) will be allowed one day. We could use it for generic implementation of `io::Write` in terms of `Update` as well. Until then the easiest solution probably will be to implement `FromKeyNonce` in implementation crates directly...

tarcieri:
@newpavlov failing everything else there's always macros

newpavlov:
Well, on a plus side with the macro approach we do not need the `FromBlockCipher` trait anymore...

It looks like this PR is more or less ready, but before merging I would like to try migrate the downstream crates first.

PR #565, opened 2021-02-26
Implement {to,char}_le_bits for MockCurve

rvolgers:
This seems like it's the most direct path to getting some other stuff upstream, and I figure it's probably pretty unobjectionable.


tarcieri:
Thanks!

Issue #31, opened 2018-11-18
Consider supporting buffer-to-buffer cipher operations


newpavlov:
For some use-cases (e.g. in TLS implementations) you need to decrypt/encrypt data from read-only source and write result into provided buffer.

To prevent code duplication I think the best solution will be to implement algorithms over enum like this:
```rust
enum CryptoBuf<'a> {
    InBuf(&'a mut [u8]),
    BufToBuf { in_buf: &'a [u8], out_buf: &'a mut [u8] },
}

trait Cipher: Sized {
    /// Users are heavily discouraged from using this method
    fn _encrypt(self, data: CryptoBuf);
    fn encrypt(self, buf: &mut [u8]) {
        self._encrypt(CryptoBuf::InBuf(buf));
    }
    fn encrypt_b2b(self, in_buf: &mut [u8], out_buf: &mut [u8]) -> Result<(), Error> {
        if check_lengths(in_buf, out_buf) { Err(Error)? }
        self._encrypt(CryptoBuf::BufToBuf { in_buf, out_buf });
        Ok(())
    }
}
```

Any thoughts?

tarcieri:
Isn't the general idea more simply expressed as:

```rust
in_buf: Option<&'a [u8]>
```

...with the idea being that if `in_buf` is `None`, then `out_buf` is also the input buffer?

newpavlov:
Hm, you are right, but I think using a separate enum will be more ergonomic (for algorithm implementations), e.g. we will be able to offload to it a lot of common functionality, like: reading, writing, indexing, slicing, ensuring that buffers have equal length, etc.

I think it also could be worth to think about how we could integrate `Read` and `Write` here.

PR #567, opened 2021-02-26
cipher: add block mode traits and wrappers

### LINKED ISSUE #564: New block-modes API design
(from https://github.com/RustCrypto/block-ciphers/pull/227#issuecomment-783074558)

* At the type system, support modes requiring only an block encryption function (as opposed to both encryption and decryption)
* bring back separate types for encryptors and decryptors
* splitting the trait into two parts: core which works only on blocks (and thus padding-independent) and a buffered wrapper on top of it which would work with slices and will be generic over padding.

newpavlov:
Closes #564


newpavlov:
Unfortunately, the block mode wrappers are a bit sub-optimal. They first copy data to an output buffer and only then encrypt/decrypt it, i.e. they effectively implement a two-pass approach. But to solve this issue we would have to expose block mode parallelism capabilities... Also ideally we would need a solution for #31. So I think it's worth to postpone it until later versions.

Another interesting observation is that mutable block cipher traits are quite symmetric with the block mode traits. I think it could be worth to explore merging them, e.g. we could say that hardware blocks implement not a block cipher, but ECB mode, on top of which we can implement other block cipher modes.

tarcieri:
> Another interesting observation is that mutable block cipher traits are quite symmetric with the block mode traits. I think it could be worth to explore merging them, e.g. we could say that hardware blocks implement not a block cipher, but ECB mode, on top of which we can implement other block cipher modes.

Interesting idea

aead/src/stream.rs
```diff
@@ -322,12 +323,13 @@ macro_rules! impl_stream_object {
...
-                self.stream.$in_place_op(self.position, true, associated_data, buffer)
+                self.stream
+                    .$in_place_op(self.position, true, associated_data, buffer)
             }
         }
-    }
+    };
 }
 
 impl_stream_object!(
```
tarcieri, line 335: Huh, did `rustfmt` add new formatting rules for macros or something?

tarcieri (approved):

aead/src/stream.rs
newpavlov: Hm, it could be a nightly thing. I will revert the `aead` formatting changes.

newpavlov:
The experimental changes which I currently explore have become too big, so I will merge this PR and open a new one instead.

PR #573, opened 2021-03-04
digest: allow DynDigest usage without `alloc` enabled

### LINKED ISSUE #545: DynDigest requires alloc when it does not have to
It only provides functions which return the digest in a `Box`.

Perhaps this is under the assumption that `DynDigest` will always be boxed and so alloc is required anyway, but this does not have to be the case. You can have the digest living on the stack in a statically dispatched function, which then creates a `&mut dyn DynDigest` to  pass it on to dynamically-dispatched functions. This might be done to keep code size down, because parameterizing a lot of functions by `D: Digest` will cause them all to be duplicated for every digest used in the program, which can be quite a lot if the digest to use can be specified in user input.

I propose adding `finalize_slice` and `finalize_reset_slice` to `DynDigest` and to `Digest` (TODO: ... and to one of the lower level traits, presumably?), which take a `&mut [u8]` and simply copy as many bytes of the digest into them as will fit. This makes these functions also useful for various use cases where hashes are used to expand key material into a variable sized stream of bytes.

The caller can simply use the existing `output_size` function to determine the size of the digest ahead of time, and use that information to pass an appropriate slice to `finalize_slice`.

Edit: originally I had the function names end in "`_into_slice`", but this makes it sound like it references the `Into` trait and related functions. Using just "`_slice`" sounds a bit more unique and may draw more attention to the unique part of its behavior, namely truncation if too short a slice is passed. Yak shaving welcome.

tarcieri:
Moves `alloc`-based gating for `DynDigest` to just the methods that use `Box`, which enables usage of the trait in "heapless" environments.

Closes #545 (cc @rvolgers)

### DIFF
--- a/digest/src/dyn_digest.rs
+++ b/digest/src/dyn_digest.rs
@@ -1,12 +1,12 @@
-use alloc::boxed::Box;
-use core::fmt;
-
 use super::{FixedOutput, FixedOutputReset, Reset, Update};
+use core::fmt;
 use generic_array::{typenum::Unsigned, GenericArray};
+
+#[cfg(feature = "alloc")]
+use alloc::boxed::Box;
 
 /// The `DynDigest` trait is a modification of `Digest` trait suitable
 /// for trait objects.
-#[cfg_attr(docsrs, doc(cfg(feature = "alloc")))]
 pub trait DynDigest {
     /// Digest input data.
     ///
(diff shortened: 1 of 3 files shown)


tarcieri:
Hmm, `clippy` isn't a fan of the `finalize` definition:

```
error: local variable doesn't need to be boxed here
  --> digest/src/dyn_digest.rs:28:17
   |
28 |     fn finalize(mut self: Box<Self>) -> Box<[u8]> {
   |                 ^^^^^^^^
   |
   = note: `-D clippy::boxed-local` implied by `-D warnings`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#boxed_local
```

@newpavlov why isn't the method definition just `fn finalize(mut self)`?

digest/src/dyn_digest.rs
```diff
@@ -36,6 +49,8 @@ pub trait DynDigest {
     fn output_size(&self) -> usize;
 
     /// Clone hasher state into a boxed trait object
+    #[cfg(feature = "alloc")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "alloc")))]
     fn box_clone(&self) -> Box<dyn DynDigest>;
```
tarcieri, line 54: While I was able to make the two other alloc-gated methods provided, this one is notably required, and somewhat annoyingly means that downstream impls of `DynDigest` would also need to `alloc`-gate this as well.

Given that `DynDigest` is largely intended as a wrapper for `Digest` anyway, I'm not sure it's a problem.

newpavlov (commented):
Looks good, but you are missing the `alloc`-independent finalization method, without which this trait is effectively useless with disabled `alloc`. Design of this method is the biggest problem for solving the issue.

tarcieri:
The `::finalize_into` and `::finalize_into_reset` methods are `alloc`-independent after this PR.

newpavlov:
Ah, indeed! I forgot about them. Do you think it could worth to add the truncating method discussed in the issue?

tarcieri:
Personally I'd worry about silent failures due to truncation, but that aside, perhaps we can revisit that in a separate PR if there's still interest

newpavlov (approved):

PR #590, opened 2021-04-12
Add support for PointCompaction trait

vihu:
This PR enables support for `PointCompaction` specifically as detailed in [IETF-draft-jivsov-ecc-compact-05](https://tools.ietf.org/html/draft-jivsov-ecc-compact-05#section-3).

The key idea being that the `y` coordinate of a point on the curve can be derived from the `x` coordinate allowing a more concise representation of the points. Note that this is different from `PointCompression`.

### DIFF
--- a/elliptic-curve/src/public_key.rs
+++ b/elliptic-curve/src/public_key.rs
@@ -5,7 +5,7 @@ use crate::{
     sec1::{
         EncodedPoint, FromEncodedPoint, ToEncodedPoint, UncompressedPointSize, UntaggedPointSize,
     },
-    weierstrass::{Curve, PointCompression},
+    weierstrass::{Curve, PointCompaction, PointCompression},
     AffinePoint, Error, FieldBytes, NonZeroScalar, ProjectiveArithmetic, ProjectivePoint, Result,
     Scalar,
 };
@@ -224,7 +224,7 @@ where
 
 impl<C> From<PublicKey<C>> for EncodedPoint<C>
 where
-    C: Curve + ProjectiveArithmetic + PointCompression,
+    C: Curve + ProjectiveArithmetic + PointCompression + PointCompaction,
     Scalar<C>: PrimeField<Repr = FieldBytes<C>>,
     AffinePoint<C>: Copy + Clone + Debug + Default + FromEncodedPoint<C> + ToEncodedPoint<C>,
     ProjectivePoint<C>: From<AffinePoint<C>>,
@@ -238,7 +238,7 @@ where
 
 impl<C> From<&PublicKey<C>> for EncodedPoint<C>
 where
-    C: Curve + ProjectiveArithmetic + PointCompression,
+    C: Curve + ProjectiveArithmetic + PointCompression + PointCompaction,
     Scalar<C>: PrimeField<Repr = FieldBytes<C>>,
     AffinePoint<C>: Copy + Clone + Debug + Default + FromEncodedPoint<C> + ToEncodedPoint<C>,
     ProjectivePoint<C>: From<AffinePoint<C>>,
--- a/elliptic-curve/src/sec1.rs
+++ b/elliptic-curve/src/sec1.rs
@@ -193,10 +193,17 @@ where
         self.tag().is_compressed()
     }
 
+    /// Is this [`EncodedPoint`] compact?
+    pub fn is_compact(&self) -> bool {
+        self.tag().is_compact()
+    }
+
     /// Compress this [`EncodedPoint`], returning a new [`EncodedPoint`].
     pub fn compress(&self) -> Self {
         match self.coordinates() {
-            Coordinates::Identity | Coordinates::Compressed { .. } => self.clone(),
+            Coordinates::Identity
+            | Coordinates::Compressed { .. }
+            | Coordinates::Compact { .. } => self.clone(),
             Coordinates::Uncompressed { x, y } => Self::from_affine_coordinates(x, y, true),
         }
     }
@@ -218,6 +225,7 @@ where
                     .into()
             }
             Coordinates::Uncompressed { .. } => Some(self.clone()),
+            Coordinates::Compact { .. } => Some(self.clone()),
         }
     }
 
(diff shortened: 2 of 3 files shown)


tarcieri:
The `draft-jivsov-ecc-compact` draft appears to have expired nearly 7 years ago. Are you aware of any plans to revive it? Generally we would prefer not to implement draft constructions unless there is a strong case that they will eventually be standardized in some form or another.

Also, I'm not sure I understand the purpose of this draft, especially when it comes to SEC1 encodings. The draft notes:

> The requirement to store one bit to identify the 'y' means that
> the whole byte is required.

...but this benefit seems to be lost as soon as a SEC1 tag is added, as the SEC1 tag can encode this extra bit.

I guess the benefit is only realized after stripping the SEC1 tag and operating on the untagged bytes, but even then: why introduce a new SEC1 tag?

tarcieri:
My suggestion would be to put this functionality in its own crate. It should be possible to implement generically over curves. The SEC1 extension could be implemented as a newtype.

tarcieri:
> to check whether an AffinePoint is compactable

The `k256` crate has a semi-hidden `expose-field` feature which should be sufficient for this purpose. We could potentially add a similar feature to `p256`.

Generally we've tried to keep field elements hidden from the public API to prevent misuse.

> ...to do the actual compaction/decompaction, CURVE_EQUATION_A and B need to be available

It seems like if you use the "black box" key generation method based on rejection sampling, exposing the curve equation constants is unnecessary? You can produce an uncompressed `EncodedPoint`, call `coordinates` to obtain the coordinates, interpret the `y`-coordinate as a `FieldElement`, and then check if `y == min(y,p-y)`.

Likewise, it seems like it would be possible to implement decompaction with a similar approach, decompressing the `x` coordinate with both the y-is-even and y-is-odd possibilities, and then selecting the one where `y == min(y,p-y)`, but perhaps there's something I'm missing.

If that is actually sufficient, we can add an `expose-field` feature to `p256` as well. In the meantime you can prototype with `k256`.

If this is insufficient, perhaps we can look into adding a trait to compute something like `C(x)` you can use in the core of your implementation.

madninja:
> > to check whether an AffinePoint is compactable
> 
> The `k256` crate has a semi-hidden `expose-field` feature which should be sufficient for this purpose. We could potentially add a similar feature to `p256`.
> 
> Generally we've tried to keep field elements hidden from the public API to prevent misuse.

Makes total sense... I'll take a look at the k256 feature and propose  a PR for a p256 version since that's what we're using in Helium today. 

> 
> > ...to do the actual compaction/decompaction, CURVE_EQUATION_A and B need to be available
> 
> It seems like if you use the "black box" key generation method based on rejection sampling, exposing the curve equation constants is unnecessary? You can produce an uncompressed `EncodedPoint`, call `coordinates` to obtain the coordinates, interpret the `y`-coordinate as a `FieldElement`, and then check if `y == min(y,p-y)`.

That would work, except that the current FieldElement implementation does not have a `min` implementation which is what the "informed_subtract" works around (to avoid a full/impossible(?) ordering implementation on FieldElements.  

> 
> Likewise, it seems like it would be possible to implement decompaction with a similar approach, decompressing the `x` coordinate with both the y-is-even and y-is-odd possibilities, and then selecting the one where `y == min(y,p-y)`, but perhaps there's something I'm missing.

I think that may work if there was a `min` implementation available. Lacking that I was going down the path of a PR exposing FieldElement limbs and implementing informed_subtract in the wrapping crate

> 
> If that is actually sufficient, we can add an `expose-field` feature to `p256` as well. In the meantime you can prototype with `k256`.

Worth a quick attemp, yeah

tarcieri:
> That would work, except that the current FieldElement implementation does not have a min implementation 

I think you can hack around that by serializing each `FieldElement` to bytes and comparing the bytes.

madninja:
> > That would work, except that the current FieldElement implementation does not have a min implementation
> 
> I think you can hack around that by serializing each `FieldElement` to bytes and comparing the bytes.

Hmm, really? That'd be interesting. I could not figure out how to check if there was "underflow" in a subtraction that caused a wrap to the top end of the p field, because that fact is lost in the subtraction. And I'm unclear (and highly uneducated on this) if there's any to compare serialized FieldElement bytes either before or after they're converted to Montgomery form. 

We'd still need access to MODULUS to compute p-y though.

tarcieri:
The byte serialization "lifts" field elements into big endian integers. You should be able to compare them with those, I believe.

If not, we can add something like a `Min` trait that returns a `CtOption`

madninja:
> The byte serialization "lifts" field elements into big endian integers. You should be able to compare them with those, I believe.

I'll have try that again. I'm not versed well enough what happens there to be able to compare them to get a "min" like effect at all

> 
> If not, we can add something like a `Min` trait that returns a `CtOption`

That'd be neat yeah..I believe it would likely still need to do the informed subtract as part of it's implementation to detect modulo field wraps

madninja:
I see some potentially related work in the `Order` trait. Any hints if and how best to use this?

tarcieri:
@madninja I think it'd be interesting to expose more of the curve parameters through traits, particularly in the next release.

It'd be nice if we had a proper const-friendly bignum type for this first, particularly if const generics were advanced enough for it.

tarcieri:
I'm curious if we could work towards making the functionality needed to implement this functionality a part of the public API of the `elliptic-curve` crate, but gated under the `hazmat` feature.

It wouldn't be stable until the next release of the `elliptic-curve` crate, but perhaps we could plan to have something ready before then and you could test prototypes and give feedback?

madninja:
> I'm curious if we could work towards making the functionality needed to implement this functionality a part of the public API of the `elliptic-curve` crate, but gated under the `hazmat` feature.
> 
> It wouldn't be stable until the next release of the `elliptic-curve` crate, but perhaps we could plan to have something ready before then and you could test prototypes and give feedback?

Most definitely! We'd be very interested in getting past the patch dance we're currently having to do to keep up!

Issue #681, opened 2021-07-09
Improving `OsRng` access


tarcieri:
We have several crates, some in this repo, that use traits from `rand_core`. An example is the `aead` crate:

https://docs.rs/aead/0.4.1/aead/trait.NewAead.html#method.generate_key

However, while some crates document how to use `OsRng` as the concrete RNG, we typically don't make this easy.

Some suggestions for improving this:

## Re-export `rand_core`

This would make it easier to access both the traits and the built-in RNG.

It could be exported verbatim as `rand_core`, ~~or perhaps we could consider making the feature name `rand` and re-exporting it as `rand`. But that might be a bit less clear than just the `rand_core` name, even though it's a bit uglier.~~ (edit: so far we're just doing `rand_core` for clarity that it's a re-export)

## Have the `std` feature activate `rand_core/std`

This would mean that `rand_core` is always pulled in whenever `std` is enabled (at least until weak feature activation ships), but would make it easy for users to access `rand_core::OsRng` when the `std` feature is enabled, e.g. as `aead::rand_core::OsRng`.

## Implementation status

- [x] #682
- [ ] `cipher`
- [ ] `crypto-mac`
- [ ] `elliptic-curve`
- [ ] `password-hash`
- [ ] `signature`

tarcieri:
#682 implemented this for `aead`, re-exporting `rand_core`.

I'll go ahead and add a table for the others to the toplevel description.

PR #682, opened 2021-07-10
aead: re-export `rand_core`

tarcieri:
This PR implements the changes proposed in #681 to simplify access to the `rand_core` traits and `OsRng`.

`rand_core` is conditionally re-exported when the eponymous crate feature is enabled.

Additionally, the `std` crate feature enables `rand_core/std`, which permits access to `OsRng` as `aead::rand_core::OsRng`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -305,6 +305,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "getrandom"
+version = "0.2.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


newpavlov (approved):
Looks good! Although I am not sure about enabling it by default. Right now in most implementation crates we use `default-features = false` on the `aead` dependency, so it will have no effect. It could be better to enable it by default in implementation crates and then forward it to `aead`. It will be similar to how we handle `std` for example in the has crates.

tarcieri:
Removed the default feature.

PR #690, opened 2021-07-20
crypto-mac: pin subtle to v2.4

newpavlov:


newpavlov:
Whoops, I should've updated `crypto-mac` v0.11, not v0.10... Well, since no harm is done, it's fine I guess.

PR #691, opened 2021-07-20
Pin subtle to v2.4 for crypto-mac v0.11

newpavlov:
#690 has pinned `subtle` for `crtypto-mac` v0.10.

### DIFF
--- a/crypto-mac/CHANGELOG.md
+++ b/crypto-mac/CHANGELOG.md
@@ -5,11 +5,23 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.11.1 (2021-07-20)
+### Changed
+- Pin `subtle` dependency to v2.4 ([#691])
+
+[#691]: https://github.com/RustCrypto/traits/pull/691
+
 ## 0.11.0 (2021-04-28)
 ### Changed
 - Bump `cipher` dependency to v0.3 ([#621])
 
 [#621]: https://github.com/RustCrypto/traits/pull/621
+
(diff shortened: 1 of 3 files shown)


newpavlov:
The clippy warning is irrelevant for changes in this PR.

bridiver:
was there a specific reason for this? We needed to bump subtle to 2.6 and it ended up taking us back to 0.11.0 from 0.11.1

newpavlov:
It was done in an attempt to preserve MSRV of the crate. Later we found this approach to be flawed and no longer use it.

I would suggest to update your dependencies. `crypto-mac` was deprecated a fair time ago.

bridiver:
> I would suggest to update your dependencies. `crypto-mac` was deprecated a fair time ago.

I wish it were that easy when dealing with transitive dependencies in rust crates, but thanks for the info!

tarcieri:
@bridiver out of curiosity, which crate is still using `crypto-mac`?

bridiver:
> @bridiver out of curiosity, which crate is still using `crypto-mac`?

We have several long chains of deps actually

fvm_shared 0.7 -> libsecp256k1 0.7 -> hmac_drbg 0.3 -> hmac 0.8 -> crypto_hmac 0.8
fvm_shared 0.7 -> bls-signatures 0.15 -> hkdf 0.11 -> hmac 0.11 -> crypto_hmac 0.11

tarcieri:
`hmac-drbg` looks abandoned, having not received updates in over 4 years.

Perhaps we should consider moving the HMAC-DRBG implementation in the `rfc6979` crate into the `hmac` crate so others can use it? /cc @newpavlov 

It's unfortunate `bls-signatures` has not updated to a newer version of `hkdf`

newpavlov:
I think we can create a new HMAC-DRBG crate in the CSPRNs repo.

PR #699, opened 2021-08-02
password-hash: Add pepper to API

### LINKED ISSUE #694: password-hash: Add pepper to API
As discussed in [RustCrypto/password-hashes#194](https://github.com/RustCrypto/password-hashes/issues/194) it could be considered to add pepper with a MAC to the traits as it's currently particularly uncomfortable to have a password-hash with a pepper.
The changes could look like this:
```rust
trait PasswordHasher {
...
    fn hash_password_with_pepper<'a>(
        &self,
        password: &[u8],
        algorithm: Option<Ident<'a>>,
        params: Self::Params,
        salt: impl Into<Salt<'a>>,
        pepper_method: impl Pepper,
    ) -> Result<PasswordHash<'a>>;
}
```

TheAlgorythm:
Resolves #694

### DIFF
--- a/password-hash/src/errors.rs
+++ b/password-hash/src/errors.rs
@@ -57,6 +57,8 @@ pub enum Error {
 
     /// Invalid algorithm version.
     Version,
+
+    Pepper,
 }
 
 impl fmt::Display for Error {
@@ -78,6 +80,7 @@ impl fmt::Display for Error {
             Self::SaltTooShort => write!(f, "salt too short"),
             Self::SaltTooLong => write!(f, "salt too long"),
             Self::Version => write!(f, "invalid algorithm version"),
+            Self::Pepper => write!(f, "pepper"),
         }
     }
 }
--- a/password-hash/src/errors.rs
+++ b/password-hash/src/errors.rs
@@ -58,6 +58,7 @@ pub enum Error {
     /// Invalid algorithm version.
     Version,
 
+    /// An error during peppering.
     Pepper,
 }
 
--- a/password-hash/src/lib.rs
+++ b/password-hash/src/lib.rs
@@ -107,6 +107,9 @@ pub trait PasswordHasher {
         )
     }
 
+    /// Compute a [`PasswordHash`] with the given algorithm [`Ident`]
+    /// (or `None` for the recommended default), password, salt,
+    /// parameters and peppering.
     fn hash_password_with_pepper<'a, P>(
         &self,
         password: &[u8],
@@ -122,11 +125,7 @@ pub trait PasswordHasher {
 
         hash.hash = hash
             .hash
-            .map(|hash| {
-                pepper_method
-                    .pepper(hash.as_bytes())
-                    .map_err(|_e| Error::Pepper)
-            })
+            .map(|hash| pepper_method.pepper(hash).map_err(|_e| Error::Pepper))
             .transpose()?;
         hash.pepper_algorithm = Some(pepper_method.ident());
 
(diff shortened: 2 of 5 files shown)


password-hash/src/errors.rs
```diff
@@ -57,6 +57,9 @@ pub enum Error {
 
     /// Invalid algorithm version.
     Version,
+
+    /// An error during peppering.
+    Pepper,
```
TheAlgorythm, line 62: I don't know how to use `Pepper::Error` in `no_std`

password-hash/src/lib.rs
```diff
@@ -185,13 +212,14 @@ pub trait McfHasher {
 /// PHC strings have the following format:
 ///
 /// ```text
-/// $<id>[$v=<version>][$<param>=<value>(,<param>=<value>)*][$<salt>[$<hash>]]
+/// $<id>[$v=<version>][$pepper=<pepper_algorithm>][$<param>=<value>(,<param>=<value>)*][$<salt>[$<hash>]]
```
TheAlgorythm, line 215: How should we call the field? A shorter name would be good but it shouldn't interfere with a `param` key.

password-hash/src/pepper.rs
```diff
@@ -0,0 +1,16 @@
...
+    type Error;
+
+    /// Peppers the `data`.
+    fn pepper(&self, data: Output) -> Result<Output, Self::Error>;
+
+    /// Verifies that the `data` is peppered to the same as `to_verify`.
+    fn verify(&self, data: Output, to_verify: &Output) -> Result<(), Self::Error>;
+
+    /// Identifies the algorithm uniquely.
+    fn ident<'a>(&'a self) -> Ident<'a>;
```
TheAlgorythm, line 15: I would like to make it `fn ident() -> Ident<'static>` but eg. the `str "HMAC-SHA256"` can only produced by something hacky like [const-concat](https://lib.rs/crates/const-concat) or the a bit more professional [const_format](https://lib.rs/crates/const_format).

password-hash/src/lib.rs
tarcieri: These fields are part of the [PHC string format](https://github.com/P-H-C/phc-string-format/blob/master/phc-sf-spec.md), so it doesn't make sense to add additional fields to it.

If we are going to include information about pepper, it should probably be part of the algorithm identifier field.

Note that certain algorithms, e.g. Argon2, already have built-in support for pepper.

TheAlgorythm: I thought you [pleaded](https://github.com/RustCrypto/traits/issues/694#issuecomment-889138538) for not using the algorithm identifier field as it would violate the PHC format.

What do you recommend?

tarcieri: Adding a new bespoke field is worse. Regardless of what the solution is, it shouldn't work outside the scope of the PHC format.

The spec does not encode the pepper into the hash whatsoever.

tarcieri:
My gut feeling, looking over this, is pepper is something that each password hashing algorithm should optionally be able to define on an algorithm-by-algorithm basis.

Argon2 already has it built-in.

password-hash/src/lib.rs
TheAlgorythm: You mean only the hash and no other fields, or?
So it would be pending on a potential PHC update.

tarcieri: The spec already mentions pepper. The example Argon2 hash includes pepper, however it does not change how the hash is serialized.

tarcieri:
The more I think about this, the more I feel like perhaps pepper doesn't need any API changes.

It's really a property of individual algorithms, and the `PasswordHasher` trait already uses `&self` to support some configuration/initialization of an algorithm. The [`argon2` crate already uses this to support pepper in the form of an optional `secret` field](https://docs.rs/argon2/0.2.2/argon2/struct.Argon2.html#method.new).

TheAlgorythm:
> My gut feeling, looking over this, is pepper is something that each password hashing algorithm should optionally be able to define on an algorithm-by-algorithm basis.

Imho this solution would be more flexible but as it requires external changes, I think I should pause/stop this PR.

Issue #706, opened 2021-08-22
Salt with invalid character fails with error `SaltToLong`


Szymongib:
### Overview 

When trying to create `Salt` via `Salt::new(...)` with an invalid PHC character, the operation fails with the `Error::SaltToLong` error which was a little confusing until I looked into the code here:
https://github.com/RustCrypto/traits/blob/master/password-hash/src/salt.rs#L108

Example salt to reproduce: `1234_mysalt`.

From the left todo, I assume it might have been a known issue.

### Possible solution

Perhaps the `Error::ParamValueInvalid` could be extended to contain the cause of the invalid value:
```rust
...
pub enum Error {
...
    ParamValueInvalid(ParamValueError)
...
}

pub enum ParamValueError {
    ToLong,
    ToShort,
    InvalidChar,
    // ... and any other possible issues
}
```

I can try to draft the PR if there is any interest in this solution, otherwise happy to discuss further 🙂

tarcieri:
#707 should at least make this return `ParamValueInvalid`.

It'd be nice to add something like `Error::SaltInvalid`, but it'd be a breaking change because `Error` was not defined as `non_exhaustive`

Your idea of an enum for the various problem is good as well, although also a breaking change. Feel free to open a PR, and we can get it in the next breaking release.

PR #713, opened 2021-08-24
feat: Add more details to `ParamValueInvalid` and introduce `SaltInvalid` error

Szymongib:
### Overview

As briefly discussed in #706 this PR:
- Adds more context to `Error::ParamValueInvalid` in from of `ParamValueError` enum.
- Introduces `Error::SaltInvalid`.
- Maps `Error::ParamValueInvalid` to proper salt error.

Also as discussed in the issue **this is a breaking change.**

### Notes

`ParamValueError` could be an an `Option` in `ParamValueInvalid`: `ParamValueInvalid(Option<ParamValueError>)` if in some cases we do not want to specify a more detailed error but it might be more cumbersome to handle and I am not sure if there is such need, therefore, I went for implementation without it. Just wanted to highlight it as my knowledge here is limited.

Fixes: #706

### DIFF
--- a/password-hash/src/errors.rs
+++ b/password-hash/src/errors.rs
@@ -33,7 +33,7 @@ pub enum Error {
     ParamNameInvalid,
 
     /// Invalid parameter value.
-    ParamValueInvalid,
+    ParamValueInvalid(ParamValueError),
 
     /// Maximum number of parameters exceeded.
     ParamsMaxExceeded,
@@ -56,8 +56,21 @@ pub enum Error {
     /// Salt too long.
     SaltTooLong,
 
+    /// Salt invalid.
+    SaltInvalid,
+
     /// Invalid algorithm version.
     Version,
+}
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+#[non_exhaustive]
+pub enum ParamValueError {
+    ToLong,
+    ToShort,
+    NotProvided,
+    InvalidChar,
+    InvalidFormat,
 }
 
 impl fmt::Display for Error {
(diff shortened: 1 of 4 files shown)


password-hash/src/errors.rs
```diff
@@ -56,10 +56,23 @@ pub enum Error {
...
+    /// Salt invalid.
+    SaltInvalid,
+
     /// Invalid algorithm version.
     Version,
 }
 
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+#[non_exhaustive]
+pub enum ParamValueError {
```
tarcieri, line 68: What about calling this `InvalidValue` instead, and also parameterizing `Error::SaltInvalid` with it, then eliminating `SaltTooShort` and `SaltTooLong`?

I say this because at least in the current implementation salts and parameter values follow the same rules, so it would be nice to deduplicate the error handling for them.

password-hash/src/errors.rs
```diff
@@ -70,14 +83,27 @@ impl fmt::Display for Error {
...
+                ParamValueError::NotProvided => {
+                    f.write_str("invalid parameter value: required value not provided")
+                }
+                ParamValueError::InvalidChar => {
+                    f.write_str("invalid parameter value: contains invalid character")
+                }
+                ParamValueError::InvalidFormat => {
+                    f.write_str("invalid parameter value: value format is invalid")
+                }
+            },
```
tarcieri, line 98: It seems like this could be more nicely factored as an `fmt::Display` impl on `ParamValueError` (or whatever it gets renamed to, if it's renamed)

password-hash/src/salt.rs
```diff
@@ -134,6 +135,17 @@ impl<'a> Salt<'a> {
...
+fn map_error_to_salt_error(err: Error) -> Error {
+    match err {
+        Error::ParamValueInvalid(value_err) => match value_err {
+            ParamValueError::ToLong => Error::SaltTooLong,
+            ParamValueError::ToShort | ParamValueError::NotProvided => Error::SaltTooShort,
+            _ => Error::SaltInvalid,
+        },
+        _ => Error::SaltInvalid,
+    }
+}
```
tarcieri, line 147: Seems like you could get rid of this if the two error types shared a common inner error enum type as suggested above.

password-hash/src/errors.rs
Szymongib: Yeah, that makes sense and allows to simplify things a little bit. I have adjusted it.

password-hash/src/errors.rs
Szymongib: Oh yeah, that totally makes sense. Added.

password-hash/src/salt.rs
Szymongib: Not sure if we can get rid of it completely but it definitely simplifies a bit. I have removed the new method and did it inline as it is simpler now. Please let me know in case I have missed some way to do it better 🤔

PR #727, opened 2021-08-30
New crate versions, including cipher v0.4

newpavlov:
This PR is a further development of ideas from #589. It depends on RustCrypto/utils#566.

The changes are quite extensive and I plan to describe them here a bit later.


cipher/src/block.rs
```diff
@@ -10,228 +10,398 @@
...
+        post_fn: impl FnMut(InTmpOutBuf<'_, Block<Self>>),
+    ) {
+        blocks.process_chunks::<U1, _, _, _, _, _>(
+            self,
+            pre_fn,
+            post_fn,
+            |state, mut chunk| state.encrypt_block_inout(chunk.get(0)),
+            |state, mut chunk| state.encrypt_block_inout(chunk.get(0)),
+        )
+    }
```
tarcieri, line 40: @newpavlov just as a general statement, this is rather abstract and complicated code with no explanatory comments. I have absolutely no idea what's going on here, especially compared to the straightforward APIs we used to have before.

What are the callbacks for? Why is it using callbacks at all?

What is the advantage over a simple slice-based API?

newpavlov: Yeah, I know that the code severely lacks documentation, but I tried several different approaches and each time documentation got outdated, so eventually I left documentation for the last after design will be mostly finalized (i.e. I should work on documentation soon).

As I've roughly described in the predecessor PR, the main idea is to perform splitting into chunks inside cipher methods. [Here](https://github.com/RustCrypto/utils/blob/pad_error/inout/src/lib.rs#L452-L491) you can see code for `process_chunks`. If we are to forget about the tail processing, then the main code is:
```rust
let (chunks, tail) = self.into_chunks::<N>();
for chunk in chunks {
    let mut tmp = GenericArray::<T, N>::default();
    let mut buf = chunk.extend_with_tmp(&mut tmp);
    let chunk = match pre_fn(buf.reborrow().into_buf()) {
        InSrc::In => buf.reborrow().inout_from_intmp(),
        InSrc::Tmp => buf.reborrow().inout_from_tmptmp(),
    };
    proc_chunk(&mut state, chunk);
    post_fn(buf.into_buf());
}
```
Effectively, the only reason for `pre_fn` existence is implementation of CTR modes (since it has to generate blocks to be encrypted). I guess it also may be useful for AtE modes. But almost always it's simply equal to `|_| InSrc::In`. In other words, usually we encrypt data from the input chunk into the temporary chunk using the `proc_chunk` function (for AES with AES-NI it will be `aesenc8` with `state` being equal to encryption keys), then we process trio of input, temporary, and output chunks in the `post_fn` callback. In the case of the ECB mode `post_fn` would simply copy data from temporary chunk into output chunk. With CTR it XORs data in input and temporary chunks and writes it into output chunk. In other block modes which support parallel block processing `post_fn` can be a bit more involved.

Why do we need the temporary chunk instead of writing immediately into output? For some block modes we need this temporary for implementation, but the main reason is that input and output chunks can represent the same memory if we use in-place methods. In other words, by writing into the output chunk you may the input one as well. I have designed the `inout` crate in a way which allows to do it safely (i.e. you can not get reference to input and mutable reference to output simultaneously).

Since callbacks are generics, they get inlined on monomorphization, which should make it significantly easier for compiler to eliminate redundant memory copies and help with keeping data in registers without spilling them. Also, information about the chunk size is available to compiler at compile time, so it can optimize the callbacks accordingly. Effectively, the callbacks are used to "inject" code into blocks processing loop defined inside cipher. Thus chunking is fully controlled by a cipher implementation. Having such "inverted" loop [allows](https://github.com/RustCrypto/block-ciphers/blob/cipher_v0.4/aes/src/autodetect.rs#L103-L116) us to branch on target features outside of the loop.

[Here](https://github.com/RustCrypto/block-ciphers/blob/cipher_v0.4/aes/src/autodetect.rs#L103-L116) you can see how this method looks in the AES-NI case. It also could be useful to take a look at the [mode crates](https://github.com/RustCrypto/block-ciphers/tree/cipher_v0.4/modes).

tarcieri:
That does look like an impressive performance gain. I'll try to make some time to look through this more over the weekend.

PR #734, opened 2021-09-04
Defined the SignerMut trait

rozbb:
As per our discussion on the RCIG Zulip, there seems to be a need for a stateful version of the `Signer` trait. This lets us support "Key Evolving Signature Schemes" such as [XMSS](https://eprint.iacr.org/2011/484).

One thing that is unclear so far is whether we also want to do stateful versions of `DigestSigner`, `RandomizedSigner`, and `AsyncSigner`. I think the latter two won't be an issue for now, since XMSS is a deterministic scheme, and HSMs do not support XMSS.

### DIFF
--- a/signature/CHANGELOG.md
+++ b/signature/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Re-export `rand_core`. Emit compilation error if unstable functionality
 is enabled by bypassing the preview features. ([#683])
+- Defined the `StatefulSigner` trait
 
 [#683]: https://github.com/RustCrypto/traits/pull/683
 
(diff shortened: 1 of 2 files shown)


tarcieri:
Our other stateful types are named `*Mut`, for example:

- https://docs.rs/aead/0.4.3/aead/trait.AeadMut.html
- https://docs.rs/cipher/0.3.0/cipher/trait.BlockEncryptMut.html
- https://docs.rs/cipher/0.3.0/cipher/trait.BlockDecryptMut.html

rozbb:
Fixed

tarcieri (approved):
This looks fine to me.

Our other `*Mut` traits also have blanket impls of the stateful `*Mut` trait for types which impl the stateless trait, which might also make sense here.

tarcieri:
Thank you!

rozbb:
Sorry one more thing: I just pushed another commit to this branch. It wasn't 100% clear to me that my blanket impl wasn't trivially gonna loop forever. Feel free to copy into the tree

tarcieri:
Oh huh, can you make sure one way or another and open another PR if it's broken?

rozbb:
Will do

rozbb:
Update: it's not broken. I still don't love how it looks (under the hood, the steps it takes are x := &&mut self, y := &**x, y.sign(...)), but you can decide if it's worth changing.

PR #748, opened 2021-09-10
fixed fmt::Display for PasswordHash

TheAlgorythm:
The hash should only be added if there is a salt according to the PHC specification.
```
$<id>[$v=<version>][$<param>=<value>(,<param>=<value>)*][$<salt>[$<hash>]]
```
This shouldn't be a vulnerability as it is **not** in the parser. Even though the parser code looks the same it shouldn't be susceptible due to the use of a iterator.


tarcieri:
Good catch, thanks!

PR #768, opened 2021-09-17
elliptic-curve: add `Reduce` trait

tarcieri:
Adds a trait for modular reduction, primarily intended for use cases like hashing to a field element.

The trait is generic around a `UInt` to permit overlapping impls, allowing either "narrow" or "wide" reduction, selected by the size of the input.


tarcieri:
Will go ahead and merge this so we can try it out. Worst case we can back it out prior to another release.

PR #782, opened 2021-10-02
elliptic-curve: make `FromEncodedPoint` return a `CtOption`

tarcieri:
Internally `k256`/`p256` return a `CtOption` for this method, then convert to an `Option`.

It would be helpful when implementing other traits from the `group` crate, e.g. `GroupEncoding`, if this instead returned a `CtOption`.

### DIFF
--- a/elliptic-curve/src/dev.rs
+++ b/elliptic-curve/src/dev.rs
@@ -367,8 +367,13 @@ impl ConstantTimeEq for AffinePoint {
 }
 
 impl ConditionallySelectable for AffinePoint {
-    fn conditional_select(_a: &Self, _b: &Self, _choice: Choice) -> Self {
-        unimplemented!();
+    fn conditional_select(a: &Self, b: &Self, choice: Choice) -> Self {
+        // Not really constant time, but this is dev code
+        if choice.into() {
+            *b
+        } else {
+            *a
+        }
     }
 }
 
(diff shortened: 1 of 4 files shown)


madninja:
I believe a corresponding change to ToCompactEncoding  to return a CtOption instead of an Option would help with chaining key construction. The complexity in switching between CtOption and Option in the steps in: https://github.com/helium/helium-crypto-rs/pull/20/files#diff-522b3f5cfeb12570fd7f3eb9445fbe8a5151dadd10e4dc51ca79d2de9b51c9deR211-R230 doesn't smell right, does it.

I tried putting together a patch here but got lost in the elliptic_curves crate picking up that patch without newer commits creeping in

tarcieri:
Yes, sorry, this got missed when `group` and the other APIs switched to `CtOption`.

Regarding that code example, there's an impl of `From<CtOption<T>> for Option<T>`, so you can use `Option::from(ct_option)` to do the conversion (or `ct_option.into()` if type inference permits it)

madninja:
> Yes, sorry, this got missed when `group` and the other APIs switched to `CtOption`.

I guess I'll await the change to trickle in since I couldn't get the parts to fit together right trying to make a PR for the two relevant crates. 

> Regarding that code example, there's an impl of `From<CtOption<T>> for Option<T>`, so you can use `Option::from(ct_option)` to do the conversion (or `ct_option.into()` if type inference permits it)

Ah yes! I missed that, updated. Thanks for the pointer

madninja:
> It's a breaking change, so we can't fix it until the next minor version bump

completely understand

PR #795, opened 2021-10-21
elliptic-curve: Rust 2021 edition upgrade

tarcieri:
The RustCrypto/formats repo was bumped to Rust 2021 edition, and some of those crates (e.g. `der`, `sec1`) are hard dependencies for the `elliptic-curve` crate:

https://github.com/RustCrypto/formats/pull/136

This is a corresponding Rust 2021 edition bump for the `elliptic-curve` crate, as well as the omnibus `crypto` crate whose MSRV is determined by the highest MSRV crate in this repo.


tarcieri:
Oh wow, that's annoying. It seems we do need to bump MSRV of entire repos at a time.

Given that, I'm going to leave `elliptic-curve` (and `crypto`) as 2018 edition crates, but bump their MSRV as they now use 2021 edition dependencies.

PR #800, opened 2021-11-10
Return early if `Ident` is too long

noncombatant:
At this point, we (can) know whether `too_long` is true, so don't iterate over all the `bytes`. (Both invalid chars and excessive length result in the same kind of error.) I don't think I see a constant-time/timing-independence concern — if a malicious input is too long, saying so immediately does not give the attacker any information about the `bytes` value.


tarcieri:
Thanks!

PR #837, opened 2021-12-07
Add optional support for bytes crate

semtexzv:
The `BytesMut` struct from `bytes` crate already provides an interface that is basically same as the `Buffer` trait.

This PR adds an optional dependency on the bytes crate, and implements the buffer for the `BytesMut`. This should be useful when writing network protocols, that require encryption.

### DIFF
--- a/aead/Cargo.toml
+++ b/aead/Cargo.toml
@@ -20,6 +20,7 @@ generic-array = { version = "0.14", default-features = false }
 # optional dependencies
 blobby = { version = "0.3", optional = true }
 heapless = { version = "0.7", optional = true, default-features = false }
+bytes = { version = "1.1.0", optional = true }
 rand_core = { version = "0.6", optional = true }
 
 [features]
--- a/aead/src/lib.rs
+++ b/aead/src/lib.rs
@@ -43,6 +43,10 @@ pub use generic_array::{self, typenum::consts};
 #[cfg_attr(docsrs, doc(cfg(feature = "heapless")))]
 pub use heapless;
 
+#[cfg(feature = "bytes")]
+#[cfg_attr(docsrs, doc(cfg(feature = "heapless")))]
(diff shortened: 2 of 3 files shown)


aead/src/lib.rs
```diff
@@ -43,6 +43,10 @@ pub use generic_array::{self, typenum::consts};
 #[cfg_attr(docsrs, doc(cfg(feature = "heapless")))]
 pub use heapless;
 
+#[cfg(feature = "bytes")]
+#[cfg_attr(docsrs, doc(cfg(feature = "heapless")))]
```
tarcieri, line 47:
```suggestion
#[cfg_attr(docsrs, doc(cfg(feature = "bytes")))]
```

aead/Cargo.toml
```diff
@@ -20,6 +20,7 @@ generic-array = { version = "0.14", default-features = false }
 # optional dependencies
 blobby = { version = "0.3", optional = true }
 heapless = { version = "0.7", optional = true, default-features = false }
+bytes = { version = "1.1.0", optional = true }
```
tarcieri, line 23: Perhaps...

```suggestion
bytes = { version = "1.1.0", optional = true, default-features = false }
```

...which will avoid linking `std`

aead/src/lib.rs
```diff
@@ -52,6 +56,8 @@ use generic_array::{typenum::Unsigned, ArrayLength, GenericArray};
 
 #[cfg(feature = "alloc")]
 use alloc::vec::Vec;
+use bytes::BytesMut;
+use std::io::Bytes;
```
tarcieri, line 60: These aren't properly feature gated, and really if you're adding `bytes` support, I think it should be able to support an `alloc`-only profile

tarcieri:
Implemented in #956

Issue #498, opened 2021-01-29
Key exchange / KEM traits


jyao1:
May I know if there is any key_exchange traits defined, or to be defined in the future?

For example, the finite field diffie-hellman ephemeral (FFDHE) or ecliptic curve diffie-hellman ephemeral (ECDHE) - https://github.com/briansmith/ring/blob/main/src/agreement.rs

tarcieri:
The `elliptic-curve` crate has an `ecdh` crate feature/API which supports both static DH and ECDHE:

https://docs.rs/elliptic-curve/latest/elliptic_curve/ecdh/index.html

See the `p256` for concrete usage:

https://docs.rs/p256/latest/p256/ecdh/index.html

We don't presently have any sort of implementation of FFDH.

jyao1:
Thank you very much for the quick response.
I am looking for a way to abstract key_exchange, because the crypto agile requirement for both ECDHE and FFDHE.

There are traits for AEAD/Digest/Signature/MAC. HKDF is discussed in https://github.com/RustCrypto/traits/issues/5
It seems only KeyExchange trait is not there. Is that something this project want to add in the future?

tarcieri:
Sure, it's something we should eventually address

tarcieri:
For various protocols, it may make sense to define [KEM](https://en.wikipedia.org/wiki/Key_encapsulation) traits. Protocols like [HPKE](https://datatracker.ietf.org/doc/draft-irtf-cfrg-hpke/) are defined in terms of KEMs, and it's the API shape being targeted by post-quantum crypto standardization processes.

rozbb:
Hi all, I was thinking about what a set of KEM traits would look like. My starting point was the `signature` traits since they're very similar.

I'd love to have feedback on this initial foray into KEM traits. I suspect the first option is the better one, but it's not as ergonomic as I'd like it to be.

### A note: AuthEncap

First, we need "sender-authenticated key encapsulation", as in HPKE's [`AuthEncap()`](https://cfrg.github.io/draft-irtf-cfrg-hpke/draft-irtf-cfrg-hpke.html#section-4.1), which, in Diffie-Hellman land, involves a static-eph + eph-static key exchange.

One current problem is in generalizability to pq-crypto: I don't think there is any known non-interactive authenticated key exchange for post-quantum primitives. That said, there are plenty of such exchanges for non-pq primitives, with different properties (everything from static-static ECDH up to X3DH with an optional 4th eph-otk exchange).

[...]

PR #845, opened 2021-12-13
KEM Traits

rozbb:
Going off our [previous discussion](https://github.com/RustCrypto/traits/issues/498), I took a crack at implementing some KEM traits.

Sample usage, from the HPKE unit test:

```rust
// Make a sender and recipient keypair. gen_keypair is completely separate from the KEM traits
let (sk_sender, pk_sender) = gen_keypair(&mut rng);
let (sk_recip, pk_recip) = gen_keypair(&mut rng);

// Try an unauthed encap first. Check that the derived shared secrets are equal
let encapper = X25519Encap;
let (ek, ss1) = encapper.try_encap(&mut rng, &pk_recip).unwrap();
let ss2 = sk_recip.try_decap(&ek).unwrap();
assert_eq!(ss1, ss2);

// Now do an authenticated encap
let encapper = X25519AuthEncap(sk_sender, pk_sender.clone());
let (ek, ss1) = encapper.try_encap(&mut rng, &pk_recip).unwrap();
let ss2 = sk_recip.try_auth_decap(&ek, &pk_sender).unwrap();
assert_eq!(ss1, ss2);

// Now do an invalid authenticated encap, where the sender uses the wrong private key. This
// should produce unequal shared secrets.
let (rand_sk, _) = gen_keypair(&mut rng);
let encapper = X25519AuthEncap(rand_sk, pk_sender.clone());
let (ek, ss1) = encapper.try_encap(&mut rng, &pk_recip).unwrap();
let ss2 = sk_recip.try_auth_decap(&ek, &pk_sender).unwrap();
assert_ne!(ss1, ss2);

```

## Notes

* I made `EncappedKey` hold other KEM data in the form of associated types. I don't know of any cases where an encapsulated key type could and should be compatible with multiple pubkey or shared-secret types, so I think this is OK. It also makes the interface a lot cleaner.
* ~~I wrote `Serializable` and `Deserializable` traits because `PublicKey: AsRef<[u8]>` is not generally true (e.g., when `PublicKey` is a compressed curve point and its serialized format must be uncompressed). For the same reason, I require `EncappedKey`s to be serializable.~~ I actually disagree with this move. I reverted this and made `EncappedKey` a bag of bytes again. I also removed all trait bounds on `PublicKey`. We might want to make some guarantees about `PublicKey` though. Open to ideas.
* I dropped `AuthEncapsulator` altogether. I realized that you can just define `sk.try_encap()` to do the authenticated encapsulation. Thus, there's just 1 `Encapsulator` trait.
* In the same vein, you could also drop `AuthDecapsulator` by just folding the sender pubkey into the decapsulator itself. I don't think this makes much sense, though, since passing around a `Decapsulator` that only works for a single sender doesn't seem like a reasonable way to use a KEM interface.
*  I haven't written the infallible `encap()`, `decap()`, and `decap_auth()` methods yet. This is simple though.
* Besides the HPKE test I plan on writing a test using the KEMs in `pqcrypto`. Eventually, those impls will be added to `pqcrypto-compat`.


rozbb:
One issue right now is that `EncappedKey::PublicKey` is supposed to represent both the recipient's public key and the sender's public key used for authentication. These are not the same key in general.

For example, in X3DH, the "public key" you encapsulate to is actually a tuple of an identity pubkey, a medium-term pubkey, and a signed one-time-prekey. The "public key" you use when auth-decapsulating is _just_ an identity pubkey.

A simple fix is to split `EncappedKey::PublicKey` into two types: `SenderPublicKey` and `RecipPublicKey`

rozbb:
Ok I wrote 3 test implementations and I'm pretty happy with how it looks so far. I have it working for a very simple unauthenticated KEMs (Saber), a complicated authenticated KEMs (X3DH), and somewhere in between (HPKE's DHKEM)

kem/tests/saber.rs
```diff
@@ -0,0 +1,75 @@
+#[cfg(test)]
+mod tests {
```
tarcieri, line 2: This is implicit for any tests in `tests/`

kem/tests/x3dh.rs
```diff
@@ -0,0 +1,144 @@
+#[cfg(test)]
+mod tests {
```
tarcieri, line 2: Likewise here... this can be removed

kem/README.md
```diff
@@ -0,0 +1,61 @@
...
+[![Build Status][build-image]][build-link]
+
+This crate contains traits which provide generic, object-safe APIs for
+establishing shared secrets using [a key encapsulation mechanism][1].
+
+[Documentation][docs-link]
+
+## Minimum Supported Rust Version
+
+Rust **1.41** or higher.
```
tarcieri, line 17: Not sure what actual MSRV you have in mind, but this doesn't match what's in lib.rs, which is 1.56.

If you make this a 2021 edition crate, you can unify the toplevel rustdoc with what's in README.md, as it were.

rozbb:
Thanks. Yeah those paragraphs were all stolen from the signature crate. Will fix

rozbb:
I made the edition 2021 and standardized on whatever rustc version I'm running right now (1.56.1 2021-11-01). Is there a more principled way of picking an MSRV I should do?

tarcieri:
We usually specify MSRV in terms of a minor version like 1.56 instead of 1.56.0. That works with the `rust-version` attribute as well.

In terms of selecting a version, Rust 2021 is MSRV 1.56 by definition, so that’s that.

kem/src/lib.rs
```diff
@@ -0,0 +1,73 @@
...
+//! such a parameter for the return value, allowing it to be inferred by the type checker based on
+//! the desired KEM algorithm.
+//!
+//! The [`Signature`] trait is requires `as_bytes()` and `from_bytes()` methods, enforcing that
+//! encapsulated key types are thin wrappers around a "bag-of-bytes" serialization. Inspiration for
+//! this approach comes from the Ed25519 signature system, which was based on the observation that
+//! past systems were not prescriptive about how signatures should be represented on-the-wire, and
+//! that lead to a proliferation of different wire formats and confusion about which ones should be
+//! used. This crate aims to provide similar simplicity by minimizing the number of steps involved
+//! to obtain a serializable encapsulated key.
```
tarcieri, line 54: You can use the following on Rust 1.56 to import the README.md into the rustdoc (and also doctest any examples in the README!)

```rust
#![doc = include_str!("../README.md")]
```

If you'd like to add additional toplevel rustdoc this syntax works:

```rust
#![doc = include_str!("../README.md")]

//! # Extra stuff here
//! 
//! This just shows up as additional concatenated Markdown
```

rozbb:
I think I updated the deps in `Cargo.lock` beyond where they should've gone. This seems delicate. Should I just not include `Cargo.lock` in the diff at all?

Also separately I thought `Cargo.lock` wasn't really included in libs. It's not present in a few of the other RustCrypto libs I checked.

tarcieri:
>  Should I just not include Cargo.lock in the diff at all?

I would suggest only including relevant changes, e.g. adding a new crate to the workspace.

> Also separately I thought Cargo.lock wasn't really included in libs. It's not present in a few of the other RustCrypto libs I checked.

This gets asked often enough we should probably write a FAQ.

While not checking in Cargo.lock for libraries is a common rule-of-thumb, it has resulted in some extremely hard-to-debug situations for us in the past.

Just as an example, we recently tried to debug a situation where Ed25519 signature verification was failing for an individual. Nobody else could reproduce it, but it was fully reproducible for that individual.

Turns out their Cargo.lock contained a release of the `sha2` crate with a miscomputation bug. Since we had patched the bug, anyone who had an older Cargo.lock with an earlier release didn't see the bug, and anyone who cloned the repo fresh and got the version with the bugfix didn't see the bug. But this perplexed several people for days.

Including Cargo.lock in the repo makes the build fully reproducible and eliminates these sorts of hard-to-debug scenarios.

rozbb:
Ok, updated the lock file. I just copied the existing one and ran `cargo check`. I could use some help on the failing builds. Why is, e.g., `digest` failing to build because of the `edition = "2021"` key in `kem`?

tarcieri:
Unfortunately if you have any 2021 edition crates in the workspace, the MSRV of the whole workspace becomes 1.56.

The workaround is to keep 2021 edition crates out of that workspace. Here's the hack I'm using with the `elliptic-curve` crate:

https://github.com/RustCrypto/traits/blob/master/elliptic-curve/Cargo.toml#L18-L20

Hopefully we can bump edition of all of the other crates in this repo in the next minor release (or in the case of `cipher`, maybe the release-after-next), but in the meantime, I think it's ok to copy this hack into `kem` as well.

rozbb:
Should be good now

tarcieri:
Looks like a promising start. Can you remove WIP? (or I can)

tarcieri:
Thank you!

PR #849, opened 2021-12-14
cipher v0.4, digest v0.10.2, and crypto-common v0.1.2 

### LINKED ISSUE #31: Consider supporting buffer-to-buffer cipher operations

### LINKED ISSUE #272: cipher: merge block-modes crate
@o-klp has kindly shared with us ownership of the `cipher` crate.

I think it could be reasonable to merge traits from the listed crates into a single crate, this would reduce size of dependency tree a bit in some cases and will allow us to naturally introduce a trait (and support types) covering all kinds of non-authenticating ciphers.

### LINKED ISSUE #890: Enable `generic-array`'s `more_lengths` feature (`digest` crate)
`generic-array`'s feature `more_lengths` enables more `Into<[T; $n]>` implementations which are useful when converting the result to an array. Right now it's only possible to do this if the hash size is at most 32 bytes. With this feature it's possible to also do it up to size 64 and some select sizes > 64, see the code below.

https://github.com/fizyk20/generic-array/blob/d4d63b7ea425f42a7e373510f95eb36ed17d34c4/src/impls.rs#L218-L251

https://github.com/RustCrypto/traits/blob/d3e544885f0f8c6bd07658470d6f5f6fd0a95b57/digest/Cargo.toml#L15

newpavlov:
Split off from #727

Closes https://github.com/RustCrypto/block-ciphers/issues/131
Closes #272
Closes #31
Closes #890


cipher/Cargo.toml
```diff
@@ -17,6 +18,7 @@ inout = "0.1"
 
 # optional dependencies
 blobby = { version = "0.3", optional = true }
+zeroize = { version = ">=1, <1.6", optional = true, default-features = false }
```
tarcieri, line 21: Realistically I don't foresee bumping `zeroize`'s MSRV above 1.56 any time in the foreseeable future (it's presently 1.51), so I think we can get rid of this pinning.

```suggestion
zeroize = { version = "1", optional = true, default-features = false }
```

cipher/src/stream_wrapper.rs
```diff
@@ -0,0 +1,231 @@
...
+            buffer: Default::default(),
+            pos: 0,
+        }
+    }
+}
+
+#[cfg(feature = "zeroize")]
+impl<T> Zeroize for StreamCipherCoreWrapper<T>
+where
+    T: BlockSizeUser + Zeroize,
```
newpavlov, line 222: @tarcieri 
IIUC we can not implement `ZeroizeOnDrop` for this wrapper, so users would have to rely on manually wrapping it with `Zeroizing`, correct?

tarcieri: You could potentially impl a `Drop` handler that calls `zeroize` on the fields, as well as manually impl `ZeroizeOnDrop` as a marker trait, but it would be feature-gated, which would be a little weird.

Impling `ZeroizeOnDrop` would require bumping the `zeroize` version to 1.5.

newpavlov: No, the issue is that `Drop` impl requires the same bounds as struct definition. We would have to bind the wrapper on `T: Zeroize` with enabled `zeroize` feature, meaning that build could break if it's used with a type which does not implement `Zeroize`.

tarcieri: Aah yes, the bounds need to be consistent between the `Drop` impl and the struct

newpavlov: I've decided to implement zeroizing `Drop`, but not `Zeroize` itself. It looks like we should not implement both of them simultaneously. Having `Zeroize` implemented on complex types (e.g. cipher states) can result in users erroneously zeroizing them, which can lead to security vulnerabilities.

The `Drop` impl zeroizes only buffer and cursor, while zeroization of the core type `T` is left for its own `Drop` impl. Implementation of `ZeroizeOnDrop` is gated on `T: ZeroizeOnDrop`.

saiintbrisson:
Hi! I've been facing some issues while migrating one of my old codes to the new cipher v0.4, and I can´t really find a solution when it comes to the AsyncStreamCipher changes. Is there a reason for it to consume `self` now?

tarcieri:
@saiintbrisson see discussion on #964

PR #850, opened 2021-12-14
signature: bump `digest` dependency to v0.10

tarcieri:


signature/derive/src/lib.rs
```diff
@@ -24,7 +24,7 @@ fn derive_signer(mut s: synstructure::Structure) -> TokenStream {
             Self: signature::DigestSigner<S::Digest, S>
         {
             fn try_sign(&self, msg: &[u8]) -> Result<S, signature::Error> {
-                self.try_sign_digest(S::Digest::new().chain(msg))
+                self.try_sign_digest(S::Digest::new().chain_update(msg))
```
newpavlov, line 27: You can use the new `Digest::new_with_prefix` method.

PR #854, opened 2021-12-22
feat: add osswu trait

mikelodder7:
As requested in [PR 495](https://github.com/RustCrypto/elliptic-curves/pull/495) this upstreams the OSSWU method.


elliptic-curve/src/osswu.rs
```diff
@@ -0,0 +1,84 @@
...
+    /// The Z parameter
+    pub z: F,
+}
+
+/// Trait for determining the parity of the field
+pub trait Sgn0: Field {
+    /// Return the parity of the field
+    /// 1 == negative
+    /// 0 == non-negative
+    fn sgn0(&self) -> u8;
```
tarcieri, line 26: How about...

```suggestion
    fn sgn0(&self) -> Choice;
```

...since that seems how it's used below anyway?

elliptic-curve/src/lib.rs
```diff
@@ -94,6 +94,10 @@ pub mod ecdh;
 #[cfg(feature = "jwk")]
 mod jwk;
 
+/// Optimized simplified Shallue-van de Woestijne-Ulas methods
+#[cfg(feature = "osswu")]
```
tarcieri, line 98:
```suggestion
#[cfg(feature = "osswu")]
#[cfg_attr(docsrs, doc(cfg(feature = "osswu")))]
```

elliptic-curve/src/osswu.rs
```diff
@@ -0,0 +1,84 @@
...
+}
+
+/// The optimized simplified Shallue-van de Woestijne-Ulas method
+/// for mapping elliptic curve scalars to affine points.
+pub trait OsswuMap: Sgn0 {
```
tarcieri, line 31: I think it might be more straightforward if this were:

```suggestion
pub trait OsswuMap: Field + Sgn0 {
```

...since this trait is the one that actually uses methods of `Field`.

I think you could debatably remove the field bound from `Sgn0` in that case as well.

tarcieri (approved):
Looks good! I can go ahead and merged if there's no additional changes you'd like to make.

mikelodder7:
Go ahead. I'm good.

PR #855, opened 2021-12-22
feat: add hash to field

mikelodder7:
Upstream hash2field as described in [PR 495](https://github.com/RustCrypto/elliptic-curves/pull/495)

### DIFF
--- a/elliptic-curve/Cargo.toml
+++ b/elliptic-curve/Cargo.toml
@@ -29,6 +29,7 @@ zeroize = { version = "1", default-features = false }
 
 # optional dependencies
 base64ct = { version = "1", optional = true, default-features = false }
+digest = { version = "0.9", optional = true, default-features = false }
 ff = { version = "0.11", optional = true, default-features = false }
 group = { version = "0.11", optional = true, default-features = false }
 hex-literal = { version = "0.3", optional = true }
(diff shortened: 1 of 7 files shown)


elliptic-curve/Cargo.toml
```diff
@@ -47,6 +48,7 @@ arithmetic = ["ff", "group"]
 bits = ["arithmetic", "ff/bits"]
 dev = ["arithmetic", "hex-literal", "pem", "pkcs8"]
 ecdh = ["arithmetic"]
+hashing = ["digest"]
```
tarcieri, line 51: Elsewhere in RustCrypto we generally just keep `digest`-related functionality gated behind the `digest` feature

elliptic-curve/src/hash2field/expand_msg_xmd.rs
```diff
@@ -0,0 +1,72 @@
...
+    generic_array::{typenum::Unsigned, GenericArray},
+    BlockInput, Digest,
+};
+use subtle::{Choice, ConditionallySelectable};
+
+/// Placeholder type for implementing expand_message_xmd based on a hash function
+#[derive(Debug)]
+pub struct ExpandMsgXmd<HashT> {
+    phantom: PhantomData<HashT>,
+}
```
tarcieri, line 13: Not sure I quite understand the purpose of this

tarcieri: Oh wait, I see, it's to make the blanket impl work.

Curious if there's a better way to get this done, but in the interests of upstreaming I guess this is probably fine for now.

tarcieri:
Going to go ahead and merge this. Can fix up `hashing` => `digest` separately.

While I do have some general concerns about how this is structured, I think trying to play with some alternatives really needs to have everything wired up with concrete implementations in the `elliptic-curve` crates to see what actually works and what doesn't.

PR #857, opened 2021-12-30
Add impl of `Mul` for `NonZeroScalar` * `NonZeroScalar`

fjarri:
It is convenient to be able to do that without unwraps in the client code. The only problem I see is that it assumes that the scalar operations are modulo a prime - is that always the case?

### DIFF
--- a/elliptic-curve/src/scalar/nonzero.rs
+++ b/elliptic-curve/src/scalar/nonzero.rs
@@ -9,7 +9,7 @@ use crate::{
 };
 use core::{
     fmt,
-    ops::{Deref, Neg},
+    ops::{Deref, Mul, Neg},
     str,
 };
 use crypto_bigint::{ArrayEncoding, Integer};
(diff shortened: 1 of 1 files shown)


elliptic-curve/src/scalar/nonzero.rs
```diff
@@ -199,6 +199,36 @@ where
...
+{
+    type Output = Self;
+
+    fn mul(self, other: Self) -> Self {
+        // Assumes that the multiplication is modulo a prime,
+        // so the product of two non-zero scalars is also non-zero.
+        let scalar = self.scalar * other.scalar;
+        debug_assert!(!bool::from(scalar.is_zero()));
+        NonZeroScalar { scalar }
+    }
```
tarcieri, line 214: It should be possible to implement one of these in terms of the other, and tag it `#[inline]` so the code isn't repeated between them.

PR #863, opened 2022-01-03
gate behind hashing => digest

mikelodder7:

PR #865, opened 2022-01-05
feat: add traits for group digest

mikelodder7:
Traits for hash to curve in groups and fields


elliptic-curve/src/group_digest.rs
```diff
@@ -0,0 +1,86 @@
...
+/// Convert a set of field elements into a point
+pub trait ConvertToPoint {
+    /// The field element that represents a point
+    type FieldElement: Field;
+    /// The result of converting two field elements to a Point
+    type Output;
+
+    /// Converts two coordinates into a point
+    fn convert_to_point(x: Self::FieldElement, y: Self::FieldElement) -> Self::Output;
+}
```
daxpedda, line 16: I think there is a missing trait `MapToCurve` that should replace this. It should represent `map_to_curve` from the spec.

daxpedda, line 16: I think there is a missing trait `MapToCurve` that should replace this. It should represent `map_to_curve` from the spec.

elliptic-curve/src/group_digest.rs
```diff
@@ -0,0 +1,86 @@
...
+
+/// Method for points to be cleared by the curve cofactor
+pub trait ClearCofactor {
+    /// Multiply this point by the curve cofactor to ensure
+    /// correct subgroup membership
+    fn clear_cofactor(&mut self);
+}
```
daxpedda, line 23: I think this can be removed in favor of [`CofactorGroup`](https://docs.rs/group/0.11.0/group/cofactor/trait.CofactorGroup.html#tymethod.clear_cofactor)?

elliptic-curve/src/group_digest.rs
```diff
@@ -0,0 +1,86 @@
...
+
+/// Adds hashing arbitrary byte sequences to a valid group element
+pub trait GroupDigest {
+    /// The field element representation for a group value with multiple elements
+    type FieldElement: FromOkm + OsswuMap + Isogeny;
+    /// The resulting group element
+    type Output: Group
+        + ClearCofactor
+        + ConvertToPoint<FieldElement = Self::FieldElement, Output = Self::Output>;
```
daxpedda, line 32: I think this was done to avoid additional `where` bounds because associated type bounds isn't stable?
As pointed above I think a missing trait here would be `MapToCurve` that would fill the gap without requiring additional bounds on the `FieldElement`.

elliptic-curve/src/group_digest.rs
```diff
@@ -0,0 +1,86 @@
...
+    /// hash function is used.
+    ///
+    /// Example using a fixed size hash function
+    ///
+    /// let pt = ProjectivePoint::hash_from_bytes::<hash2field::ExpandMsgXmd<sha2::Sha256>>(b"test data", b"CURVE_XMD:SHA-256_SSWU_RO_");
+    ///
+    ///
+    /// Example using an extendable output function
+    ///
+    /// let pt = ProjectivePoint::hash_from_bytes::<hash2field::ExpandMsgXof<sha3::Shake256>>(b"test data", b"CURVE_XOF:SHAKE-256_SSWU_RO_");
```
daxpedda, line 51:
```suggestion
    /// # Examples
    /// Using a fixed size hash function:
    /// ```
    /// let pt = ProjectivePoint::hash_from_bytes::<hash2field::ExpandMsgXmd<sha2::Sha256>>(b"test data", b"CURVE_XMD:SHA-256_SSWU_RO_");
    /// ```
    ///
    /// Using an extendable output function:
    /// ```
    /// let pt = ProjectivePoint::hash_from_bytes::<hash2field::ExpandMsgXof<sha3::Shake256>>(b"test data", b"CURVE_XOF:SHAKE-256_SSWU_RO_");
    /// ```
```

elliptic-curve/src/group_digest.rs
```diff
@@ -0,0 +1,86 @@
...
+        let (q1x, q1y) = u[1].osswu();
+        let (r1x, r1y) = Self::FieldElement::isogeny(q1x, q1y);
+        let mut q0 = Self::Output::convert_to_point(r0x, r0y);
+        let mut q1 = Self::Output::convert_to_point(r1x, r1y);
+        // Ideally we could add and then clear cofactor thus saving a multiplication
+        // but the field elements may not reduced by the isogeny modulus
+        // which would result in an incorrect subgroup
+        q0.clear_cofactor();
+        q1.clear_cofactor();
+        q0 + q1
```
daxpedda, line 67: I think the spec should be followed more closely here:
https://www.ietf.org/archive/id/draft-irtf-cfrg-hash-to-curve-13.html#section-3-7
```
hash_to_curve(msg)

Input: msg, an arbitrary-length byte string.
Output: P, a point in G.

Steps:
1. u = hash_to_field(msg, 2)
2. Q0 = map_to_curve(u[0])
3. Q1 = map_to_curve(u[1])
4. R = Q0 + Q1              # Point addition
5. P = clear_cofactor(R)
6. return P
```
Again, I think a `MapToCurve` trait would fill the gap. OSSWU isn't applicable to all kind of groups. I understand that currently this is targeted a P-256 & co., but these traits will be used by other groups too.

elliptic-curve/src/hash2field.rs
```diff
@@ -2,32 +2,32 @@ mod expand_msg;
...
+    fn from_okm(data: &GenericArray<u8, Self::Length>) -> Self;
 }
 
 /// Convert an arbitrary byte sequence according to
 /// <https://tools.ietf.org/html/draft-irtf-cfrg-hash-to-curve-11#section-5.3>
-pub fn hash_to_field<E, T, const L: usize, const COUNT: usize, const OUT: usize>(
-    data: &[u8],
-    domain: &[u8],
-) -> [T; COUNT]
+pub fn hash_to_field<E, T>(data: &[u8], domain: &'static [u8], out: &mut [T])
```
daxpedda, line 21: I think it would be great to add a `hash_to_scalar` method to `GroupDigest`. For example the OPAQUE protocol requires the `hash_to_field` function to derive a `Scalar`.

elliptic-curve/src/group_digest.rs
```diff
@@ -0,0 +1,86 @@
...
+    /// Example using a fixed size hash function
+    ///
+    /// let pt = ProjectivePoint::hash_from_bytes::<hash2field::ExpandMsgXmd<sha2::Sha256>>(b"test data", b"CURVE_XMD:SHA-256_SSWU_RO_");
+    ///
+    ///
+    /// Example using an extendable output function
+    ///
+    /// let pt = ProjectivePoint::hash_from_bytes::<hash2field::ExpandMsgXof<sha3::Shake256>>(b"test data", b"CURVE_XOF:SHAKE-256_SSWU_RO_");
+    ///
+    fn hash_from_bytes<X: ExpandMsg, M: AsRef<[u8]>>(msg: M, dst: &'static [u8]) -> Self::Output {
```
daxpedda, line 53:
```suggestion
    fn hash_from_bytes<X: ExpandMsg, M: AsRef<[u8]>>(msg: M, dst: &[u8]) -> Self::Output {
```
Is there a reason this is `'static`?

daxpedda (changes requested):
I think this is a bit too narrowly targeted towards P-256 & co., I believe a `CurveToMap` trait to fill the gap would alleviate that problem.

elliptic-curve/src/group_digest.rs
mikelodder7: It's clear to me that this naming doesn't convey what I want it to and that is we need a trait for converting two field elements into a point. MapToCurve is a separate method which I will add as I think that's a great suggestion. However, this trait is necessary since we don't have a method anywhere like From or anything else that allows passing in two field elements for x and y and returns the point. This is available directly with specific curves like `p256` and `k256` where one can use AffinePoint { x, y, infinity: false } but outside of that its not currently possible.

elliptic-curve/src/group_digest.rs
mikelodder7: Yep. Wasn't aware of that method. Thanks for pointing it out.

elliptic-curve/src/group_digest.rs
mikelodder7: Yep will do.

elliptic-curve/src/hash2field.rs
mikelodder7: Can you give an example of your preference for `hash_to_scalar`?

elliptic-curve/src/group_digest.rs
mikelodder7: Yes. HashToCurve requires handling a DST longer than 255 chars. See `Domain`. Having implemented this multiple times across various languages and libraries, 99% (I've never seen a dynamic one yet but there must be one somewhere) of the time the domain is static or known and not composed at run time. To be `no-std` compliant I can't use `Vec`. That leaves a couple of options: use `[u8; 255]` or `GenericArray<u8, U255>`, use `&'a [u8]` or `&'static [u8]`. The first option allocates more than enough space to handle any DST < 255 but overly so since they are usually < 32. This also requires copying the DST into the other array which is not really necessary. That leaves the other two. The first `&'a [u8]` requires adding the lifetime to all traits up the chain and more complex work to make Rust happy. This is undesirable and complicated. Thus the use of `&'static [u8]` which is not that strict of a requirement and keeps the API clean.

elliptic-curve/src/hash2field.rs
daxpedda: I believe this is already it. `hash_to_scalar` would just be used to build a `Scalar` directly from `out`.
See https://www.ietf.org/archive/id/draft-irtf-cfrg-hash-to-curve-13.html#section-5-4:
> The hash_to_field function is also suitable for securely hashing to scalars. For example, when hashing to the scalar field for an elliptic curve (sub)group with prime order r, it suffices to instantiate hash_to_field with target field GF(r).

I'm unsure what "instantiate hash_to_field with target field GF(r)" means exactly. Please correct me if I'm just talking nonsense here.

daxpedda: Actually thinking about this, my initial goal was to provide `HashToScalar` from the VOPRF spec: https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-voprf-08#section-4.1. But reading now through this again, VOPRF specifies specific routines for every group that aren't really part of `hash_to_scalar` from the hash2curve spec.

So I don't need this anymore. But maybe it's still useful for other people?

elliptic-curve/src/group_digest.rs
daxpedda: I actually thought that `&'a [u8]` was a perfectly fine solution here. This can be improved in a follow-up PR, not really a big issue.

elliptic-curve/src/hash2field.rs
tarcieri: I believe hash-to-scalar should already be covered by the `Reduce` trait: https://docs.rs/elliptic-curve/latest/elliptic_curve/ops/index.html

See `k256` as an example (for some reason this renders badly in rustdoc): https://docs.rs/k256/latest/src/k256/arithmetic/scalar.rs.html#586-590

You can also use `Reduce` with `NonZeroScalar` if you would like the result to be guaranteed to be non-zero: this performs the same wide reduction on a `U512`, but reduced modulo MODULUS - 1, and then adds one.

Unfortunately this is only implemented for `k256` at present.

daxpedda: I actually just discovered the `NonZeroScalar` part. That's amazing, and it's infallible too!

tarcieri: Here's a PR to add `digest` support to the `Reduce` trait: https://github.com/RustCrypto/traits/pull/869/files

elliptic-curve/src/hash2field/expand_msg_xmd.rs
```diff
@@ -1,72 +1,97 @@
...
     HashT: Digest + BlockInput,
 {
-    fn expand_message(msg: &[u8], dst: &[u8]) -> [u8; LEN_IN_BYTES] {
+    fn expand_message(msg: &[u8], dst: &'static [u8], len_in_bytes: usize) -> Self {
         let b_in_bytes = HashT::OutputSize::to_usize();
-        let ell = (LEN_IN_BYTES + b_in_bytes - 1) / b_in_bytes;
+        let ell = (len_in_bytes + b_in_bytes - 1) / b_in_bytes;
         if ell > 255 {
             panic!("ell was too big in expand_message_xmd");
         }
```
daxpedda, line 60: I played around with how to circumvent this locally and I achieved some pretty ugly code to check this on a type level.

So I believe the goal is to keep these functions infallible, certainly not panicking. I don't know how else to do this without doing some ugly `where` bounds on the length. To do this we will need to keep the previous implementation you had of storing the `LEN_IN_BYTES` as a generic (probably using `ArrayLength` instead of const `usize`).

That way we can put all these kind of checks on the `ExpandMsgXmd` type itself and keep functions infallible.
Another check that's missing here:
https://www.ietf.org/archive/id/draft-irtf-cfrg-hash-to-curve-13.html#section-5.4.1-4
> [..] For correctness, H requires b <= s.

This could be easily checked on the type level: `HashT::OutputSize: IsLessOrEqual<HashT::BlockSize>`.

The constraint pointed above that is currently handled by panicking:
https://www.ietf.org/archive/id/draft-irtf-cfrg-hash-to-curve-13.html#section-5.4.1-6
> - len_in_bytes, the length of the requested output in bytes,
  not greater than the lesser of (255 * b_in_bytes) or 2^16-1.

Would be a bit harder, but not impossible.

Alternatively of course, we can just return a `Result`, but panicking is not a long-term solution, or am I missing something? Maybe I'm overcautious or not understanding something properly.

WDYT? I'm willing to help, either by making follow-up PR's or PR against your fork @mikelodder7.

elliptic-curve/src/hash2field/expand_msg_xmd.rs
```diff
@@ -1,72 +1,97 @@
...
+where
+    HashT: Digest + BlockInput,
+{
+    b_0: GenericArray<u8, HashT::OutputSize>,
+    b_vals: GenericArray<u8, HashT::OutputSize>,
+    domain: Domain<HashT::OutputSize>,
+    index: usize,
+    offset: usize,
+    ell: usize,
     phantom: PhantomData<HashT>,
```
daxpedda, line 19: Now that `ExpandMsgXmd` holds data now, `PhantomData` isn't necessary anymore.

elliptic-curve/src/hash2field/expand_msg_xof.rs
```diff
@@ -1,27 +1,38 @@
...
 
 /// Placeholder type for implementing expand_message_xof based on an extendable output function
-#[derive(Debug)]
-pub struct ExpandMsgXof<HashT> {
+pub struct ExpandMsgXof<HashT>
+where
+    HashT: Default + ExtendableOutput + ExtendableOutputDirty + Update,
+{
+    reader: <HashT as ExtendableOutput>::Reader,
     phantom: PhantomData<HashT>,
```
daxpedda, line 13: As above.

daxpedda (changes requested):

elliptic-curve/src/group_digest.rs
```diff
@@ -0,0 +1,59 @@
...
+use crate::map2curve::MapToCurve;
+use group::cofactor::CofactorGroup;
+
+/// Adds hashing arbitrary byte sequences to a valid group element
+pub trait GroupDigest {
+    /// The field element representation for a group value with multiple elements
+    type FieldElement: FromOkm + Default + Copy;
+    /// The resulting group element
+    type Output: CofactorGroup<Subgroup = Self::Output>
+        + MapToCurve<FieldElement = Self::FieldElement, Output = Self::Output>;
```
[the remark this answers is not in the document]
mikelodder7: I like where you're going with this but I prefer the methods be directly accessible on the traits if possible rather than standalone functions.

elliptic-curve/src/group_digest.rs
```diff
@@ -0,0 +1,59 @@
...
+    ///
+    /// Examples
+    ///
+    /// Using a fixed size hash function
+    ///
+    /// let pt = ProjectivePoint::hash_from_bytes::<hash2field::ExpandMsgXmd<sha2::Sha256>>(b"test data", b"CURVE_XMD:SHA-256_SSWU_RO_");
+    ///
+    /// Using an extendable output function
+    ///
+    /// let pt = ProjectivePoint::hash_from_bytes::<hash2field::ExpandMsgXof<sha3::Shake256>>(b"test data", b"CURVE_XOF:SHAKE-256_SSWU_RO_");
```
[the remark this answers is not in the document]
tarcieri: You can use:

` ```ignore `

...to avoid doctesting it

elliptic-curve/src/group_digest.rs
```diff
@@ -0,0 +1,59 @@
...
+    fn hash_from_bytes<X: ExpandMsg>(msg: &[u8], dst: &'static [u8]) -> Self::Output {
+        let mut u = [Self::FieldElement::default(), Self::FieldElement::default()];
+        hash_to_field::<X, _>(msg, dst, &mut u);
+        let q0 = Self::Output::map_to_curve(u[0]);
+        let q1 = Self::Output::map_to_curve(u[1]);
+        // Ideally we could add and then clear cofactor once
+        // thus saving a call but the field elements may not
+        // add properly due to the underlying implementation
+        // which could result in an incorrect subgroup
+        q0.clear_cofactor() + q1.clear_cofactor()
```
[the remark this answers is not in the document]
mikelodder7: Mostly because the curve coefficients are different. `FieldElement` expects the `a` and `b` to be the original values and the isogenies are different with curves like `k256` and `bls12-381`. Look at [PR](https://github.com/algorand/pairing-plus/pull/18) for example. This problem doesn't manifest for curves with no isogeny like `p256`. For `k256` clear_cofactor doesn't do anything anyway so it will be a no-op.

elliptic-curve/src/hash2field/expand_msg_xmd.rs
mikelodder7: A follow PR would be great if we can somehow enforce it with type bounds. I'm not sure how to do this cleanly though. We could just change it to be a `Result` and declare it good enough. The odds of this happening are pretty low when using any hash function from RustCrypto. That won't stop someone from shooting themselves if they have a custom function that has a ridiculous OutputSize

elliptic-curve/src/hash2field/expand_msg_xmd.rs
mikelodder7: Good call. I'll update.

elliptic-curve/src/group_digest.rs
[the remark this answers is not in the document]
mikelodder7: Cool didn't know that. Fixing.

elliptic-curve/src/hash2field/expand_msg_xmd.rs
daxpedda: Maybe an easy solution would be just limit the `Outputsize` to something sane and be done with it. But yes, let's clear that in a follow-up.

elliptic-curve/src/group_digest.rs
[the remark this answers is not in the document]
daxpedda: Thanks! Could you add that to the comment too?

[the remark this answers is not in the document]
mikelodder7: Done.

mikelodder7:
I propose we merge this PR but have a follow-up to address before publishing the update to the crate:

Limit `HashT::OutputSize` so it can't possibly result in an `ell` larger than 255. This could possibly be done with Traits somehow. @daxpedda I'm curious about your approach to this.

We could also try to do the blanket impl. This could be done by breaking `GroupDigest` into two traits where the first requires to implement the required `const` values and the 2nd is the blanket impl.

daxpedda:
I agree. It would be great to keep https://github.com/RustCrypto/elliptic-curves/pull/495 up-to-date to see how it actually plays out. Preferably also run it over the test-vectors.

I will find some time to make a follow-up PR with some experiments on how to do the constraint for `ell`.

elliptic-curve/Cargo.toml
```diff
@@ -47,10 +47,10 @@ alloc = ["der/alloc", "sec1/alloc", "zeroize/alloc"] # todo: use weak activation
 arithmetic = ["ff", "group"]
 bits = ["arithmetic", "ff/bits"]
 dev = ["arithmetic", "hex-literal", "pem", "pkcs8"]
+digest = ["digest_traits", "ff", "group"]
```
tarcieri, line 50: Perhaps naming this `hash2curve` would be clearer?

```suggestion
hash2curve = ["arithmetic", "digest"]
```

mikelodder7: What happened to gating behind `digest`?!?

tarcieri:
Perhaps it would make sense to put all of this functionality in a `hash2curve` module to keep it organized and avoid polluting the toplevel crate namespace.

It would also make gating for conditional compilation easier.

It seems like modules like `osswu` could be placed in there as well?

elliptic-curve/Cargo.toml
daxpedda: See https://github.com/RustCrypto/traits/pull/855#discussion_r777522971 and https://github.com/RustCrypto/traits/pull/863.

tarcieri: > What happened to gating behind digest?!?

The changes in this PR:

The `hashing` feature used to have no other effect on other crate dependencies and was therefore redundant.

Now you do want to activate additional features, and to do that you're renaming the `digest` feature.

Since you now need that level of indirection, I'd suggest keeping `digest` as the crate dependency, and adding a separately named (and more specifically named) feature to handle that indirection, rather than renaming the package.

`digest` can now also be independently used for hash-to-scalar without activating `hash2curve`.

mikelodder7:
@tarcieri Moved all traits to `hash2curve` module. `digest` has been restored and now these traits are gated under the `hash2curve` feature now.

elliptic-curve/Cargo.toml
```diff
@@ -29,7 +29,7 @@ zeroize = { version = "1", default-features = false }
 
 # optional dependencies
 base64ct = { version = "1", optional = true, default-features = false }
-digest = { version = "0.9", optional = true, default-features = false }
+digest = { version = "0.9", optional = true, default-features = false, package = "digest" }
```
tarcieri, line 32:
```suggestion
digest = { version = "0.9", optional = true, default-features = false }
```

elliptic-curve/src/hash2curve.rs
```diff
@@ -0,0 +1,14 @@
+/// Traits for handling hash to curve
+mod group_digest;
+/// Traits for mapping an isogeny to another curve
+/// <https://datatracker.ietf.org/doc/draft-irtf-cfrg-hash-to-curve>
+mod isogeny;
+/// Traits for mapping field elements to points on the curve
+mod map2curve;
+/// Optimized simplified Shallue-van de Woestijne-Ulas methods
+mod osswu;
```
tarcieri, line 9: Much of the rest of this crate uses toplevel comments in the respective modules for these, rather than commenting them here

elliptic-curve/src/lib.rs
```diff
@@ -94,17 +94,18 @@ pub mod ecdh;
...
-#[cfg_attr(docsrs, doc(cfg(feature = "digest")))]
+#[cfg(feature = "hash2curve")]
+#[cfg_attr(docsrs, doc(cfg(feature = "hash2curve")))]
 pub mod hash2field;
 
+/// Traits for computing hash to curve as described in
+/// <https://datatracker.ietf.org/doc/draft-irtf-cfrg-hash-to-curve>
+#[cfg(feature = "hash2curve")]
+#[cfg_attr(docsrs, doc(cfg(feature = "hash2curve")))]
+pub mod hash2curve;
```
tarcieri, line 107: Here's another place it'd be nice to use toplevel module comments rather than comments on the respective modules.

Additionally I'd suggest splitting the links onto separate lines and keeping the first line short and sweet so it renders well in the rustdoc.

Here's what it looks like at present:

Screen Shot 2022-01-06 at 2 37 59 PM

Keeping the descriptions short will prevent the `hash2curve` feature label from linewrapping.

tarcieri:
> I propose we merge this PR but have a follow-up to address before publishing the update to the crate:

Sounds good to me!

A few nits but otherwise this seems like a reasonable start.

I can hold off on any releases until you've had time to play around with the RustCrypto/elliptic-curves PR and are happy with the API.

mikelodder7:
Done w/nits @tarcieri

tarcieri:
@mikelodder7 had one more suggestion for avoiding overflow here: https://github.com/RustCrypto/traits/pull/865#discussion_r779242589

elliptic-curve/src/hash2field/expand_msg_xmd.rs
```diff
@@ -1,72 +1,102 @@
...
+        // }
+        // However the where bound prevents this from being possible
+        let domain = Domain::xmd::<HashT>(dst);
         let b_0 = HashT::new()
             .chain(GenericArray::<u8, HashT::BlockSize>::default())
             .chain(msg)
-            .chain([(LEN_IN_BYTES >> 8) as u8, LEN_IN_BYTES as u8, 0u8])
-            .chain(dst)
-            .chain([dst.len() as u8])
+            .chain([(len_in_bytes >> 8) as u8, len_in_bytes as u8, 0u8])
```
daxpedda, line 71: If we want to solve some overflows in this PR, this can overflow too.

tarcieri:
Going to go ahead and merge this and follow up with a small PR for some of the remaining nits

PR #871, opened 2022-01-07
elliptic-curve: make `ExpandMsg` fallible

tarcieri:
As noted on #865 there are some potential overflow problems constructing the messages used for expansion, particularly around input lengths.

This commit changes `ExpandMsg::expand_message`, and all of its transitive callers, to be fallible in order to handle these cases.

cc @mikelodder7 @daxpedda

### DIFF
--- a/elliptic-curve/src/hash2curve/group_digest.rs
+++ b/elliptic-curve/src/hash2curve/group_digest.rs
@@ -1,5 +1,8 @@
 use super::MapToCurve;
-use crate::hash2field::{hash_to_field, ExpandMsg, FromOkm};
+use crate::{
+    hash2field::{hash_to_field, ExpandMsg, FromOkm},
+    Result,
+};
 use group::cofactor::CofactorGroup;
 
 /// Adds hashing arbitrary byte sequences to a valid group element
@@ -34,9 +37,9 @@ pub trait GroupDigest {
     /// let pt = ProjectivePoint::hash_from_bytes::<hash2field::ExpandMsgXof<sha3::Shake256>>(b"test data", b"CURVE_XOF:SHAKE-256_SSWU_RO_");
     /// ```
     ///
-    fn hash_from_bytes<X: ExpandMsg>(msg: &[u8], dst: &'static [u8]) -> Self::Output {
+    fn hash_from_bytes<X: ExpandMsg>(msg: &[u8], dst: &'static [u8]) -> Result<Self::Output> {
         let mut u = [Self::FieldElement::default(), Self::FieldElement::default()];
-        hash_to_field::<X, _>(msg, dst, &mut u);
+        hash_to_field::<X, _>(msg, dst, &mut u)?;
         let q0 = Self::Output::map_to_curve(u[0]);
         let q1 = Self::Output::map_to_curve(u[1]);
         // Ideally we could add and then clear cofactor once
@@ -49,7 +52,7 @@ pub trait GroupDigest {
         // isogenies are different with curves like k256 and bls12-381.
         // This problem doesn't manifest for curves with no isogeny like p256.
         // For k256 and p256 clear_cofactor doesn't do anything anyway so it will be a no-op.
-        q0.clear_cofactor() + q1.clear_cofactor()
+        Ok(q0.clear_cofactor() + q1.clear_cofactor())
     }
 
     /// Computes the encode to curve routine according to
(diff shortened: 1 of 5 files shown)


daxpedda:
I'm currently working on making it safe on the type-level instead. In any case, we don't want to limit the message-length to 255 I believe. Looking into it.

mikelodder7 (commented):
Would it make sense to return an Option instead?
That’s how `invert` works in `ff:Field`

tarcieri:
> I'm currently working on making it safe on the type-level instead.

How? I'm unclear what part of this can be addressed with type-level safety.

> In any case, we don't want to limit the message-length to 255 I believe

It's limited to 65,535 bytes, as it's being encoded as a 16-bit integer (although I'm not sure what the additional trailing zero is for).

> Would it make sense to return an Option instead?
> That’s how invert works in ff:Field

It's a possible alternative, although in this particular case it really feels like an error.

As it were, `Invert` returns a `CtOption` (to handle inverting zero) and is easily amenable to constant-time implementations.

mikelodder7:
Hash2curve is also supposed to be constant time.

tarcieri:
The implementation is still constant time with respect to the set of valid input messages.

This is just adding error handling for invalid inputs.

daxpedda:
> > I'm currently working on making it safe on the type-level instead.
> 
> How? I'm unclear what part of this can be addressed with type-level safety.

I'm still looking into it, but for example the length in bytes should be addressable on a type level, because we only use a length of one or two internally.

The reason is to make all of this infallible, as it's totally possible.

mikelodder7:
It’s probably fine the way it is. I just noticed the similarity

mikelodder7 (approved):
I agree that I don’t know how to enforce at the type level. Believe me I lost many hours before submitting this since I couldn’t create one. This is good enough

daxpedda:
I'm already on it, if you give me just a bit I will submit another PR to do that :).

mikelodder7:
I’m curious since the calculation comes from multiplying the number of needed bytes per field element by the number of field elements to create

tarcieri:
It's possible to compute it with `typenum` but be aware that complex arithmetic expressions in `typenum` adds to compile time and if you're making the buffer into a `GenericArray` it's going to be a real pain for the callers and monomorphizes for every possible size

daxpedda:
I'm attempting to make is a light-weight as possible, looking forward to your feedback then :smiley:

daxpedda:
See https://github.com/RustCrypto/traits/pull/872.

mikelodder7 (commented):
Since we're fallible now, we should probably check that `len_in_bytes`≠0 since that would yield points at infinity.

tarcieri:
I can take a crack at reworking this with CtOption as well, at least for comparison

daxpedda:
I migrated fixes here: https://github.com/khonsulabs/traits/tree/expand-msg-improvements-2. Will notate them for your convenience here, but if you like to merge this as is and me making a PR, say the word.

elliptic-curve/src/hash2curve/group_digest.rs
```diff
@@ -34,9 +37,9 @@ pub trait GroupDigest {
     /// let pt = ProjectivePoint::hash_from_bytes::<hash2field::ExpandMsgXof<sha3::Shake256>>(b"test data", b"CURVE_XOF:SHAKE-256_SSWU_RO_");
     /// ```
     ///
-    fn hash_from_bytes<X: ExpandMsg>(msg: &[u8], dst: &'static [u8]) -> Self::Output {
+    fn hash_from_bytes<X: ExpandMsg>(msg: &[u8], dst: &'static [u8]) -> Result<Self::Output> {
         let mut u = [Self::FieldElement::default(), Self::FieldElement::default()];
-        hash_to_field::<X, _>(msg, dst, &mut u);
+        hash_to_field::<X, _>(msg, dst, &mut u)?;
```
daxpedda, line 42: Theoretically this can't fail if users are only using `ExpandMsg` specified in the spec and don't implement `FromOkm` on a curve with an incredibly high number, again, not specified in the spec.

WDYT? Should we unwrap? Honestly I'm fine either way.

tarcieri: How about I merge this PR and you can open a follow-up to play with that?

tarcieri:
Going to merge this and we can experiment with things like converting to `CtOption` or making some of the APIs infallible in follow-up PRs

PR #872, opened 2022-01-07
Experiment: enforce constraints of `ExpandMsg` on the type level

daxpedda:
Just a start, I'm gonna continue to pick them off one by one.

### DIFF
--- a/elliptic-curve/src/hash2field/expand_msg.rs
+++ b/elliptic-curve/src/hash2field/expand_msg.rs
@@ -1,5 +1,5 @@
 use digest::{Digest, ExtendableOutputDirty, Update, XofReader};
-use generic_array::typenum::{IsLessOrEqual, U256};
+use generic_array::typenum::{IsLess, U256};
 use generic_array::{ArrayLength, GenericArray};
 
 /// Salt when the DST is too long
@@ -25,7 +25,7 @@ pub trait ExpandMsg<L: ArrayLength<u8>> {
 /// [dst]: https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-13#section-5.4.3
 pub(crate) enum Domain<L>
 where
-    L: ArrayLength<u8> + IsLessOrEqual<U256>,
+    L: ArrayLength<u8> + IsLess<U256>,
 {
     /// > 255
     Hashed(GenericArray<u8, L>),
@@ -35,7 +35,7 @@ where
 
 impl<L> Domain<L>
 where
-    L: ArrayLength<u8> + IsLessOrEqual<U256>,
+    L: ArrayLength<u8> + IsLess<U256>,
 {
     pub fn xof<X>(dst: &'static [u8]) -> Self
     where
--- a/elliptic-curve/src/hash2field/expand_msg.rs
+++ b/elliptic-curve/src/hash2field/expand_msg.rs
@@ -1,4 +1,5 @@
 use digest::{Digest, ExtendableOutputDirty, Update, XofReader};
+use generic_array::typenum::{IsLessOrEqual, U256};
 use generic_array::{ArrayLength, GenericArray};
 
 /// Salt when the DST is too long
@@ -22,14 +23,20 @@ pub trait ExpandMsg<L: ArrayLength<u8>> {
 /// Implements [section 5.4.3 of `draft-irtf-cfrg-hash-to-curve-13`][dst].
 ///
 /// [dst]: https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-13#section-5.4.3
-pub(crate) enum Domain<L: ArrayLength<u8>> {
+pub(crate) enum Domain<L>
+where
+    L: ArrayLength<u8> + IsLessOrEqual<U256>,
+{
     /// > 255
     Hashed(GenericArray<u8, L>),
     /// <= 255
     Array(&'static [u8]),
 }
 
-impl<L: ArrayLength<u8>> Domain<L> {
+impl<L> Domain<L>
+where
+    L: ArrayLength<u8> + IsLessOrEqual<U256>,
+{
     pub fn xof<X>(dst: &'static [u8]) -> Self
     where
         X: Default + ExtendableOutputDirty + Update,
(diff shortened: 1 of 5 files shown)


elliptic-curve/src/hash2field/expand_msg_xmd.rs
```diff
@@ -53,23 +54,28 @@ where
...
     HashT::OutputSize: IsLessOrEqual<HashT::BlockSize>,
+    L: ArrayLength<u8> + IsLess<U65536>,
 {
-    fn expand_message(msg: &[u8], dst: &'static [u8], len_in_bytes: usize) -> Self {
+    fn expand_message(msg: &[u8], dst: &'static [u8]) -> Self {
         let b_in_bytes = HashT::OutputSize::to_usize();
-        let ell = (len_in_bytes + b_in_bytes - 1) / b_in_bytes;
+        let ell = (L::to_usize() + b_in_bytes - 1) / b_in_bytes;
         // if ell > 255 {
         //     panic!("ell was too big in expand_message_xmd");
```
tarcieri, line 68: What about this?

daxpedda: That's what I'm trying to figure out at the moment. It could actually be made very easy by just reducing the maximum `len_in_bytes` to 64.772, which would make it impossible to reach 255. In any case, still playing around with it.

daxpedda: Alright, I re-read the spec a lot and couldn't figure out a clean way, I will just make the calculation on the type and call it a day, coming in shortly.

daxpedda: This should be correct now.

elliptic-curve/src/hash2field/expand_msg_xmd.rs
```diff
@@ -1,17 +1,20 @@
...
     BlockInput, Digest,
 };
+use generic_array::{typenum::Prod, ArrayLength};
 
 /// Placeholder type for implementing expand_message_xmd based on a hash function
 pub struct ExpandMsgXmd<HashT>
 where
     HashT: Digest + BlockInput,
-    HashT::OutputSize: IsLess<U256>,
+    HashT::OutputSize: IsLessOrEqual<U256>,
```
daxpedda, line 17: This should have been `<=256`, not `<256` as far as I can tell from the spec?

mikelodder7 (commented):
Can you make the prod by a nonzero?
That would prevent the case of accidentally calling  hash to field for 0 bytes and getting points at infinity

mikelodder7:
This is cool By the way

daxpedda:
I believe I covered all possible overflows now. The only thing that's not according to spec is the limit on the output size of the hash. It should actually be `Or<<L as IsLess<U256>>, <HashT::OutputSize as IsLessOrEqual<U256>>>`, sadly there is no such thing as an `Or`, but I think that's alright for now.

daxpedda:
> Can you make the prod by a nonzero? That would prevent the case of accidentally calling hash to field for 0 bytes and getting points at infinity

I'm surprised I can't find anything in the spec, it says nowhere that it has to be a non-zero as far as I searched.

mikelodder7:
> I'm surprised I can't find anything in the spec, it says nowhere that it has to be a non-zero as far as I searched.
I know but it’s a case we should check anyway

daxpedda:
Maybe we should report it?

mikelodder7:
Agree. I’ll file a ticket at IETF

daxpedda:
I hope the comments I added aren't too much, otherwise this is good to go @tarcieri.

elliptic-curve/src/hash2field/expand_msg_xmd.rs
tarcieri: I originally suggested `<256` so the cast to `u8` wouldn't overflow.

I'm a bit confused what's even happening now as you appear to be passing `[usize]` to `chain`?!

elliptic-curve/src/hash2field/expand_msg.rs
```diff
@@ -65,10 +72,11 @@ impl<L: ArrayLength<u8>> Domain<L> {
...
     }
 
-    pub fn len(&self) -> usize {
+    pub fn len(&self) -> u8 {
         match self {
-            Self::Hashed(_) => L::to_usize(),
-            Self::Array(d) => d.len(),
+            Self::Hashed(_) => L::to_u8(),
+            // Can't overflow because enforced on a type level.
+            Self::Array(d) => d.len() as u8,
```
tarcieri, line 79: You changed the bound to `IsLessOrEqual<U256>`, but `256usize as u8` overflows to 0?

Are you preventing that overflow in some way I'm missing?

elliptic-curve/src/hash2field/expand_msg_xmd.rs
daxpedda: No, I messed up, it should be less then 256, I was a lunatic and thought `u8` can reach 256.

daxpedda: Fixed.

daxpedda:
I was hoping that in practice only certain sizes are used in very low ranges so that this wouldn't be an issue. I don't disagree though.

A proposal: we could provide an alternative fallible run-time checked version, but keep this one around to make sure the hash2curve implementation stays infallible. We could keep the infallible version of `ExpandMsg` internal too.

WDYT?

tarcieri:
You should definitely change the internal implementation to a fallible, type-erased one that operates on slices to avoid monomorphization code bloat. The type safe one could call `expect` (although of course the compiler won't be able to prove it can remove the panic)

Infallibility/fallibility creates a pretty big divide per #871. If you want to support public fallible, slice-based APIs in addition to infallible type-safe ones you will need to have this paired fallible/infallible approach for 6 different methods at least (12 methods total).

The monomorphization code bloat problem will impact all of these methods as well, so maybe at the very least you should have private, fallible slice-based methods the infallible type-safe ones call.

Having gotten burned by large `generic-array`s in the past I really do worry about the ergonomics here. These are not going to be fun APIs to use if you have messages >1024 bytes.

daxpedda:
Alright, let me see if I can remove these constraints without the other improvements I've done so far.

daxpedda:
I'm pretty sure I basically reverted everything I've done so far, sad that this can't be salvaged. Let's continue on #871.

tarcieri:
I think the core idea is still good, but we need `generic_const_exprs` to stabilize to address some of the issues I highlighted and make it more ergonomic

PR #874, opened 2022-01-07
`ExpandMsg` improvements

daxpedda:
Follow-up for #400.

- Prevent overflows and add zero check.


daxpedda:
So should we make `hash_from_bytes` and `encode_from_bytes` infallible?
https://github.com/RustCrypto/traits/pull/871#discussion_r780348603

daxpedda:
Some outstanding issues:
- Updating to digest 0.10.
- Make `hash_to_field` publicly accessible.

We can merge this and make separate PR's for each or handle it here. Are there any other unresolved questions @tarcieri @mikelodder7.

tarcieri:
> Updating to digest 0.10.

This is going to be a pretty big effort and needs to happen in a coordinated way across not only `elliptic-curve` but also `ecdsa` and `k256`/`p256`. It will require a major version bump of all of these crates and I'm still working on getting all of the downstream dependencies of these crates I maintain updated to the latest releases.

I did the first step of cutting a new `signature` release which bumps to `digest` v0.10 but updating everything else is going to be a lot of work.

daxpedda:
That's alright, we will do this later then. The remaining issue is `hash_to_field` then. Any suggestions on what to do? I was actually thinking of adding it to `GroupDigest` and letting it return a `Scalar`, which should solve the problem.

tarcieri:
Isn't `hash_to_field` primarily intended to hash an input message to elements of the base field?

Are there valid use cases (interop?) for hash-to-scalar that make it a better choice than the existing `Reduce` trait?

daxpedda:
Implementing VOPRF requires calling `hash_to_field`:
https://www.ietf.org/archive/id/draft-irtf-cfrg-voprf-08.html#section-4.3-1.1.2.2

tarcieri:
Okay, well if it's needed for interop I'd be fine with a public `hash_to_scalar` function, but I think it should be called that and not `hash_to_field`, to avoid confusion with the base field

tarcieri:
Also VOPRF seems like the sort of thing that probably belongs in the respective RustCrypto/elliptic-curve crates?

daxpedda:
> ... but I think it should be called that and not `hash_to_field`, to avoid confusion with the base field

Sounds good to me.
> Also VOPRF seems like the sort of thing that probably belongs in the respective RustCrypto/elliptic-curves crates?

Yes, we just need a trait we can be generic over.

mikelodder7:
`hash_to_scalar` would be just a convenience wrapper for Digest output to 64 bytes then calling Reduce. I wouldn't use`hash_to_field` for this purpose since it does a more complex mixing of bytes than is necessary for that purpose.

elliptic-curve/src/hash2field/expand_msg_xmd.rs
```diff
@@ -57,34 +57,39 @@ where
...
+    // If `len_in_bytes` is bigger then 256, length of the `DST` will depend on
+    // the output size of the hash, which is still not allowed to be bigger then 256:
+    // https://www.ietf.org/archive/id/draft-irtf-cfrg-hash-to-curve-13.html#section-5.4.1-6
+    HashT::OutputSize: IsLess<U256>,
+    // Constraint set by `expand_message_xmd`:
+    // https://www.ietf.org/archive/id/draft-irtf-cfrg-hash-to-curve-13.html#section-5.4.1-4
+    HashT::OutputSize: IsLessOrEqual<HashT::BlockSize>,
 {
     fn expand_message(msg: &[u8], dst: &'static [u8], len_in_bytes: usize) -> Result<Self> {
-        if len_in_bytes > 0xFFFF {
```
mikelodder7, line 63:
```suggestion
if len_in_bytes == 0 || len_in_bytes > 0xFFFF {
```

daxpedda:
Well I have to name it something :smile:. Preferable I would like to stick with `hash_to_field` as this is what it's called in the spec, but I understand @tarcieri concern. I will just name it `hash_to_scalar` for now and let you guys hash it out :wink:.

elliptic-curve/src/hash2field/expand_msg.rs
```diff
@@ -66,10 +73,18 @@ impl<L: ArrayLength<u8>> Domain<L> {
...
 
-    pub fn len(&self) -> usize {
+    pub fn len(&self) -> u8 {
         match self {
-            Self::Hashed(_) => L::to_usize(),
-            Self::Array(d) => d.len(),
+            // Can't overflow because it's enforced on a type level.
+            Self::Hashed(_) => L::to_u8(),
+            // Can't overflow because it's checked on creation.
+            Self::Array(d) => d.len() as u8,
```
tarcieri, line 81: Maybe it would be better to actually codify that comment?

```suggestion
            Self::Array(d) => u8::try_from(d.len()).expect("length overflow"),
```

mikelodder7:
> Well I have to name it something 😄. Preferable I would like to stick with `hash_to_field` as this is what it's called in the spec, but I understand @tarcieri concern. I will just name it `hash_to_scalar` for now and let you guys hash it out 😉.

Dalek uses the name [hash_from_bytes](https://docs.rs/curve25519-dalek/3.2.0/curve25519_dalek/scalar/struct.Scalar.html#method.hash_from_bytes). That could work.

tarcieri:
> hash_to_scalar would be just a convenience wrapper for Digest output to 64 bytes then calling Reduce. I wouldn't usehash_to_field for this purpose since it does a more complex mixing of bytes than is necessary for that purpose.

And that's what's already provided by these methods on `Reduce` (added #869):

Screen Shot 2022-01-07 at 7 53 16 AM

daxpedda:
> Maybe it would be better to actually codify that comment?

I left the comment to explain why this can't fail nonetheless.

> Dalek uses the name [hash_from_bytes](https://docs.rs/curve25519-dalek/3.2.0/curve25519_dalek/scalar/struct.Scalar.html#method.hash_from_bytes). That could work.

I agree with @tarcieri, it's not the same, and already provided.

mikelodder7:
Yep that's just fine the way it is

tarcieri (approved):
Looks good now!

@mikelodder7 WDYT?

daxpedda:
I'm a bit busy with something else right now, still adding the other test vectors.

tarcieri:
Not in a hurry. Feel free to push some more changes.

mikelodder7 (approved):

elliptic-curve/src/hash2curve/group_digest.rs
```diff
@@ -1,17 +1,20 @@
...
 /// Adds hashing arbitrary byte sequences to a valid group element
-pub trait GroupDigest {
+pub trait GroupDigest: ProjectiveArithmetic<ProjectivePoint = Self::Output> {
     /// The field element representation for a group value with multiple elements
-    type FieldElement: FromOkm + Default + Copy;
+    type FieldElement: FromOkm
+        + MapToCurve<Output = Self::Output>
+        + Default
+        + Copy
+        + Into<Self::Scalar>;
```
daxpedda, line 15: I added this to implement `hash_to_scalar`, unsure if this is really what we need here. My assumption is that `FieldElement` to `Scalar` conversion doesn't need modulo or anything special, is that wrong?

elliptic-curve/src/hash2curve/map2curve.rs
```diff
@@ -2,11 +2,9 @@
...
 pub trait MapToCurve {
-    /// The input values representing x and y
-    type FieldElement;
     /// The output point
     type Output;
 
     /// Map a field element into a point
-    fn map_to_curve(u: Self::FieldElement) -> Self::Output;
+    fn map_to_curve(u: Self) -> Self::Output;
 }
```
daxpedda, line 10: I changed this a bit, instead of being implemented on the projective point it's now implemented on the field element.

mikelodder7: Makes sense. I'm okay with this change.

mikelodder7: However, could we implement it as `fn map_to_curve(u: &self) -> Self::Output;`

mikelodder7: That's similar to how `osswu` works.

daxpedda: That makes a lot of sense indeed.

mikelodder7:
> Looks good now!
> 
> @mikelodder7 WDYT?

I think it works for now once all the checks are cleared. Some minor nits that I can make after this merge.

elliptic-curve/src/hash2curve/group_digest.rs
tarcieri: `FieldElement` is an element of the base field.

`Scalar` is an element of the scalar field.

They're two completely different fields with two different moduli.

Outside of an extremely limited set of legacy algorithms (i.e. ECDSA) you shouldn't be converting an element of one field into the other.

mikelodder7: Consumers shouldn't be operating on `FieldElement` directly as they do with `Scalar`. I'm not even sure why you would want to expose `FieldElement`. I used to need it to implement hash_to_curve but now that's taken care of.

tarcieri: I would definitely prefer `FieldElement` remain hidden from the public API, because it's extremely easy to misuse (especially `k256` with lazy normalization).

If things legitimately need direct access to `FieldElement`, I'd prefer they get added directly to `k256`/`p256`.

daxpedda: I don't want to expose it, I want to somehow use `hash_to_field` an return a Scalar.

> `FieldElement` is an element of the base field.
> 
> `Scalar` is an element of the scalar field.
> 
> They're two completely different fields with two different moduli.

Clearly I don't know what I'm doing, if we can't convert a `FieldElement` to a `Scalar`, what's the way to make `hash_to_scalar` work?

daxpedda: I reverted the whole thing for now.

daxpedda:
This is good to go.

It seems I am out of my depth when it comes to how to implement `hash_to_scalar`, any guidance is welcome, in the meantime I will do more research.

elliptic-curve/src/hash2curve/group_digest.rs
tarcieri: `hash_to_field` seems to be defined for any field, so you perform the same set of operations you would with `FieldElement`, but use operations on `Scalar` instead.

Sidebar: it's really annoying they defined a complicated algorithm for this rather than just use a wide modular reduction. I see Ristretto255 does the latter. Kind of confused why they didn't do the same for the NIST P-curves but oh well.

daxpedda: Yeah, I didn't realize that, so what I'm currently missing is how to do the `FromOkm`, the way that @mikelodder7 did it is something I don't understand, I simply did modulo reduction, which seemed to work. I will look into it further.

elliptic-curve/src/hash2curve/group_digest.rs
```diff
@@ -1,17 +1,16 @@
...
 use crate::{
     hash2field::{hash_to_field, ExpandMsg, FromOkm},
-    Result,
+    ProjectiveArithmetic, Result,
 };
 use group::cofactor::CofactorGroup;
 
 /// Adds hashing arbitrary byte sequences to a valid group element
-pub trait GroupDigest {
+pub trait GroupDigest: ProjectiveArithmetic<ProjectivePoint = Self::Output> {
```
tarcieri, line 9: This looks like `hash2curve` feature is now dependent on the `arithmetic` feature in Cargo.toml.

CI isn't catching it because we don't currently test the crate builds with `hash2curve` alone enabled. We need to add it to the list here:

https://github.com/RustCrypto/traits/blob/master/.github/workflows/elliptic-curve.yml#L45

And it would be good to add it here too:

https://github.com/RustCrypto/traits/blob/master/.github/workflows/elliptic-curve.yml#L53

tarcieri: I'd suggest updating the CI config to confirm this, and if it breaks the build, adjusting this line in Cargo.toml:

https://github.com/RustCrypto/traits/blob/master/elliptic-curve/Cargo.toml#L50

....and changing it to:

```toml
```

daxpedda: This constraint was only necessary to implement `hash_to_scalar`, I will re-add it later when I've figured it out.

tarcieri (approved):
Looks good!

PR #875, opened 2022-01-07
elliptic-curve: refactor `hash2curve::expand_msg*` module structure

tarcieri:
The `hash2curve` module has three submodules which start with the `expand_msg` prefix:

- `expand_msg`
- `expand_msg_xmd`
- `expand_msg_xof`

This commit refactors them into a structure I think its more idiomatic, placing `xmd` and `xof` into submodules:

- `expand_msg`
  - `xmd`
  - `xof`

Accessing them is effectively the same:

- `expand_msg_xmd` => `expand_msg::xmd`
- `expand_msg_xof` => `expand_msg::xof`

I'd cite this as precedent (which is similar, but not quite exactly the same thing):

https://github.com/rust-lang/rfcs/blob/master/text/0356-no-module-prefixes.md


daxpedda (approved):

mikelodder7:
Yeah looks good. We probably don’t need to make hash_to_field public. Just the expand stuff. Maybe that can be part of this refactoring

tarcieri:
Removed `pub` from `hash2field` in 696bc4f

daxpedda:
I'm using `hash_to_field` to test against test vectors. Considering it can't be misused because `FieldElement` isn't public, can it be just hidden or behind a test feature?
https://github.com/RustCrypto/elliptic-curves/blob/7d3063a4981b8594df1355c5890233167782d8fb/p256/src/arithmetic/hash2curve.rs#L202

mikelodder7 (commented):
Minor nit. Do we need the cfg_attr doc on line 99 now that it’s private

tarcieri:
@daxpedda those tests still pass. It's still accessible from within the crate itself.

mikelodder7:
> I'm using `hash_to_field` to test against test vectors. Considering it can't be misused because `FieldElement` isn't public, can it be just hidden or behind a test feature?
> https://github.com/RustCrypto/elliptic-curves/blob/7d3063a4981b8594df1355c5890233167782d8fb/p256/src/arithmetic/hash2curve.rs#L202

Honestly there’s no reason to expose it even for testing. We can do all testing in this crate

daxpedda:
> @daxpedda those tests still pass. It's still accessible from within the crate itself.

This isn't in this crate, it's in p256 and k256.

daxpedda:
> > I'm using `hash_to_field` to test against test vectors. Considering it can't be misused because `FieldElement` isn't public, can it be just hidden or behind a test feature?
> > https://github.com/RustCrypto/elliptic-curves/blob/7d3063a4981b8594df1355c5890233167782d8fb/p256/src/arithmetic/hash2curve.rs#L202
> 
> Honestly there’s no reason to expose it even for testing. We can do all testing in this crate

That's possible of course.

tarcieri:
I could make it `#[doc(hidden)]`. We do that for a few other power user features we don't want people to accidentally stumble upon

mikelodder7:
Hash_to_field is curve agnostic. All it takes is the message and domain and returns bytes

mikelodder7:
> I could make it `#[doc(hidden)]`. We do that for a few other power user features we don't want people to accidentally stumble upon

Cool. I’m fine with that

daxpedda:
Actually, we were discussing earlier how to expose `hash_to_scalar` as part of `GroupDigest`, maybe that's not necessary, and it can just be left here as a low level primitive, like [`diffie_hellman`](https://docs.rs/elliptic-curve/latest/elliptic_curve/ecdh/fn.diffie_hellman.html)?

mikelodder7:
I view that as a separate issue

mikelodder7:
That should probably be added to a ScalarOps trait

daxpedda:
Alright.

tarcieri:
Yeah hopefully `#[doc(hidden)]` works for now, and also avoid people getting confused about whether they should be using `hash2curve` or `hash2field`

mikelodder7 (approved):

daxpedda (approved):

PR #877, opened 2022-01-07
elliptic-curve: hash2curve rustdoc improvements

tarcieri:
- Moves doc comments on `mod <modname>` into toplevel module comments. These have the same effect, but also self-document the module you are looking at, and are more amenable to rich top-of-module documentation. It's also consistent with the rest of the existing rustdoc in the RustCrypto project.
- Breaks apart the rustdoc "headline" from the rest of the comments for better formatting on the HTML rendering.
- Uses backticks and [`TypeName`] syntax to linkify type names where appropriate.


mikelodder7 (approved):

mikelodder7:
LGTM

PR #878, opened 2022-01-09
Introduce `VoprfParameters` trait

daxpedda:
Fixes RustCrypto/elliptic-curves#500.
Implementation: RustCrypto/elliptic-curves#506.


daxpedda:
So the trait introduced here as to pull in the corresponding hash trait. I've no clue how to circumvent that when you need the ID but not the hash. Alternatively I could split them into two traits and activate them with separate crate features.

Does this make any sense or is it overkill?

tarcieri (approved):
This looks fine to me

PR #883, opened 2022-01-14
Update to digest 0.10

daxpedda:
As requested.

### DIFF
--- a/elliptic-curve/Cargo.toml
+++ b/elliptic-curve/Cargo.toml
@@ -30,7 +30,7 @@ zeroize = { version = "1.5", default-features = false }
 
 # optional dependencies
 base64ct = { version = "1", optional = true, default-features = false }
-digest = { version = "0.9", optional = true, default-features = false }
+digest = { version = "0.10", optional = true, default-features = false, features = ["core-api"] }
 ff = { version = "0.11", optional = true, default-features = false }
 group = { version = "0.11", optional = true, default-features = false }
 hex-literal = { version = "0.3", optional = true }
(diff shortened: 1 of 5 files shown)


elliptic-curve/Cargo.toml
```diff
@@ -29,7 +29,7 @@ zeroize = { version = "1", default-features = false }
 
 # optional dependencies
 base64ct = { version = "1", optional = true, default-features = false }
-digest = { version = "0.9", optional = true, default-features = false }
+digest = { version = "0.10", optional = true, default-features = false, features = ["core-api"] }
```
newpavlov, line 32: `core-api` is enabled by default and it's the only default feature, so you can simply use `digest = "0.10"`.

elliptic-curve/src/hash2field/expand_msg/xmd.rs
```diff
@@ -53,24 +54,24 @@ where
...
 
         let b_0 = b_0
-            .chain(len_in_bytes_u16.to_be_bytes())
-            .chain([0])
-            .chain(domain.data())
-            .chain([domain.len()])
+            .chain_update(len_in_bytes_u16.to_be_bytes())
+            .chain_update([0])
+            .chain_update(domain.data())
+            .chain_update([domain.len()])
```
newpavlov, line 67: I am not sure compiler is able to properly optimize such long chains. To be safe it could be worth to use `HashT::new` and  `update`.

daxpedda: Interesting, is it okay then if I just replace all uses of `chain_update`?

daxpedda: Went ahead and just removed all `chain_update`s.

newpavlov: Yes, `chain_update` is a convenience method defined in terms of `update`.

tarcieri (approved):
LGTM.

Will wait a few days to merge this in case there are any other non-breaking hash2curve-related changes that are needed, then we can merge this and start making breaking changes again.

Issue #162, opened 2020-06-02
Test minimal versions of dependencies


newpavlov:
Lock file can be generated with `cargo generate-lockfile -Z minimal-versions`.

PR #885, opened 2022-01-15
digest: bump `generic-array` dependency to v0.14.4

tesaguri:
`generic-array` v0.14.4 uses `typenum` v1.12.0, which makes `Unsigned`(which `ArrayLength` implies) imply `Default`, which, in turn, is required for `CoreWrapper` to to implement `Digest`.

This is needed to make upstream crates which uses `Digest` implementation of `CoreWrappper` (such as `hmac`) to successfully build with `-Z minimal-versions`.

### DIFF
--- a/digest/Cargo.toml
+++ b/digest/Cargo.toml
@@ -12,7 +12,7 @@ keywords = ["digest", "crypto", "hash"]
 categories = ["cryptography", "no-std"]
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


newpavlov:
Thank you! I think I'll release it a bit later together with changes from #849.

And we really need to solve #162 for all our repositories...

PR #905, opened 2022-01-19
Add missing jobs to excluded workspace crates

daxpedda:
See RustCrypto/signatures#440 and RustCrypto/signatures#441.

I could move Clippy and Rustftm to each workflow instead of a common workspace workflow?


tarcieri:
Ugh, this is quickly becoming a maintenance nightmare.

Hopefully we can reverse this relationship soon so the workspace is 2021 edition and the 2018 edition straggler crates have their own pre-2021 edition workspaces.

cc @newpavlov

PR #884, opened 2022-01-14
Implement `ZeroizeOnDrop` on appropriate items

daxpedda:
I'm assuming this is a breaking change.

PR #923, opened 2022-01-27
elliptic-curve v0.11.10

tarcieri:
### Changed
- Revert #884 to support a wider range of `zeroize` versions.

### Notes
We've received a number of complaints about `zeroize` compatibility (see RustCrypto/utils#723), which was exacerbated by #884 which bumped the `zeroize` version within a minor version series.

I had hoped the cargo resolver would've been able to select an older compatible version of zeroize in these cases, but I've gotten a lot of reports that isn't happening.

This release reverts #884 in hopes of supporting a wider range of `zeroize` versions for now. It's technically breaking in the event anyone is actually depending on `ZeroizeOnDrop` marker traits on these types, but it seems we've already broken things for current users and that's the much higher impact issue.

### DIFF
--- a/elliptic-curve/CHANGELOG.md
+++ b/elliptic-curve/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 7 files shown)


tarcieri:
cc @daxpedda

As it were, you originally warned about this being a breaking change on #884. Apparently I should've listened 😅

tarcieri:
Went ahead and released and tagged this branch.

It's unmergable, as `master` is already the v0.12 prerelease series. I'll open a separate PR to incorporate a changelog entry.

PR #937, opened 2022-02-14
Fix compilation

daxpedda:
Downstream users can't have concrete instantiations on these traits. For example [AEADs](https://github.com/RustCrypto/AEADs) don't compile when depending on v0.4.

See https://github.com/RustCrypto/block-ciphers/issues/302.

### DIFF
--- a/cipher/src/block.rs
+++ b/cipher/src/block.rs
@@ -527,11 +527,11 @@ impl<Alg: BlockDecrypt> BlockDecrypt for &Alg {
 }
 
 /// Closure used in methods which operate over separate blocks.
-struct BlockCtx<'inp, 'out, BS: ArrayLength<u8>> {
+struct BlockCtx<'inp, 'out, BS: 'static + ArrayLength<u8>> {
     block: InOut<'inp, 'out, Block<Self>>,
 }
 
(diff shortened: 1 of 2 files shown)


newpavlov:
Note that AEAD crates are not ported to `cipher v0.4` yet.

tarcieri:
@newpavlov that is what @daxpedda is trying to work on

newpavlov:
Have you tried to compile it after `cargo update`? IIRC `U*` types on older `typenum` do not have `'static` lifetime, i.e. our trait crates currently are not quite correct around minimum dependency versions. I was planning to fix it in the next `crypto-common`/`cipher` releases.

Also I was thinking about a more thorough refactoring of the `aead` crate. But I guess we could do a quick port to `cipher v0.4` first.

tarcieri:
@newpavlov yeah there are some changes I'd like to make to `aead` as well

daxpedda:
> Have you tried to compile it after `cargo update`? IIRC `U*` types on older `typenum` do not have `'static` lifetime, i.e. our trait crates currently are not quite correct around minimum dependency versions.

True, updating typenum removes the need for `'static` and everything compiles. Should I close this PR then and instead update the dependency requirement?

tarcieri:
@daxpedda bumping the `typenum` version sounds good

PR #941, opened 2022-02-16
cipher: rename BlockDecryptMut::decrypt_padded_vec to decrypt_padded_vec_mut

newpavlov:
It's a breaking change, but the method was introduced in today's release, so hopefully it will not affect anyone. I plan to yank `cipher v0.4.1` to prevent further confusion.

### DIFF
--- a/cipher/CHANGELOG.md
+++ b/cipher/CHANGELOG.md
@@ -5,7 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## 0.4.1 (2022-02-16)
+## 0.4.2 (2022-02-16)
+### Fixed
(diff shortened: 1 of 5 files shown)


jcaesar:
You've managed to affect [a crate](https://github.com/hjiayz/p12/blob/master/Cargo.toml#L27) and it's bitten [someone](https://stackoverflow.com/questions/71151465/crate-p12-rust-error-e0599-no-method-name-decrypt-padded-vecfound-for-str). I'm not complaining, I'm just amazed. 

I'll throw a PR to p12.

newpavlov:
Oh, wow... Well, this is unfortunate. @Keruspe has submitted #936 to be used in `p12`, so no wonders it was used right away after `cipher v0.4.1` was released. Unfortunately, it looks like he was offline when I [wrote](https://github.com/RustCrypto/traits/commit/b4162de124df07d34063ead9225a487a9abefce7#r66816384) about the fix release.

jcaesar:
Ah. And here I was wondering how that new function made it into other code so quickly. (Btw, the asker on Stackoverflow isn't exactly following best practice, and not using a lock file.)

PR #958, opened 2022-03-09
Reunify Cargo workspace

tarcieri:
Places all crates into a single Cargo workspace again.

Removing the toplevel Cargo.toml when running CI on legacy pre-2021 edition crates should suffice to let them retain their current MSRV.


kem/Cargo.toml
```diff
@@ -27,11 +23,12 @@ p256 = { version = "0.9", features = [ "ecdsa" ] }
...
-git = "https://github.com/rozbb/rust-hpke"
-rev = "3c795445f38317d6a0868259caf3e87916548356"
-default-features = false
-features = [ "x25519" ]
+# TODO(tarcieri): re-enable when `zeroize` compatibility can be resolved
+#[dev-dependencies.hpke]
+#git = "https://github.com/rozbb/rust-hpke"
+#rev = "3c795445f38317d6a0868259caf3e87916548356"
+#default-features = false
+#features = [ "x25519" ]
```
tarcieri, line 31: cc @rozbb 

This no longer works due to a conflict in the `zeroize` versions required by `cipher` (1.5) and `x25519-dalek` as pulled in by `hpke`:

```
error: failed to select a version for `zeroize`.
    ... required by package `x25519-dalek v1.2.0`
    ... which satisfies dependency `x25519-dalek = "^1.2"` of package `hpke v0.8.0 (https://github.com/rozbb/rust-hpke?rev=3c795445f38317d6a0868259caf3e87916548356#3c795445)`
    ... which satisfies git dependency `hpke` of package `kem v0.1.0 (src/RustCrypto/traits/kem)`
versions that meet the requirements `=1.3` are: 1.3.0

all possible versions conflict with previously selected packages.

  previously selected package `zeroize v1.5.3`
    ... which satisfies dependency `zeroize = "^1.5"` (locked to 1.5.3) of package `cipher v0.4.3 (/Users/bascule/src/RustCrypto/traits/cipher)`

failed to select a version for `zeroize` which could resolve this conflict
```

Hopefully the next releases of Dalek remove the pinning.

newpavlov (approved):
Do not forget to remove Cargo.lock from the re-unified crates.

tarcieri:
Aah yeah, good point

kem/Cargo.toml
rozbb: Ugh that's annoying. Thanks for taking care of it. I'll re-enable once x25519-dalek v2.0 drops

PR #965, opened 2022-03-14
universal-hash v0.5

newpavlov:
The new API uses approach of emulating rank-2 closures previously introduced in cipher v0.4.

Also moves `ParBlocksSizeUser` and `ParBlocks` from `cipher` to `crypto-common`.

### DIFF
--- a/universal-hash/src/lib.rs
+++ b/universal-hash/src/lib.rs
@@ -18,46 +18,91 @@
 //! [Universal Hash Functions]: https://en.wikipedia.org/wiki/Universal_hashing
 
 #![no_std]
-#![forbid(unsafe_code)]
 #![doc(
     html_logo_url = "https://raw.githubusercontent.com/RustCrypto/media/8f1a9894/logo.svg",
     html_favicon_url = "https://raw.githubusercontent.com/RustCrypto/media/8f1a9894/logo.svg",
-    html_root_url = "https://docs.rs/universal-hash/0.4.1"
+    html_root_url = "https://docs.rs/universal-hash/0.5.0"
 )]
+#![cfg_attr(docsrs, feature(doc_cfg))]
 #![warn(missing_docs, rust_2018_idioms)]
 
 #[cfg(feature = "std")]
 extern crate std;
 
-pub use generic_array::{self, typenum::consts};
+pub use crypto_common::{
+    self,
+    generic_array,
+    typenum::{self, consts},
+    KeyInit, Key, Block, ParBlocks,
+};
 
-use generic_array::typenum::Unsigned;
-use generic_array::{ArrayLength, GenericArray};
-use subtle::{Choice, ConstantTimeEq};
+use typenum::Unsigned;
+use generic_array::{GenericArray, ArrayLength};
+use crypto_common::{BlockSizeUser, ParBlocksSizeUser};
+use subtle::ConstantTimeEq;
+use core::slice;
 
-/// Keys to a [`UniversalHash`].
-pub type Key<U> = GenericArray<u8, <U as NewUniversalHash>::KeySize>;
+/// Trait implemented by UHF backends.
+pub trait UhfBackend: ParBlocksSizeUser {
+    /// Process single block.
+    fn proc_block(&mut self, block: &Block<Self>);
 
-/// Blocks are inputs to a [`UniversalHash`].
-pub type Block<U> = GenericArray<u8, <U as UniversalHash>::BlockSize>;
+    /// Process several blocks in parallel.
+    #[inline(always)]
+    fn proc_par_blocks(&mut self, blocks: &ParBlocks<Self>) {
+        for block in blocks {
+            self.proc_block(block);
+        }
+    }
+}
 
-/// Instantiate a [`UniversalHash`] algorithm.
-pub trait NewUniversalHash: Sized {
-    /// Size of the key for the universal hash function.
-    type KeySize: ArrayLength<u8>;
-
-    /// Instantiate a universal hash function with the given key.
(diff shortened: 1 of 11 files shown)


tarcieri (approved):

universal-hash/src/lib.rs
```diff
@@ -18,167 +18,153 @@
...
-pub type Block<U> = GenericArray<u8, <U as UniversalHash>::BlockSize>;
-
-/// Instantiate a [`UniversalHash`] algorithm.
-pub trait NewUniversalHash: Sized {
-    /// Size of the key for the universal hash function.
-    type KeySize: ArrayLength<u8>;
+use subtle::ConstantTimeEq;
+use typenum::Unsigned;
+
+/// Trait implemented by UHF backends.
```
str4d, line 44: Document how this trait is intended to be used.

str4d, line 44: Document how this trait is intended to be used.

universal-hash/src/lib.rs
```diff
@@ -18,167 +18,153 @@
...
+    fn call<B: UhfBackend<BlockSize = Self::BlockSize>>(self, backend: &mut B);
 }
 
 /// The [`UniversalHash`] trait defines a generic interface for universal hash
 /// functions.
-pub trait UniversalHash: Clone {
-    /// Size of the inputs to and outputs from the universal hash function
-    type BlockSize: ArrayLength<u8>;
+pub trait UniversalHash: BlockSizeUser + Sized {
+    /// Update hash function state using the provided rank-2 closure.
```
str4d, line 69: Document how this method is intended to be implemented (calling out to target-specific backends). If you don't want that documentation to live on the trait that end users will be using, then instead add a reference here to `UhfBackend` and document the implementor notes there.

str4d, line 69: Document how this method is intended to be implemented (calling out to target-specific backends). If you don't want that documentation to live on the trait that end users will be using, then instead add a reference here to `UhfBackend` and document the implementor notes there.

tarcieri:
@newpavlov can you add the docs @str4d requested and get this merged so we can make progress on https://github.com/RustCrypto/universal-hashes/pull/153 ?

newpavlov:
I plan to improve the docs, but they do not block the work on https://github.com/RustCrypto/universal-hashes/pull/153 and update of the AEAD crates. Before working on docs I plan to properly review the @str4d's PR first.

tarcieri:
Okay, well there's also https://github.com/RustCrypto/AEADs/pull/414 and https://github.com/RustCrypto/AEADs/pull/415, which means we have a lot of PRs that are interdependent on other PRs.

And I'd like to start work on updating `aes-gcm` and `aes-gcm-siv` but it doesn't make sense to me to start that until some of these interdependencies are unblocked.

universal-hash/src/lib.rs
str4d: This is also where the note about user trait usage should go, for e.g. AEADs (https://github.com/RustCrypto/AEADs/pull/414#issuecomment-1107614537).

universal-hash/src/lib.rs
```diff
@@ -18,167 +18,153 @@
...
-pub trait NewUniversalHash: Sized {
-    /// Size of the key for the universal hash function.
-    type KeySize: ArrayLength<u8>;
+use subtle::ConstantTimeEq;
+use typenum::Unsigned;
+
+/// Trait implemented by UHF backends.
+pub trait UhfBackend: ParBlocksSizeUser {
+    /// Process single block.
+    fn proc_block(&mut self, block: &Block<Self>);
```
str4d, line 47: Per https://github.com/RustCrypto/universal-hashes/pull/153#discussion_r857124619, we are currently lacking a way to inform the consumer of the `UhfBackend` that the backend is "unaligned" and cannot consume `ParBlocks` efficiently (with impls either panicking, or always falling back to a simple loop over `UhfBackend::proc_block` without any parallelism). One way to avoid this would be for the backend to expose the misalignment:
```suggestion
    fn proc_block(&mut self, block: &Block<Self>);

    /// Returns the number of blocks that should be passed to `Self::proc_block` before
    /// `Self::proc_par_blocks` can be used efficiently. This is always less than
    /// `Self::ParBlocksSize`.
    fn blocks_needed_to_align(&self) -> usize;
```

tarcieri:
In the interest of moving things along and getting additional `-pre` the AEAD crates out (it's now been 5 months since `cipher` v0.4.0 was released, FWIW), I'd suggest merging this and releasing `universal-hash` v0.5.0-pre, and then merging the respective PRs on https://github.com/rustcrypto/universal-hashes and releasing prereleases of those crates.

We can open a general tracking issue for crates which are in need of more documentation.

str4d:
Looks like this PR needs a rebase to address merge conflicts? Also I'd like feedback on my suggestion above (as currently I don't think the API changes here can ever help ChaCha20Poly1305 performance due to the immediate 1-block misalignment when initializing the AEAD).

tarcieri:
@str4d it seems like adding a "flush" API which consumes the internal buffer and returns it to an aligned state could be added to address that problem?

Regardless releases of the AEAD crates seem stalled on this PR, and the only way we could experiment with the necessary API changes without merging this is as a series of PRs to AEADs which point to PRs to this branch of `universal-hash` which seems somewhat untenable.

Getting this merged (and ideally an associated prerelease cut) means we can more easily make incremental progress towards a final release.

tarcieri:
Rebased/squashed this PR and opened a new one: #1051

tarcieri:
Merged in #1051

PR #969, opened 2022-03-25
Fix compilation with `serde` without `pem`

daxpedda:
I'm surprised this is implemented as it is, you can't actually use Serde with `PublicKey` without `alloc`. In any case, this is still necessary to make this even compile.

### DIFF
--- a/.github/workflows/elliptic-curve.yml
+++ b/.github/workflows/elliptic-curve.yml
@@ -50,6 +50,7 @@ jobs:
       - run: cargo build --target ${{ matrix.target }} --release --no-default-features --features pkcs8
       - run: cargo build --target ${{ matrix.target }} --release --no-default-features --features sec1
       - run: cargo build --target ${{ matrix.target }} --release --no-default-features --features serde
+      - run: cargo build --target ${{ matrix.target }} --release --no-default-features --features alloc,serde
       - run: cargo build --target ${{ matrix.target }} --release --no-default-features --features voprf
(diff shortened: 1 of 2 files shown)


.github/workflows/elliptic-curve.yml
```diff
@@ -50,6 +50,7 @@ jobs:
       - run: cargo build --target ${{ matrix.target }} --release --no-default-features --features pkcs8
       - run: cargo build --target ${{ matrix.target }} --release --no-default-features --features sec1
       - run: cargo build --target ${{ matrix.target }} --release --no-default-features --features serde
+      - run: cargo build --target ${{ matrix.target }} --release --no-default-features --features alloc,serde
```
daxpedda, line 53: I'm sure that there are other combinations we should test ... is this alright or should I just remove this?

tarcieri: We should probably switch to `cargo hack` to test all combinations. This is fine for now, though.

newpavlov: Do not forget about combinatorial explosion. I think it's worth to limit execution time of CI jobs by several minutes.

tarcieri: The surface of each feature is pretty small. The features mostly exist for gating dependencies, and compilation results for those should be reused across runs

tarcieri:
@daxpedda I think GitHub Actions is having issues. Can you `--amend` and force push to try to toggle a CI run?

PR #982, opened 2022-04-18
KEM v2

rozbb:
I made a few changes to the KEM API

## Deserialization
As discussed over Zulip earlier this year, there should probably be a way to deserialize an encapsulated key sent over the wire.

This breaks a bit with the rest of RustCrypto because it uses `&GenericArray` rather than `&[u8]` for bitstrings. I'm not married to this, but the pros are:
1. It fits the existing API, which already uses a `GenericArray` to describe shared secrets
2. Fixed size arrays make function preconditions clear a first glance, which is good.

## Shared Secrets
Separately, I've also switched the API off of directly returning `GenericArray`s for a shared secret and made it its own `SharedSecret` type. The con is that the values are just a little more opaque. The pros are:
1. Shared secrets are now zeroize-on-drop
2. The type signatures are more concise

### DIFF
--- a/kem/src/kem.rs
+++ b/kem/src/kem.rs
@@ -11,10 +11,10 @@ use rand_core::{CryptoRng, RngCore};
 /// essence, a bag of bytes.
 pub trait EncappedKey: AsRef<[u8]> + Debug + Sized {
     /// The size, in bytes, of an encapsulated key.
-    type NEnc: ArrayLength<u8>;
+    type EncappedKeySize: ArrayLength<u8>;
 
     /// The size, in bytes, of the shared secret that this KEM produces.
-    type NSecret: ArrayLength<u8>;
+    type SharedSecretSize: ArrayLength<u8>;
 
     /// Represents the identity key of an encapsulator. This is used in authenticated
     /// decapsulation.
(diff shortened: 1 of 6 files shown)


rozbb:
Bump on this. Any thoughts? Just resolved merge conflicts.

tarcieri:
Will try to take a look later today

kem/tests/hpke.rs
```diff
@@ -1,12 +1,14 @@
 // TODO(tarcieri): re-enable when `zeroize` dependency issues are resolved
 #![cfg(disabled)]
```
tarcieri, line 2: Perhaps this can be removed now?

tarcieri (approved):

rozbb:
Awesome I’ll make the requested change soon. IIRC I have some nice updates to pqcrypto_compat that should make the KEMs adhere to this API.

kem/tests/hpke.rs
rozbb: Done. Everything should be good to go!

kem/src/kem.rs
```diff
@@ -1,48 +1,81 @@
...
+        // the underlying bytestring is not precisely NEnc bytes long.
+        self.as_ref().into()
+    }
+}
+
+/// The shared secret that results from key exchange.
+#[derive(Zeroize)]
+pub struct SharedSecret<EK: EncappedKey>(GenericArray<u8, EK::SharedSecretSize>);
+
+impl<EK: EncappedKey> ZeroizeOnDrop for SharedSecret<EK> {}
```
tarcieri, line 42: I think you likely want this:

```suggestion
#[derive(ZeroizeOnDrop)]
pub struct SharedSecret<EK: EncappedKey>(GenericArray<u8, EK::SharedSecretSize>);
```

What you have doesn't add a `Drop` handler, it just impls the `Zeroize` trait. But it's also a case where the *only* thing that should call `zeroize` is the `Drop` impl (i.e. you don't want to zeroize a shared secret you intend to still use), which is what the `ZeroizeOnDrop` custom derive is useful for.

Also, this is a simple enough case you can write it by hand without using the custom derive support:

```rust
impl<EK: EncappedKey> Drop for SharedSecret<EK> {
    fn drop(&mut self) {
        self.0.zeroize();
    }
}

impl<EK: EncappedKey> ZeroizeOnDrop for SharedSecret<EK> {}
```

rozbb: Good catch! Just pushed the fix.

tarcieri:
@rozbb would you like for me to cut a `kem` v0.2 release with these changes?

rozbb:
Thanks!

Issue #986, opened 2022-05-02
Make `Key` more generic rather than typealias?


sonder-joker:
I'm current work on support wasi-crypto for RustCrypto https://github.com/WebAssembly/wasi-crypto/issues/52.
In wasi-crypto, `Key` is represent as handle, which mean it not pass by generic-array, and this mean it design for reused.
If I just use generic-array, I have to `import` and `close` key again and again, which mean extra cost.
I thispass a trait will be better. For example
```rust
fn new(key: &Key<Self>) -> Self
where
    Self: KeyInit;
```
a fake implementatiom
```rust
fn new(key: & impl Key) -> Self
where
    Self: KeyInit;
```
Please let me know if I make a stupid mistake.

PR #1011, opened 2022-05-11
Abstract key.

sonder-joker:
https://github.com/RustCrypto/traits/issues/986 
wasi-crypto design base on handle, which mean for key
```rust 
struct Key {
  handle: u32
}
```
rather than
```rust
struct Key {
  data: [u8:SIZE]
}
```
Which mean cannot implement `AsRef<[u8]>` like trait.
This is just a simple attempt,  it shouldn't be merge now. If you think this will work, I will continue to work on it to make it more suitable.

### DIFF
--- a/digest/src/mac.rs
+++ b/digest/src/mac.rs
@@ -17,22 +17,8 @@ pub trait MacMarker {}
 /// traits and provides additional convenience methods.
 #[cfg_attr(docsrs, doc(cfg(feature = "mac")))]
 pub trait Mac: OutputSizeUser + Sized {
-    /// Create new value from fixed size key.
-    fn new(key: &Key<Self>) -> Self
-    where
-        Self: KeyInit;
-
-    /// Generate random key using the provided [`CryptoRng`].
-    #[cfg(feature = "rand_core")]
-    #[cfg_attr(docsrs, doc(cfg(feature = "rand_core")))]
-    fn generate_key(rng: impl CryptoRng + RngCore) -> Key<Self>
-    where
-        Self: KeyInit;
-
-    /// Create new value from variable size key.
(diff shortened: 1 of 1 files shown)


newpavlov:
Your approach looks wrong to me. With it you make life harder for everyone outside of wasi-crypto users. Also it would require a larger overhaul of APIs, which should involve the other crates as well (such as `cipher`). I am not saying that we never will move from plain byte array keys (e.g. it could be useful for automatic zeroization of keys and support of external hardware), but it has to be well designed and it should include demonstration of how it will impact the downstream crates, i.e. you would need to migrate the crates in hashes and MACs repos to the new design.

IIUC in your case the overhead only applies when the same key gets reused for algorithm initialization many times. If initialized algorithm gets reused (e.g. by using the `reset` methods), there will be no overhead. It's a bit of pitfall for users, but in general its a good practice to reuse already initialized structures if it's possible.

sonder-joker:
> Your approach looks wrong to me. With it you make life harder for everyone outside of wasi-crypto users. Also it would require a larger overhaul of APIs, which should involve the other crates as well (such as `cipher`). I am not saying that we never will move from plain byte array keys (e.g. it could be useful for automatic zeroization of keys and support of external hardware), but it has to be well designed and it should include demonstration of how it will impact the downstream crates, i.e. you would need to migrate the crates in hashes and MACs repos to the new design.
> 
> IIUC in your case the overhead only applies when the same key gets reused for algorithm initialization many times. If initialized algorithm gets reused (e.g. by using the `reset` methods), there will be no overhead. It's a bit of pitfall for users, but in general its a good practice to reuse already initialized structures if it's possible.

Sorry, I didn't think about it that much. Thank you for your reply.

PR #1022, opened 2022-06-12
elliptic-curve: generic impl of complete prime order formulas

tarcieri:
Adds a generic implementation of the complete addition formulas from [Renes-Costello-Batina 2015] adapted from @str4d's original implementation for the `p256` crate:

https://github.com/RustCrypto/elliptic-curves/pull/15

This implementation has subsequently been copied-and-pasted into the `p384` crate (see https://github.com/RustCrypto/elliptic-curves/pull/565), hence the motivation to make it generic and extract it somewhere that it can be reused.

The API exposed is fairly low-level, however it's difficult to better encapsulate it without making breaking changes to the `elliptic-curve` crate. Thus this PR opts to provide an initial low-level generic implementation with the goal of exploring removing more duplication with a higher-level API as followup work to be done at a time when breaking changes are permitted.

[Renes-Costello-Batina 2015]: https://eprint.iacr.org/2015/1060

### DIFF
--- a/elliptic-curve/src/weierstrass.rs
+++ b/elliptic-curve/src/weierstrass.rs
@@ -1,0 +1,128 @@
+//! Complete projective formulas for prime order elliptic curves as described
+//! in [Renes-Costello-Batina 2015].
+//!
+//! [Renes-Costello-Batina 2015]: https://eprint.iacr.org/2015/1060
+
+#![allow(clippy::op_ref)]
+
+use ff::Field;
+
+/// Affine point whose coordinates are represented by the given field element.
+pub type AffinePoint<Fe> = (Fe, Fe);
+
+/// Projective point whose coordinates are represented by the given field element.
+pub type ProjectivePoint<Fe> = (Fe, Fe, Fe);
+
+/// Implements the complete addition formula from [Renes-Costello-Batina 2015]
+/// (Algorithm 4).
+///
+/// [Renes-Costello-Batina 2015]: https://eprint.iacr.org/2015/1060
+#[inline(always)]
+pub fn add<Fe>(
+    (ax, ay, az): ProjectivePoint<Fe>,
+    (bx, by, bz): ProjectivePoint<Fe>,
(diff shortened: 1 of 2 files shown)


elliptic-curve/src/weierstrass.rs
```diff
@@ -0,0 +1,128 @@
...
+
+#![allow(clippy::op_ref)]
+
+use ff::Field;
+
+/// Affine point whose coordinates are represented by the given field element.
+pub type AffinePoint<Fe> = (Fe, Fe);
+
+/// Projective point whose coordinates are represented by the given field element.
+pub type ProjectivePoint<Fe> = (Fe, Fe, Fe);
```
tarcieri, line 14: The use of coordinate tuples is to prevent confusion with the `AffinePoint` and `ProjectivePoint` types currently defined on a crate-by-crate basis (i.e. in `p256` and `p384`)

Perhaps it would be possible in an eventual followup PR to extract a generic `ProjectivePoint` type which impls these formulas for `C: PrimeCurve`.

tarcieri:
An unaddressed problem is testing the implementation in absence of a concrete curve implementation.

Perhaps the `dev` module could be further implemented to where it contains the core of e.g. P-256.

tarcieri:
This PR adapts the `p384` crate to use this generic implementation: https://github.com/RustCrypto/elliptic-curves/pull/601

newpavlov:
What about making a fully generic implementation of Weierstrass curves as I tried long ago with [elliptic-curves/pull/218](https://github.com/RustCrypto/elliptic-curves/pull/218)?

tarcieri:
@newpavlov yes, that would be the longer term goal, but as noted in the commit messages doing that effectively would require breaking changes which this implementation is trying to avoid for now.

I can write up some potential designs. One would be changing `PrimeCurve` from a marker trait to one which actually incorporates these formulas (and provides a blanket impl of the other traits like `ScalarArithmetic` and `ProjectiveArithmetic`)

PR #1024, opened 2022-06-13
elliptic-curve: allow bigger `c1` constant in `OsswuMapParams`

daxpedda:
Apparently this was need for https://github.com/RustCrypto/elliptic-curves/pull/600. It will also be necessary for P-512. I assume this is a breaking change?

### DIFF
--- a/elliptic-curve/src/hash2curve/osswu.rs
+++ b/elliptic-curve/src/hash2curve/osswu.rs
@@ -11,7 +11,7 @@ where
     F: Field,
 {
     /// The first constant term
-    pub c1: [u64; 4],
+    pub c1: &'static [u64],
     /// The second constant term
     pub c2: F,
     /// The ISO A variable or Curve A variable


elliptic-curve/src/hash2curve/osswu.rs
```diff
@@ -11,7 +11,7 @@ where
     F: Field,
 {
     /// The first constant term
-    pub c1: [u64; 4],
+    pub c1: &'static [u64],
```
tarcieri, line 14: Could this be represented by something like [`Curve::UInt`](https://docs.rs/elliptic-curve/latest/elliptic_curve/trait.Curve.html#associatedtype.UInt) instead?

That can parse a hex string to an word-sized number of limbs.

tarcieri:
Yeah, breaking change.

elliptic-curve/src/hash2curve/osswu.rs
daxpedda: I think that would work and would also be much better. The only problem is that `OsswuMap` is implemented over `FieldElement`, so I'm not sure how exactly to proceed.

As far as I'm aware there is not link between `Curve` and it's `FieldElement`, so I can't depend on that either.

I could depend on `GroupDigest` instead directly on `FieldElement`, which establishes exactly that link.
Alternatively I could just add an associated type in `OsswuMap` that holds the curve.

tarcieri: Aah, that's unfortunate, although short of the `ff` crate adopting something like `crypto-bigint` I'm not sure what better representation they could use

daxpedda: Alright, gonna mark this as resolved then, feel free to open it again if you can think of a solution.

tarcieri: FWIW I opened an upstream issue. Perhaps we could get a change to `ff` in before the next breaking change to the `elliptic-curve` crate:

https://github.com/zkcrypto/ff/issues/85

tarcieri:
The new `ff` and `group` are out so I guess we can begin making breaking changes again

tarcieri (approved):

elliptic-curve/src/hash2curve/osswu.rs
tarcieri: And I guess this restriction still exists in `ff` v0.13.0 so I'll call this resolved for now.

PR #1026, opened 2022-06-16
Add `password_hash::Error::OutputSize`

daxpedda:
See https://github.com/RustCrypto/password-hashes/pull/313.

Sometimes a very specific output size is expected. Any bike-shedding is very welcome.


password-hash/src/errors.rs
```diff
@@ -64,6 +67,9 @@ impl fmt::Display for Error {
             Self::B64Encoding(err) => write!(f, "{}", err),
             Self::Crypto => write!(f, "cryptographic error"),
             Self::OutputTooShort => f.write_str("PHF output too short (min 10-bytes)"),
+            Self::OutputSize(expected) => {
+                write!(f, "unexpected output size, expected {} bytes", expected)
+            }
             Self::OutputTooLong => f.write_str("PHF output too long (max 64-bytes)"),
```
tarcieri, line 73: It seems like it the best outcome would be to consolidate all of these 3 cases into `OutputSize`, and for consistency do the same for the other `*TooLong`/`*TooShort` parameters

daxpedda: I'm not sure if that's possible, just attempted it when I realized there is a lot of code without an expected size, just a minimum and maximum.

Alternatively, we could treat the minimum or maximum size as expected, but the error message might be confusing. I could store a `core::cmp::Ordering` in the error type?

Otherwise I would say the PR is fine like that.

daxpedda: @tarcieri friendly ping, how should I proceed?

tarcieri: Perhaps we could collapse these all into a single `Self::OutputSize(...)` variant with an inner `pub enum OutputSizeError` to handle the various cases?

daxpedda: Let me try!

daxpedda: So collapsed all the output size errors and even passed along actual expected sizes.
I could not consolidate it with PHC string errors because they are parsing and don't know how many bytes they are actually going to expect.

I honestly don't find this abstraction useful, PHC string errors are about not having enough or leftover bytes. Output size errors are different, they can tell you in detail what the size should be.

Suggestions:
1. Remove all the size information from output size errors, enabling consolidating everything.
2. Introduce `Option` into `OutputSizeError`, enabling it to handle both cases.
3. Leaving it like that and improve/differentiate PHC string errors in wording.

daxpedda: I basically went with suggestion 3. now. I think the PHC string errors "too short" and "too long" didn't make any sense anyway. Maybe we want to add a third error signifying when it's empty, but I believe that telling the user a PHC field is missing is still accurate.

At this point I would actually suggest removing `OutputSizeError` again and just integrate it's variants into `Error`. But gonna await your call on that.

tarcieri: The reason I suggested `OutputSizeError` in the first place was to handle potential cases where the size wasn't known, but if we don't need that, I agree the enum is too much overhead. There's little sense to have three variants that all have the same fields.

I almost suggested using `expected: core::cmp::Ordering` and `size: Option<usize>` fields to represent the various conditions but I thought a dedicated enum could potentially be clearer.

daxpedda: Alright, re-integrated.
> There's little sense to have three variants that all have the same fields.

This is still the case I guess.
> I almost suggested using `expected: core::cmp::Ordering` and `size: Option<usize>` fields to represent the various conditions but I thought a dedicated enum could potentially be clearer.

This is basically the only way to avoid that, it's what I did in `balloon-hash` too https://github.com/RustCrypto/password-hashes/pull/313.

---

I'm pretty comfortable with how it is right now, happy to change it according to any further instructions.

tarcieri:
This is back to having three different variants that effectively mean different flavors of the same thing.

What about this instead?

```rust
OutputSize { expected: Ordering, size: usize }
```

daxpedda:
> ```rust
> OutputSize { expected: Ordering, size: usize }
> ```

I wasn't sure how you mean that, I kinda turned it around, the `usize` holds the `expect`ed size and the `Ordering` holds if the actual `size` is too long, too short, or unequal (doesn't match properly I guess).

Also squashed the commits.

password-hash/src/errors.rs
```diff
@@ -20,11 +21,21 @@ pub enum Error {
...
-    /// Output too long (max 64-bytes).
-    OutputTooLong,
+    /// Output size unexpected.
+    OutputSize {
+        /// Indicates why the output size is unexpected.
+        ///
+        /// - [`Ordering::Less`]: Size is too small.
+        /// - [`Ordering::Equal`]: Size is not exactly as `expected`.
+        /// - [`Ordering::Greater`]: Size is too long.
+        size: Ordering,
```
tarcieri, line 31: How about something like `provided` instead of `size`?

daxpedda: Great idea!

daxpedda: Done.

tarcieri (approved):
This looks good to me now.

It's a breaking change though, so I might keep this open until we're ready to start working on the next version of `password-hash`.

daxpedda:
@tarcieri just saw the new update to 0.5.0-pre.0, are we ready to add this breaking change now?

password-hash/src/errors.rs
```diff
@@ -41,14 +52,11 @@ pub enum Error {
...
-
-    /// Password hash string too short.
-    PhcStringTooShort,
+    /// Password hash string invalid.
+    PhcStringField,
 
-    /// Password hash string too long.
-    PhcStringTooLong,
+    /// Password hash string contains leftover characters.
+    PhcStringLeftover,
```
tarcieri, line 59: I think this phrase has precedence in this situation:

```suggestion
    /// Password hash string contains trailing data.
    PhcStringTrailingData
```

tarcieri:
Thanks!

PR #1033, opened 2022-06-28
aead: use `crypto-common`

tarcieri:
Replaces the `NewAead` trait with the same-shaped `KeyInit` trait.


aead/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "aead"
-version = "0.4.3" # Also update html_root_url in lib.rs when bumping this
+version = "0.5.0-pre" # Also update html_root_url in lib.rs when bumping this
```
tarcieri, line 3: Note: breaking change

PR #1034, opened 2022-06-28
crypto-common: add `getrandom` feature

tarcieri:
Transitively enables `rand_core/getrandom`, making it possible to access `crypto_common::rand_core::OsRng`.


tarcieri:
This is hopefully uncontroversial enough. Let me know if you disagree.

Merging as I'd like to do some formatting cleanups on the surrounding code. Worst case we can revert before a release.

PR #1048, opened 2022-07-06
elliptic-curve: fix outdated doc

agostbiro:
The `zeroize` feature flag has been removed, zeroization on drop is now done by default.


elliptic-curve/src/secret_key.rs
```diff
@@ -3,9 +3,6 @@
 //! The [`SecretKey`] type is a wrapper around a secret scalar value which is
 //! designed to prevent unintentional exposure (e.g. via `Debug` or other
 //! logging).
-//!
-//! When the `zeroize` feature of this crate is enabled, it also handles
-//! zeroing it out of memory securely on drop.
```
tarcieri, line 8: Perhaps this should still mention the secret is zeroed out of memory, but unconditionally?

PR #1052, opened 2022-07-16
crypto-common: move `ParBlocks*` from `cipher` crate

tarcieri:
Moves `ParBlocks`/`ParBlocksSizeUser` from the `cipher` crate so it can be reused in the `universal-hash` crate (see #965, #1051).

The `cipher` crate re-exports them in an API-compatible way, so this is not a breaking change.


HACKER007-SHIP-IT:
[sm2](https://github.com/RustCrypto/elliptic-curves/actions/workflows/sm2.yml)

PR #1062, opened 2022-07-31
signature: make rnd mutable in RandomSigner and RandomDigestSigner

lumag:
Passing the random generator without an ability to change it in the
function makes little sense, as RngCore::fill_bytes() requires mutable
RngCore.

Signed-off-by: Dmitry Baryshkov <dmitry.baryshkov@linaro.org>

### DIFF
--- a/signature/src/signer.rs
+++ b/signature/src/signer.rs
@@ -95,7 +95,7 @@ where
 #[cfg_attr(docsrs, doc(cfg(feature = "rand-preview")))]
 pub trait RandomizedSigner<S: Signature> {
     /// Sign the given message and return a digital signature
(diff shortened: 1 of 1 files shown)


tarcieri:
This change is unnecessary. Also heads up: the `signature` crate is post-1.0, which significantly limits the non-additive changes we can make.

Any mutable reference to a type which impls `CryptoRng + RngCore` can be used via [blanket impls of the traits for their own mutable reference types](https://docs.rs/rand_core/0.6.3/rand_core/trait.RngCore.html#impl-RngCore-for-%26%27a%20mut%20R).

You can see it in action here:

https://github.com/RustCrypto/elliptic-curves/blob/c48ae6654ed9ece63a9a80a80c0f96efacb076e8/k256/src/ecdsa/sign.rs#L144-L148

PR #1078, opened 2022-08-10
crypto-common: add SerializableState trait

rpiasetskyi:
This trait is used for saving internal state of the hash core and
restoring it later.

The original issue was discussed here: https://github.com/RustCrypto/hashes/issues/310

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -31,7 +31,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "433cfd6710c9986c576a25ca913c39d66a6474107b406f34f91d4a8923395241"
 dependencies = [
  "cfg-if",
- "cipher 0.4.3 (registry+https://github.com/rust-lang/crates.io-index)",
+ "cipher 0.4.3",
  "cpufeatures",
 ]
 
@@ -43,7 +43,7 @@ checksum = "82e1366e0c69c9f927b1fa5ce2c7bf9eafc8f9268c0b9800729e8b267612447c"
 dependencies = [
  "aead 0.5.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "aes",
- "cipher 0.4.3 (registry+https://github.com/rust-lang/crates.io-index)",
+ "cipher 0.4.3",
  "ctr",
  "ghash",
  "subtle",
@@ -65,9 +65,9 @@ dependencies = [
 
 [[package]]
 name = "async-trait"
-version = "0.1.60"
+version = "0.1.66"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "677d1d8ab452a3936018a687b20e6f7cf5363d713b732b8884001317b0e48aa3"
+checksum = "b84f9ebcc6c1f5b8cb160f6990096a5c127f423fcb6e1ccc46c370cbdfb75dfc"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -97,9 +97,9 @@ checksum = "349a06037c7bf932dd7e7d1f653678b2038b9ad46a74102f1fc7bd7872678cce"
 
 [[package]]
 name = "base64ct"
-version = "1.6.0"
+version = "1.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8c3c1a368f70d6cf7302d78f8f7093da241fb8e8807c05cc9e51a125895a6d5b"
+checksum = "e6b4d9b1225d28d360ec6a231d65af1fd99a2a095154c8040689617290569c5c"
 
 [[package]]
 name = "bincode"
@@ -167,9 +167,9 @@ checksum = "89b2fd2a0dcf38d7971e2194b6b6eebab45ae01067456a7fd93d5547a61b70be"
 
 [[package]]
 name = "cc"
-version = "1.0.78"
+version = "1.0.79"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a20104e2335ce8a659d6dd92a51a767a0c062599c73b343fd152cb401e828c3d"
+checksum = "50d30906286121d95be3d479533b458f87493b30a4b5f79a607db8f5d11aa91f"
 dependencies = [
  "jobserver",
 ]
@@ -187,7 +187,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c7fc89c7c5b9e7a02dfe45cd2367bae382f9ed31c61ca8debe5f827c420a2f08"
 dependencies = [
  "cfg-if",
- "cipher 0.4.3 (registry+https://github.com/rust-lang/crates.io-index)",
+ "cipher 0.4.3",
  "cpufeatures",
 ]
 
@@ -199,18 +199,8 @@ checksum = "10cd79432192d1c0f4e1a0fef9527696cc039165d729fb41b3f4f4f354c2dc35"
 dependencies = [
  "aead 0.5.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "chacha20",
- "cipher 0.4.3 (registry+https://github.com/rust-lang/crates.io-index)",
+ "cipher 0.4.3",
  "poly1305",
- "zeroize",
-]
-
-[[package]]
-name = "cipher"
-version = "0.4.3"
-dependencies = [
- "blobby",
- "crypto-common 0.1.6",
- "inout",
  "zeroize",
 ]
 
@@ -220,6 +210,16 @@ version = "0.4.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d1873270f8f7942c191139cb8a40fd228da6c3fd2fc376d7e92d47aa14aeb59e"
 dependencies = [
+ "crypto-common 0.1.6",
+ "inout",
+ "zeroize",
+]
+
+[[package]]
+name = "cipher"
+version = "0.4.4"
+dependencies = [
+ "blobby",
  "crypto-common 0.1.6",
  "inout",
  "zeroize",
@@ -257,7 +257,7 @@ name = "crypto"
 version = "0.5.0-pre"
 dependencies = [
  "aead 0.5.1",
- "cipher 0.4.3 (registry+https://github.com/rust-lang/crates.io-index)",
+ "cipher 0.4.3",
  "crypto-common 0.1.6",
  "digest 0.10.6",
  "elliptic-curve 0.12.3",
@@ -337,7 +337,7 @@ version = "0.9.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0369ee1ad671834580515889b80f2ea915f23b8be8d0daa4bbaf2ac5c7590835"
 dependencies = [
- "cipher 0.4.3 (registry+https://github.com/rust-lang/crates.io-index)",
+ "cipher 0.4.3",
 ]
 
 [[package]]
@@ -510,9 +510,9 @@ dependencies = [
 
 [[package]]
 name = "glob"
-version = "0.3.0"
+version = "0.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9b919933a397b79c37e33b77bb2aa3dc8eb6e165ad809e58ff75bc7db2e34574"
+checksum = "d2fabcfbdc87f4758337ca535fb41a6d701b65693ce38287d856d1674551ec9b"
 
 [[package]]
 name = "group"
@@ -636,9 +636,9 @@ dependencies = [
 
 [[package]]
 name = "jobserver"
-version = "0.1.25"
+version = "0.1.26"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "068b1ee6743e4d11fb9c6a1e6064b3693a1b600e7f5f5988047d98b3dc9fb90b"
+checksum = "936cfd212a0155903bcbc060e316fb6cc7cbf2e1907329391ebadc1fe0ce77c2"
 dependencies = [
  "libc",
 ]
@@ -660,9 +660,9 @@ dependencies = [
 
 [[package]]
 name = "libc"
-version = "0.2.138"
+version = "0.2.139"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "db6d7e329c562c5dfab7a46a2afabc8b987ab9a4834c9d1ca04dc54c1546cef8"
+checksum = "201de327520df007757c1f0adce6e827fe8562fbc28bfd9c15571c66ca1f5f79"
 
 [[package]]
 name = "lock_api"
@@ -711,9 +711,9 @@ dependencies = [
 
 [[package]]
 name = "pem-rfc7468"
-version = "0.2.3"
+version = "0.2.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8f22eb0e3c593294a99e9ff4b24cf6b752d43f193aa4415fe5077c159996d497"
+checksum = "84e93a3b1cc0510b03020f33f21e62acdde3dcaef432edc95bea377fbd4c2cd4"
 dependencies = [
  "base64ct",
 ]
@@ -812,9 +812,9 @@ checksum = "97e91cb6af081c6daad5fa705f8adb0634c027662052cb3174bdf2957bf07e25"
 
 [[package]]
 name = "proc-macro2"
-version = "1.0.49"
+version = "1.0.51"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "57a8eca9f9c4ffde41714334dee777596264c7825420f521abc92b5b5deb63a5"
+checksum = "5d727cae5b39d21da60fa540906919ad737832fe0b1c165da3a34d6548c849d6"
 dependencies = [
  "unicode-ident",
 ]
@@ -901,27 +901,27 @@ checksum = "58bc9567378fc7690d6b2addae4e60ac2eeea07becb2c64b9f218b53865cba2a"
 
 [[package]]
 name = "serde"
-version = "1.0.151"
+version = "1.0.154"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "97fed41fc1a24994d044e6db6935e69511a1153b52c15eb42493b26fa87feba0"
+checksum = "8cdd151213925e7f1ab45a9bbfb129316bd00799784b174b7cc7bcd16961c49e"
 dependencies = [
  "serde_derive",
 ]
 
 [[package]]
 name = "serde_bytes"
-version = "0.11.8"
+version = "0.11.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "718dc5fff5b36f99093fc49b280cfc96ce6fc824317783bff5a1fed0c7a64819"
+checksum = "416bda436f9aab92e02c8e10d49a15ddd339cea90b6e340fe51ed97abb548294"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "serde_derive"
-version = "1.0.151"
+version = "1.0.154"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "255abe9a125a985c05190d687b320c12f9b1f0b99445e608c21ba0782c719ad8"
+checksum = "4fc80d722935453bcafdc2c9a73cd6fac4dc1938f0346035d84bf99fa9e33217"
 dependencies = [
  "proc-macro2",
  "quote",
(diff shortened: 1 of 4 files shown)


digest/src/core_api.rs
```diff
@@ -117,3 +117,15 @@ pub enum TruncSide {
     /// Truncate right side, i.e. `&out[m..]`.
     Right,
 }
+
+/// Core trait for saving internal state of the hash core and restoring it later.
+pub trait HashCoreInternalState {
```
tarcieri, line 122: I'd suggest using Rust's built-in conversion traits rather than defining special methods like `internal_state` and `from_internal_state`:

```suggestion
pub trait HashCoreInternalState: TryFrom<Self::InternalState, Error = Error> + Into<Self::InternalState> {
```

tarcieri:
I think the core notion here is a "serializable state" rather than an "internal state"

digest/src/core_api.rs
```diff
@@ -117,3 +117,15 @@ pub enum TruncSide {
...
+    /// Internal state of the hash core.
+    type InternalState;
```
tarcieri, line 124: It'd be good to have some bounds here to enable generic code:

```suggestion
    type InternalState: AsRef<[u8]>;
```

rpiasetskyi: If the internal state consists of several objects, is it possible to implement `TryFrom<&[u8]> + AsRef<[u8]>`: e.g. internal state of SHA-2 hashes consists of the intermediate hash value (array of `u32`/`u64`) and the number of consumed blocks (number `u64`/`u128`).

tarcieri: The use case of this API is serialization.

If you think there's a use case beyond serialization, can you explain how that use case isn't satisfied by simply cloning an existing digest instance?

rpiasetskyi: Yes, the use case is serialization only.
Should I then create a byte array and serialize the internal state on my own?
E.g. for `Sha256VarCore`, I need to allocate an array, copy the intermediate hash value (`[u32; STATE_LEN]`), and the number of consumed blocks (`u64`).
Or is it better to return an object that consists of the intermediate hash value and the number of consumed blocks, and this object is `Serializable` (`serde`)? But this object can not implement `AsRef<[u8]>` because it consists of a byte array and number.

tarcieri: It depends if the state can always be fixed-width or not. If it can, we should probably use an associated `generic_array::ArrayLength<u8>` instead.

I think it's a pretty reasonable assumption that most digests could support a fixed-width serializable state.

newpavlov (commented):
I agree that it's worth to introduce trait which outputs generic array. Something like this:
```rust
pub type SerializedState<T> = GenericArray<u8, <T as SerializableState>::SerializedStateSize>;

pub trait SerializableState {
    type SerializedStateSize: ArrayLength<u8>;
    
    fn serialize(&self) -> SerializedState<Self>;
    fn deserialize(buf: &SerializedState<Self>) -> Self;
}
```
Also do not forget to add warning that serialized state may contain sensitive information, e.g. in the hashes case it can contain parts of unprocessed message.

The trait may go to the `crypto-common` crate since it may be useful for other algorithms as well.

rpiasetskyi:
Sorry, missed the last comment. I will move the trait to `crypto-common` and use `serialize/deserialize` methods.

Should `deserialize` method return `Result` (if the buffer contains invalid values)?

rpiasetskyi:
The versions of the crates have to be updated: should they be `0.1.7/0.10.4` or `0.2.0/0.11.0`?

Right now crypto-common is `0.1.6` and digest is `0.10.3`.

digest/src/core_api/ct_variable.rs
```diff
@@ -165,3 +165,26 @@ where
         write!(f, "{}", OutSize::USIZE)
     }
 }
+
+impl<T, OutSize> SerializableState for CtVariableCoreWrapper<T, OutSize>
```
newpavlov, line 169: Do not forget to also implement it for `CoreWrapper` and ideally for `RtVariableCoreWrapper`.

newpavlov (commented):
>Should deserialize method return Result (if the buffer contains invalid values)?

Yes, I forgot to do it in my snippet. I think we can define the error type in `crypto-common` (i.e. no need to make it generic).

>The versions of the crates have to be updated: should they be 0.1.7/0.10.4 or 0.2.0/0.11.0?

The changes are backwards-compatible, so patch releases should be fine.

digest/src/core_api/ct_variable.rs
rpiasetskyi: `CoreWrapper` has [`decompose`](https://docs.rs/digest/0.10.3/digest/core_api/struct.CoreWrapper.html#method.decompose) method that returns buffered input and `CtVariableCoreWrapper` object, and [`from_core`](https://docs.rs/digest/0.10.3/digest/core_api/struct.CoreWrapper.html#method.from_core) method to restore the object.

Do we need additionally implement `SerializableState`? Also, we do not know the size of buffered input (or we need to store the length with the partially filled block of buffered input).

newpavlov: I think, yes. It would be weird to support `SerializableState` only for the core types and force users to jump through artificial error-prone hoops.

> Also, we do not know the size of buffered input (or we need to store the length with the partially filled block of buffered input).

You should attach to serialized state number of unprocessed bytes and zero-padded unprocessed data (you could use `buffer.clone().pad_with_zeros().clone()`). For SHA-256 it would mean always attaching 65 bytes, even when there are no unprocessed bytes in it. We could save one byte for the eager hashes by using the fact that the last byte will be never used, but I think it's too small of a win for the very niche feature, so it does not worth to introduce the additional complexity.

crypto-common/src/lib.rs
```diff
@@ -309,3 +309,34 @@ impl fmt::Display for InvalidLength {
...
+impl fmt::Display for DeserializeStateError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
+        f.write_str("Deserialization error")
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for DeserializeStateError {}
+
+/// Types which can serialize the internal state and be restored from it.
```
newpavlov, line 329: Do not forget to add the warning that serialized state may contain sensitive data in plain text.

tarcieri:
The serialized state needs to be consistent regardless of the target pointer width.

You should pick a sensible on-the-wire representation (`u32` or `u64`) and use a `try_from` conversion when deserializing it.

newpavlov:
I think we can simply transform `output_size` to `u8` using `TryInto` and `unwrap`. I am not aware about variable hashes which support output size bigger than 255 bytes (XOF does not count). In future versions it could be worth to replace `usize` with `u8`.

digest/src/core_api/ct_variable.rs
```diff
@@ -165,3 +166,25 @@ where
...
+    OutSize: ArrayLength<u8> + IsLessOrEqual<T::OutputSize>,
+    LeEq<OutSize, T::OutputSize>: NonZero,
+    T::BlockSize: IsLess<U256>,
+    Le<T::BlockSize, U256>: NonZero,
+{
+    type SerializedStateSize = T::SerializedStateSize;
+
+    fn serialize(&self) -> SerializedState<Self> {
+        self.inner.serialize()
+    }
```
newpavlov, line 182: As a sanity check it could be worth to add into serialized state output size (the `OutSize` type parameter) and return an error during deserialization if it does not match.

digest/src/core_api/rt_variable.rs
```diff
@@ -145,6 +149,68 @@ where
...
+impl<T> SerializableState for RtVariableCoreWrapper<T>
+where
+    T: VariableOutputCore + UpdateCore + SerializableState,
+    T::BlockSize: IsLess<U256>,
+    Le<T::BlockSize, U256>: NonZero,
+    T::SerializedStateSize: Add<U9>,
+    <T::SerializedStateSize as Add<U9>>::Output: Add<T::BlockSize>,
+    <<T::SerializedStateSize as Add<U9>>::Output as Add<T::BlockSize>>::Output: ArrayLength<u8>,
+{
+    type SerializedStateSize = Sum<Sum<T::SerializedStateSize, U9>, T::BlockSize>;
```
newpavlov, line 161: I think you forgot to change `U9` to `U2`.

digest/src/core_api/rt_variable.rs
```diff
@@ -145,6 +149,68 @@ where
...
+        }
+
+        let (block_buffer, serialization_buffer) =
+            serialization_buffer.split_at(T::BlockSize::USIZE);
+
+        let output_size = serialization_buffer[0].into();
+
+        Ok(Self {
+            core: T::deserialize(serialized_core)?,
+            buffer: BlockBuffer::new(&block_buffer[..pos]),
```
newpavlov, line 208: Note that if the buffer is "eager" and `pos` is equal to block size, `new` will panic. I guess the simplest solution will be to add `try_new` method to `BlockBuffer`.

digest/src/core_api/wrapper.rs
```diff
@@ -227,6 +228,61 @@ where
...
+        let pos = buffer[0].into();
+        if pos > T::BlockSize::USIZE {
+            return Err(DeserializeStateError);
+        }
+
+        let (block_buffer, _) = serialization_buffer.split_at(pos);
+
+        Ok(Self {
+            core: T::deserialize(serialized_core)?,
+            buffer: BlockBuffer::new(block_buffer),
```
newpavlov, line 281: Same as in the previous comment.

newpavlov (commented):
You could use the [`concat`](https://docs.rs/generic-array/latest/generic_array/struct.GenericArray.html#method.concat) method to slightly improve the code, though it could force you to make `SerializedStateSize` less... pretty. So it's quite optional.

digest/src/core_api/rt_variable.rs
rpiasetskyi: https://github.com/RustCrypto/utils/pull/799

rpiasetskyi:
Converted to draft to avoid emails about failed tests.

digest/Cargo.toml
```diff
@@ -15,7 +15,7 @@ categories = ["cryptography", "no-std"]
 crypto-common = { version = "0.1.3", path = "../crypto-common" }
 
 # optional dependencies
-block-buffer = { version = "0.10", optional = true }
+block-buffer = { version = ">=0.10.3", optional = true }
```
newpavlov, line 18: I think `0.10.3` should be sufficient, i.e. the `>=` part is redundant.

digest/src/core_api/rt_variable.rs
```diff
@@ -145,6 +156,54 @@ where
...
+    Le<T::BlockSize, U256>: NonZero,
+    T::SerializedStateSize: Add<U1>,
+    <T::SerializedStateSize as Add<U1>>::Output: Add<T::BlockSize> + ArrayLength<u8>,
+    <<T::SerializedStateSize as Add<U1>>::Output as Add<T::BlockSize>>::Output: Add<U1> + ArrayLength<u8>,
+    <<<T::SerializedStateSize as Add<U1>>::Output as Add<T::BlockSize>>::Output as Add<U1>>::Output: ArrayLength<u8> + Sub<T::SerializedStateSize>,
+    <<<<T::SerializedStateSize as Add<U1>>::Output as Add<T::BlockSize>>::Output as Add<U1>>::Output as Sub<T::SerializedStateSize>>::Output: Sub<U1> + ArrayLength<u8>,
+    <<<<<T::SerializedStateSize as Add<U1>>::Output as Add<T::BlockSize>>::Output as Add<U1>>::Output as Sub<T::SerializedStateSize>>::Output as Sub<U1>>::Output: Sub<T::BlockSize> + ArrayLength<u8>,
+    <<<<<<T::SerializedStateSize as Add<U1>>::Output as Add<T::BlockSize>>::Output as Add<U1>>::Output as Sub<T::SerializedStateSize>>::Output as Sub<U1>>::Output as Sub<T::BlockSize>>::Output: ArrayLength<u8>,
+{
+    type SerializedStateSize = Sum<Sum<Sum<T::SerializedStateSize, U1>, T::BlockSize>, U1>;
```
newpavlov, line 172: You can add ignore attribute for this associated type to work around the clippy warnings.

digest/src/core_api/rt_variable.rs
```diff
@@ -145,6 +156,54 @@ where
...
+    T: VariableOutputCore + UpdateCore + SerializableState,
+    T::BlockSize: IsLess<U256>,
+    Le<T::BlockSize, U256>: NonZero,
+    T::SerializedStateSize: Add<U1>,
+    <T::SerializedStateSize as Add<U1>>::Output: Add<T::BlockSize> + ArrayLength<u8>,
+    <<T::SerializedStateSize as Add<U1>>::Output as Add<T::BlockSize>>::Output: Add<U1> + ArrayLength<u8>,
+    <<<T::SerializedStateSize as Add<U1>>::Output as Add<T::BlockSize>>::Output as Add<U1>>::Output: ArrayLength<u8> + Sub<T::SerializedStateSize>,
+    <<<<T::SerializedStateSize as Add<U1>>::Output as Add<T::BlockSize>>::Output as Add<U1>>::Output as Sub<T::SerializedStateSize>>::Output: Sub<U1> + ArrayLength<u8>,
+    <<<<<T::SerializedStateSize as Add<U1>>::Output as Add<T::BlockSize>>::Output as Add<U1>>::Output as Sub<T::SerializedStateSize>>::Output as Sub<U1>>::Output: Sub<T::BlockSize> + ArrayLength<u8>,
+    <<<<<<T::SerializedStateSize as Add<U1>>::Output as Add<T::BlockSize>>::Output as Add<U1>>::Output as Sub<T::SerializedStateSize>>::Output as Sub<U1>>::Output as Sub<T::BlockSize>>::Output: ArrayLength<u8>,
```
newpavlov, line 170: I expected something similar, but still... :)

I think you can simplify it a fair bit by using [operator aliases](https://docs.rs/typenum/latest/typenum/operator_aliases/index.html) in bounds as well, e.g. you can replace:
```rust
<T::SerializedStateSize as Add<U1>>::Output: Add<T::BlockSize> + ArrayLength<u8>,
```
with:
```rust
Sum<T::SerializedStateSize, U1>: Add<T::BlockSize> + ArrayLength<u8>,
```

digest/src/core_api/rt_variable.rs
rpiasetskyi: I have checked that the current version of `clippy` doesn't treat this as a warning anymore. The first version that does it is `1.59.0`.
Does it make sense to update the version of clippy (to the current stable `1.62.0` or minimal version `1.59.0`)? Or is it better to add an attribute?

newpavlov: Yes, you can update clippy's version in CI.

rpiasetskyi:
Does it make sense to implement `SerializableState` for some commonly used types?

```
impl SerializableState for u8 {
    type SerializedStateSize = U1;

    fn serialize(&self) -> SerializedState<Self> {
        [*self].into()
    }

    fn deserialize(
        serialized_state: &SerializedState<Self>,
    ) -> Result<Self, DeserializeStateError> {
        Ok(serialized_state[0])
    }
}

impl SerializableState for u64 {
    type SerializedStateSize = U8;

    fn serialize(&self) -> SerializedState<Self> {
        self.to_le_bytes().into()
    }

    fn deserialize(
        serialized_state: &SerializedState<Self>,
    ) -> Result<Self, DeserializeStateError> {
        Ok(u64::from_le_bytes((*serialized_state).into()))
    }
}

impl SerializableState for [u64; 16] {
    type SerializedStateSize = U128;

    fn serialize(&self) -> SerializedState<Self> {
        let mut serialized_state = SerializedState::<Self>::default();
        for (val, chunk) in self.iter().zip(serialized_state.chunks_exact_mut(8)) {
            chunk.copy_from_slice(&val.to_le_bytes());
        }

        serialized_state
    }

    fn deserialize(
        serialized_state: &SerializedState<Self>,
    ) -> Result<Self, DeserializeStateError> {
        let mut array = Self::default();

        for (val, chunk) in array.iter_mut().zip(serialized_state.chunks_exact(8)) {
            *val = u64::from_le_bytes(chunk.try_into().unwrap());
        }

        Ok(array)
    }
}
```

It can be implemented in a similar way to this https://github.com/fizyk20/generic-array/blob/master/src/impls.rs#L183.

This can move the for loop (array serialization/deserialization) in one place and make the code a little bit simpler.

```
type State = [u64; 16];

impl SerializableState for GroestlLongVarCore {
    type SerializedStateSize = Sum<
        <u64 as SerializableState>::SerializedStateSize,
        <State as SerializableState>::SerializedStateSize,
    >;

    fn serialize(&self) -> SerializedState<Self> {
        self.state.serialize().concat(self.blocks_len.serialize())
    }

    fn deserialize(
        serialized_state: &SerializedState<Self>,
    ) -> Result<Self, DeserializeStateError> {
        let (serialized_state, serialized_blocks_len) =
            Split::<_, <State as SerializableState>::SerializedStateSize>::split(serialized_state);

        Ok(Self {
            state: State::deserialize(serialized_state)?,
            blocks_len: u64::deserialize(serialized_blocks_len)?,
        })
    }
}
```

rpiasetskyi:
With the introduction of the `AssociatedOid` trait and new generic parameter for `CtVariableCoreWrapper`, is it needed to add it to serialized state?

tarcieri:
No, those are type-level properties which are identical regardless of the state.

rpiasetskyi:
> Does it make sense to implement `SerializableState` for some commonly used types?
> 
> ```
> impl SerializableState for u8 {
>     type SerializedStateSize = U1;
> 
>     fn serialize(&self) -> SerializedState<Self> {
>         [*self].into()
>     }
> 
>     fn deserialize(
>         serialized_state: &SerializedState<Self>,
>     ) -> Result<Self, DeserializeStateError> {
>         Ok(serialized_state[0])
>     }
> }
> 
> impl SerializableState for u64 {
>     type SerializedStateSize = U8;
> 
>     fn serialize(&self) -> SerializedState<Self> {
>         self.to_le_bytes().into()
>     }
> 
>     fn deserialize(
>         serialized_state: &SerializedState<Self>,
>     ) -> Result<Self, DeserializeStateError> {
>         Ok(u64::from_le_bytes((*serialized_state).into()))
>     }
> }
> 
> impl SerializableState for [u64; 16] {
>     type SerializedStateSize = U128;
> 
>     fn serialize(&self) -> SerializedState<Self> {
>         let mut serialized_state = SerializedState::<Self>::default();
>         for (val, chunk) in self.iter().zip(serialized_state.chunks_exact_mut(8)) {
>             chunk.copy_from_slice(&val.to_le_bytes());
>         }
> 
>         serialized_state
>     }
> 
>     fn deserialize(
>         serialized_state: &SerializedState<Self>,
>     ) -> Result<Self, DeserializeStateError> {
>         let mut array = Self::default();
> 
>         for (val, chunk) in array.iter_mut().zip(serialized_state.chunks_exact(8)) {
>             *val = u64::from_le_bytes(chunk.try_into().unwrap());
>         }
> 
>         Ok(array)
>     }
> }
> ```
> 
> It can be implemented in a similar way to this https://github.com/fizyk20/generic-array/blob/master/src/impls.rs#L183.
> 
> This can move the for loop (array serialization/deserialization) in one place and make the code a little bit simpler.
> 
> ```
> type State = [u64; 16];
> 
> impl SerializableState for GroestlLongVarCore {
>     type SerializedStateSize = Sum<
>         <u64 as SerializableState>::SerializedStateSize,
>         <State as SerializableState>::SerializedStateSize,
>     >;
> 
>     fn serialize(&self) -> SerializedState<Self> {
>         self.state.serialize().concat(self.blocks_len.serialize())
>     }
> 
>     fn deserialize(
>         serialized_state: &SerializedState<Self>,
>     ) -> Result<Self, DeserializeStateError> {
>         let (serialized_state, serialized_blocks_len) =
>             Split::<_, <State as SerializableState>::SerializedStateSize>::split(serialized_state);
> 
>         Ok(Self {
>             state: State::deserialize(serialized_state)?,
>             blocks_len: u64::deserialize(serialized_blocks_len)?,
>         })
>     }
> }
> ```
> 
> And it might be possible to implement `#[derive(SerializableState)]` for types that do not need to check anything.

What do you think about https://github.com/RustCrypto/traits/pull/1078/commits/3447449814c3ee2bae77834bd23d1984895603fa?
If this is okay, I will also add `#[derive(SerializableState)]`.

waynr:
:wave: any chance of getting this merged any time in the near future?

tarcieri:
@waynr it needs a rebase, at the very least

waynr:
@tarcieri I'm attempting to rebase but running into compile errors that are pretty much totally incomprehensible to me. This repo is like a firehose of generic types and my puny little brain only has capacity for a trickle.

tarcieri:
I think @rpiasetskyi said he would take a look

waynr:
Just a quick note - it looks like the issue of the local dependency on crypto-common being broken in master branch for the digest crate is fixed in #1358 so this PR might need to wait for that before being rebased.

rpiasetskyi:
I am doing a rebase (and migration to `hybrid_array`) using https://github.com/RustCrypto/traits/pull/1358.

rpiasetskyi:
@tarcieri Sorry, I was expecting it to update the PR but not to merge it :(

rpiasetskyi:
Reopen.

tarcieri:
I think you pushed 0 commits which auto-closed the PR? Regardless I can't reopen it... I think you'd need to push some new commits

PR #1080, opened 2022-08-13
signature: add `Keypair` trait

tarcieri:
Adds a trait for types which represent a combination of both a signing key and verifying key as is common in many digital signature systems.

The `Keypair` name follows Rust standard capitalization rules for the closed compound word "keypair" as commonly used in cryptography.

### DIFF
--- a/signature/src/keypair.rs
+++ b/signature/src/keypair.rs
@@ -1,0 +1,17 @@
+//! Signing keypairs.
+
+use crate::{Signature, Signer, Verifier};
+
+/// Signing keypair with an associated verifying key.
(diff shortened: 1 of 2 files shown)


tony-iqlusion:
Looks like this would require an MSRV bump due to cyclical bounds

tarcieri:
Looks like this would require an MSRV bump to 1.52 due to cyclical bounds.

Might as well do a 2021 edition bump in that case.

tomtau (approved):

dignifiedquire (approved):

PR #1088, opened 2022-08-22
Replace Choice::unwrap_u8() == 1 with Choice::into()

newpavlov:
Improvement of #1079


tarcieri (approved):

tarcieri (changes requested):
Hmm, was there a `cargo update` included in this PR? Seems something is breaking

newpavlov:
There is an [issue](https://github.com/RustCrypto/traits/runs/7957075435) with the `kem` crate related to the AEAD traits. Can you look into it?

tarcieri:
I think if you remove the changes to `Cargo.lock` the tests will pass

newpavlov:
For now I simply disabled the `rust-hpke` dependency and the relevant `kem` tests which were causing the compilation issue. I think it would be better for you to properly fix them in a different PR.

tarcieri:
Why was Cargo.lock changed at all? That's what's causing the test failures, and it doesn't look like `subtle` was modified.

newpavlov:
No, the update only revealed the dormant issue. I don't think it's worth to create a separate PR for only updating `Cargo.lock`.

newpavlov:
If you prefer it, I could split the Cargo.lock changes into a separate PR, but they will look exactly the same as here.

tarcieri:
I think it would make more sense to split those changes out.

It's a lot of unrelated changes to both try to update Cargo.lock and disable the `kem` tests because of it, in a PR which is otherwise just trying to modify 5 lines of code.

PR #1090, opened 2022-08-22
KEM: Updated HPKE deps for testing

rozbb:
HPKE tests in KEM compile again. Addresses problem noted in #1089


newpavlov:
Can you add a comment to the Cargo.toml regarding why we use `rust-hpke` from the repository instead of using a version published on crates.io? I found it quite confusing when I found that part.

rozbb:
Done. Let me know if that's clear. It's a real mess lately bc x25519-dalek hasn't been updated in a year. I may have to unilaterally cut a new version if nobody else steps up

newpavlov:
Thanks, looks good!

Issue #1069, opened 2022-08-02
digest: type name readability regression


tarcieri:
Digest now makes use of several vaguely-named generic wrapper structs and type aliases to give them human-readable names.

This makes both the type names emitted by rustc and the resulting rustdoc harder to read.

For example, what used to render in rustdoc as:

```rust
impl PrehashSignature for Signature {
    type Digest = Sha256;
}
```

now renders as:

```rust
impl PrehashSignature for Signature {
    type Digest = CoreWrapper<CtVariableCoreWrapper<Sha256VarCore, UInt<UInt<UInt<UInt<UInt<UInt<UTerm, B1>, B0>, B0>, B0>, B0>, B0>>>
}
```

e.g. https://docs.rs/k256/0.11.3/k256/schnorr/struct.Signature.html#impl-PrehashSignature

I think this usability issue and various others (e.g. confusing error messages relating to blanket impls of sealed traits) could be fixed by using newtypes around the generic core, rather than just type aliases.

newpavlov:
>could be fixed by using newtypes around the generic core, rather than just type aliases.

It will involve a quite significant amount of boilerplate in every hash crate.

tarcieri:
I think with some experimentation we can find a solution which allows newtypes for each hash type while minimizing per-crate boilerplate.

tarcieri:
Alternatively perhaps we could find more meaningful names for the generic types which make the generic types more readable.

Or a combination thereof.

tarcieri:
In https://github.com/RustCrypto/RSA/pull/179 we ran into a somewhat related issue:

It would be nice to be able to use the [`const_oid::AssociatedOid`](https://docs.rs/const-oid/latest/const_oid/trait.AssociatedOid.html) trait to associate an OID with types like `Sha224`/`Sha256`/`Sha384`/`Sha512`.

Unfortunately this isn't possible because these types are actually `CoreWrapper<CtVariableCoreWrapper<...>` as defined by the `digest` crate.

Something I think is worth noting is the orphan rules would probably permit such an impl if the inner generic type were defined in the current crate (leveraging the same exemption `From` uses), but that doesn't work with multiple layers of generics like this.

newpavlov:
I think `digest` can implement `AssociatedOid` for the wrappers if a wrapped type implements it. In other words, we would implement `AssociatedOid` for core types and `digest` will handle "bubbling" it to the types used by users.

newpavlov:
[...]

PR #1098, opened 2022-09-07
digest: add implementations of the AssociatedOid trait

newpavlov:
This PR adds implementation of the [const_oid::AssociatedOid](https://docs.rs/const-oid/latest/const_oid/trait.AssociatedOid.html) trait for the core wrappers. It allows to define OID for a core type and make it accessible for higher-level wrapped types.

Lack of these impls was mentioned in https://github.com/RustCrypto/traits/issues/1069#issuecomment-1238777845


newpavlov:
During creation of this PR I found that, unfortunately, it will not work as-is. SHA-224 and SHA-256 share the same core, but have different OIDs. I don't think we can resolve this issue without introducing breaking changes.

digest/src/lib.rs
```diff
@@ -60,6 +60,9 @@ mod mac;
 #[cfg(feature = "core-api")]
 #[cfg_attr(docsrs, doc(cfg(feature = "core-api")))]
 pub use block_buffer;
+#[cfg(feature = "const-oid")]
```
tarcieri, line 63: FWIW `der` uses `oid` as the feature name

lumag:
@newpavlov I suppose that using newtype for the digests would allow defining `AssociateOid`s. However this is definitely a bigger change.

newpavlov:
@lumag 
Manual newtypes would mean a HUGE amount of boilerplate. The only practical solution is to introduce a bunch of macros to `digest` instead of the generic wrappers, but I would like to avoid such solution if possible.

Unfortunately, we hit some of Rust limitations (e.g. currently it's impossible to use `ObjectIdentifier` as a generic const argument) which prevent straightforward implementation of optional OIDs on non-core types, but I have some ideas about potential workarounds, which I would like to try.

tarcieri:
See https://github.com/RustCrypto/traits/issues/1069 for some discussion of newtypes as a potential alternative (probably a better place to continue discussion than this particular PR)

newpavlov (commented):
I have added optional type parameter to `CtVariableCoreWrapper` which allows to pass OID for resulting hash function. Unfortunately, it means that we have to generate separate "carrier" types, but otherwise I think the workaround works quite well.

How addition of OIDs looks in implementation crates can be seen in https://github.com/RustCrypto/hashes/pull/405.

newpavlov:
Note that we still do not have a way to add `AssociatedOid` implementation for "compound" algorithms such as HMAC-SHA256, but I think it's better to discuss it in a separate issue.

digest/src/core_api/ct_variable.rs
```diff
@@ -5,17 +5,24 @@ use super::{
...
+/// Dummy type used with [`CtVariableCoreWrapper`] in cases when
+/// resulting hash does not have a known OID.
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
+pub struct NoOid;
+
 /// Wrapper around [`VariableOutputCore`] which selects output size
 /// at compile time.
 #[derive(Clone)]
-pub struct CtVariableCoreWrapper<T, OutSize>
+pub struct CtVariableCoreWrapper<T, OutSize, O = NoOid>
```
tarcieri, line 25: Instead of using an extra generic type here, could `CtVariableCoreWrapper` have an impl of `AssociatedOid` which delegates to `inner`, and is bounded on `inner` having an impl of `AssociatedOid`?

newpavlov: If I understand you correctly, then no, it will not work. See the SHA-224 vs SHA-256 issue raised in my earlier comment. They have the same core (i.e. the same `inner` type), so with such solution they can not have different OIDs.

tarcieri: Ugh, ok. That's really rather unfortunate.

tarcieri (approved):
Would be good to get this merged/release to make progress on https://github.com/RustCrypto/RSA/pull/183

PR #1099, opened 2022-09-09
signature: add `hazmat-preview` feature

tarcieri:
Adds a `hazmat` module gated under a newly added `hazmat-preview` feature which calls out the relevant functionality as subject to change with minor versions.

It adds the following traits:

- `PrehashSigner`
- `PrehashVerifier`

These APIs accept the digest to be signed/verified as a raw byte slice. This comes with potential misuses like failing to use a cryptographically secure hash function as the `prehash`, which could enable existential forgeries of signatures, hence gating it under a `hazmat-preview` feature and placing it in a `hazmat` module.

Note that we previously explored APIs like this for `DigestSigner`. They were removed in RustCrypto/signatures#17 due to the afforementioned misuse potential.

However, these APIs are occasionally needed for implementing protocols that use special rules for computing hashes (e.g. EIP-712 structured hashes), or for implementing things like network signing services which want to accept a prehash of a message to be signed rather than the full message (to cut down on network bandwidth).

The traits accept a byte slice `prehash`, which permits multiple lengths and allows the implementation to decide which lengths are valid. This makes it possible for e.g. ECDSA implementations to automatically truncate message prehashes which are larger than the field size.

### DIFF
--- a/signature/src/hazmat.rs
+++ b/signature/src/hazmat.rs
@@ -1,0 +1,49 @@
+//! Hazardous Materials: low-level APIs which can be insecure if misused.
+//!
+//! The traits in this module are not generally recommended, and should only be
+//! used in special cases where they are specifically needed.
+//!
+//! Using them incorrectly can introduce security vulnerabilities. Please
+//! carefully read the documentation before attempting to use them.
+//!
+//! To use them, enable the `hazmat-preview` crate feature. Note that this
+//! feature is semi-unstable and not subject to regular 1.x SemVer guarantees.
+//! However, any breaking changes will be accompanied with a minor version bump.
+
+use crate::{Error, Signature};
+
+/// Sign the provided message prehash, returning a digital signature.
+pub trait PrehashSigner<S: Signature> {
+    /// Attempt to sign the given message digest, returning a digital signature
+    /// on success, or an error if something went wrong.
+    ///
+    /// The `prehash` parameter should be the output of a secure cryptographic
+    /// hash function.
+    ///
+    /// This API takes a `prehash` byte slice as there can potentially be many
+    /// compatible lengths for the message digest for a given concrete signature
+    /// algorithm.
+    ///
+    /// Allowed lengths are algorithm-dependent and up to a particular
+    /// implementation to decide.
+    fn try_sign_prehash(&self, prehash: &[u8]) -> Result<S, Error>;
+}
+
+/// Verify the provided message prehash using `Self` (e.g. a public key)
+pub trait PrehashVerifier<S: Signature> {
+    /// Use `Self` to verify that the provided signature for a given message
+    /// `prehash` is authentic.
+    ///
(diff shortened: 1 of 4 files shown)


signature/src/hazmat.rs
```diff
@@ -0,0 +1,49 @@
...
+    /// The `prehash` parameter should be the output of a secure cryptographic
+    /// hash function.
+    ///
+    /// This API takes a `prehash` byte slice as there can potentially be many
+    /// compatible lengths for the message digest for a given concrete signature
+    /// algorithm.
+    ///
+    /// Allowed lengths are algorithm-dependent and up to a particular
+    /// implementation to decide.
+    fn try_sign_prehash(&self, prehash: &[u8]) -> Result<S, Error>;
```
tarcieri, line 29: I thought about a few different alternatives to `prehash: &[u8]` (i.e. a byte slice), including:

- Adding a const generic parameter to the trait, which would permit overlapping impls for different sizes
- Using `generic-array`

While both of those are worth considering, algorithms like DSA and RSASSA-PKCS#1v15/PSS both currently use `num-bigint-dig` with heap-allocated bignums that don't know their own size at a type level.

So I think a slice really is the only option that's general enough.

tarcieri:
It would probably be nice to add custom derive support for `DigestSigner`/`DigestVerifier` to add boilerplate impls which thunk to `PrehashSigner`/`PrehashVerifier`.

signature/src/hazmat.rs
newpavlov: We could introduce two separate methods: array and slice based. The latter would have a default impl in terms of the former. The mentioned algorithm implementations would overwrite the slice-based method and implement the array-based one using slice-based.

tarcieri:
@lumag FYI, if we merge these, you may want to impl them for the RSA signature keys

signature/src/hazmat.rs
tarcieri: The only algorithm that would benefit for now is ECDSA, and that would need `generic-array` so the implementation in the `ecdsa` crate can be generic around field sizes.

However, using the slice-based API means we can implement hash truncation / zero padding in one place, which would make it easier to support some exotic combinations of curve and hash functions, such as ECDSA/P-256 with SHA384, or ECDSA/P-384 with SHA-256.

lumag:
@tarcieri I think I should abstain from implementing it for now. Let's not complicate unreleased dependencies anymore. Not to mention that it adds no direct benefit, as one can use old API to operate on raw (prehashed) data.

tarcieri:
@lumag sure, it'd be great if we could wrap up the `rsa` work and get another release out soon.

Note that the intended use case of these traits is abstracting across signing implementations where the `rsa` crate's software implementation is just one, however I think we can hold off on such support until we start getting requests for it and it's available in other hardware signer crates (e.g. `yubikey` and `yubihsm`)

lumag:
agree

PR #1107, opened 2022-09-15
signature: relax `Keypair` type bounds

### LINKED ISSUE #1106: signature: Keypair not usable for keys not implementing the Signer trait
The `Keypair` trait not usable for keys not implementing the `Signer` trait. For example, I looked upon implementing it for the new RSA key structures. Implementing the `Keypair` for the `pkcs1v15::SigningKey` involves some pain, put should be possible. For the `pss::SigningKey` the `Keypair` can not be implemented at all, since the PSS keys only implement the `RandomizedSigner` and `RandomizedVerified` traits.

lumag:
The Keypair trait might be useful for representing relationship between
DigestSigner and DigestVerifier, RandomizedSigner and
RandomizedVerifier, etc. in addition to just Signer/Verfifier pair.

Relax type boundaries for the Keypair and AsyncKeypair traits.

Fixes #1106

### DIFF
--- a/signature/async/src/lib.rs
+++ b/signature/async/src/lib.rs
@@ -13,7 +13,6 @@ pub use signature::{self, Error, Signature};
 pub use signature::digest::{self, Digest};
 
 use async_trait::async_trait;
-use signature::Verifier;
 
 /// Asynchronously sign the provided message bytestring using `Self`
 /// (e.g. client for a Cloud KMS or HSM), returning a digital signature.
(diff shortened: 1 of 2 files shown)


tarcieri:
This is technically a breaking change, in that downstream consumers that can presently rely on `Keypair::VerifyingKey` to impl `Verifier<S>`.

However, given I just shipped the first impls of this trait yesterday and the day before, I think it's rather unlikely there are any current users of the trait in its current form and we can just yank the associated releases.

PR #1117, opened 2022-09-16
digest: MSRV fix release

newpavlov:


Cargo.toml
```diff
@@ -4,11 +4,13 @@ members = [
...
     "crypto",
-    "crypto-common",
-    "digest",
     "elliptic-curve",
     "kem",
     "signature",
     "signature/async",
     "universal-hash",
 ]
+exclude = [
```
tarcieri, line 13: Aha! So that's how you do this properly

tarcieri (approved):

Cargo.toml
newpavlov: Unfortunately, I am not sure it will work as we would like. Running `cargo +1.41.0 build` in the `digest` folder results in the following warning:
```
warning: unused manifest key: workspace.resolver
```

To be safe I simply will republish `digest` using 1.41 toolchain.

PR #1130, opened 2022-10-03
signature: expand hazmat to cover sign-with-rng case

lumag:
Expand hazmat with the RandomizedPrehashSigner trait, declaring fn sign_prehash_with_rng(). This is necessary for hazmat implementation for the RSA PSS keys.

Signed-off-by: Dmitry Baryshkov <dmitry.baryshkov@linaro.org>


tarcieri (approved):
Looks good, thanks!

PR #1133, opened 2022-10-07
KEM: Updated HPKE deps for testing

rozbb:
Finally release an HPKE v0.10, and fixed broken. This removes the hacky dep that we previously used for KEM testing. Also I alphabetized the deps.

### DIFF
--- a/.github/workflows/kem.yml
+++ b/.github/workflows/kem.yml
@@ -59,7 +59,7 @@ jobs:
     strategy:
       matrix:
         rust:
-          - 1.56.0 # MSRV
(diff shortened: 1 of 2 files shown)


rozbb:
Ope, MSRV on HPKE v0.10 is 1.57. Can't fix that. Feel free to reject this PR, or leave it open for when MSRV gets a bump.

tarcieri:
@rozbb you could bump the version used for the tests.

The `build` step will ensure it still builds on MSRV for the crate.

rozbb:
Alright, should be good to go

tarcieri:
Thanks!

PR #1157, opened 2022-11-19
OSSWU update

daxpedda:
I found some code I had lying around that I never got to upstream because it has some error in it.

When we were working on hash2curve, I was always confused by the algorithm @mikelodder7 was using for OSSWU, because it wasn't the one used in the specification, later when I finally started digging I found that he was using the "original one" from the paper.

Considering we are not implementing the paper but the specification, I would have liked to update it to the specification, which should basically be the same but definitely yield the exact same results: https://www.ietf.org/archive/id/draft-irtf-cfrg-hash-to-curve-16.html#straightline-sswu.

This would also have the advantage of removing the limitation to q = 3 mod 4 curves in the future, as the spec defines a subroutine that works for all: https://www.ietf.org/archive/id/draft-irtf-cfrg-hash-to-curve-16.html#name-sqrt_ratio-for-any-field. The optimized one can be overwritten for when implementing the trait.

For some reason though I never got it to work. It fails the tests in the `p256` crate and I also tried it against the P-384 implementation I did: https://github.com/RustCrypto/elliptic-curves/pull/600.

I would appreciate somebody taking a look so this work doesn't go to waste (or somebody telling me why this isn't a good idea).
Cc @mikelodder7.

### DIFF
--- a/elliptic-curve/src/hash2curve/osswu.rs
+++ b/elliptic-curve/src/hash2curve/osswu.rs
@@ -1,9 +1,11 @@
 //! Optimized simplified Shallue-van de Woestijne-Ulas methods.
 //!
-//! <https://eprint.iacr.org/2009/340.pdf>
+//! <https://www.ietf.org/archive/id/draft-irtf-cfrg-hash-to-curve-16.html#straightline-sswu>
 
 use ff::Field;
 use subtle::Choice;
+use subtle::ConditionallySelectable;
+use subtle::ConstantTimeEq;
 
 /// The Optimized Simplified Shallue-van de Woestijne-Ulas parameters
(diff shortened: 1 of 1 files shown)


mikelodder7:
Thanks for reviewing. I’ll take a look after the week of thanksgiving. I’m swamped until then

daxpedda:
Rebased after https://github.com/RustCrypto/traits/pull/1166.

daxpedda:
Rebased after #1197.

daxpedda:
@tarcieri this suddenly started working now on p256, p384 and k256. So there was a bug that is fixed now.

tarcieri:
Thanks!

PR #1159, opened 2022-11-22
Add HashWriter and HashReader structs

joebebel:
Not sure if this is the right place for this, but we've found frequently needing to `read` or `write` data while simultaneously hashing the data. Based on the implementation in the `phase2` crate, gated by the "std" feature.

While creating this PR I noticed this previous [PR](https://github.com/RustCrypto/hashes/issues/22) which is somewhat related, but I think a little different, and I also note the documentation describes that hashers can implement `io::Write`, which I think is also slightly different.

### DIFF
--- a/digest/src/hashreader.rs
+++ b/digest/src/hashreader.rs
@@ -1,0 +1,114 @@
+//! Adds hashing to any reader
+use super::{Digest, FixedOutputReset, Output, Reset};
+use std::io;
+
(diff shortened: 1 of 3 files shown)


piegamesde:
I've found myself in need for this as well, multiple times. However, I wonder whether some more generic construction like a Writer broadcasting to multiple other Writers (in a separate crate) would be a better solution here.

PR #1172, opened 2022-12-19
crypto-common v0.2

newpavlov:
### Added
- Sealed `BlockSizes` trait implemented for types from `U1` to `U255`

### Changed
- `BlockUser::BlockSize` is now bounded by the `BlockSizes` trait
- Edition changed to 2021 and MSRV bumped to 1.56


crypto-common/Cargo.toml
```diff
@@ -1,11 +1,11 @@
 [package]
 name = "crypto-common"
 description = "Common cryptographic traits"
-version = "0.1.6"
+version = "0.2.0"
```
tarcieri, line 4: FWIW, if this is starting a new release cycle it would be nice to have more time to do more extensive breaking changes, like experiment with a migration to `hybrid-array`.

Perhaps you could bump the version to a prerelease instead?

tarcieri (approved):

PR #1176, opened 2022-12-24
Implement `NonIdentity`

### LINKED ISSUE #1170: Introduce `NonIdentityPoint`
This is similar to `NonZeroScalar`.

I often found myself handrolling custom serde implementations to make sure that a `ProjectivePoint` is not an identity element.

Additionally it could be useful to allow arithmetic that can guarantee some outputs without having to constantly check for identity elements. For example multiplying a `NonIdentityElement` with a `NonZeroScalar` should yield a `NonIdentityElement` without having to do any checks or returning an `(Ct)Option`.

Interactions with `PublicKey` could be similar to whats done between `NonZeroScalar` and `SecretKey`, enabling non-fallible conversions.

Happy to make a PR of course.

daxpedda:
It would be nice to implement math that is generic over both projective and affine points with a common prime marker trait, see https://github.com/zkcrypto/group/issues/43.
I believe in the meantime we could use `Into<ProjectivePoint<C>> where C: PrimeCurve`.

In the future I would like to implement the following math (which I believe are correct):
- Multiplication between `NonIdentity` points.
- Multiplication with `NonZeroScalar`.
- Introduce `fn add_identity(&self) -> NonIdentity`.

Additionally the `PublicKey` type has to be adjusted to return `NonIdentity` when appopriate, which is a breaking change.

On a last point I would like to suggest renaming this to `NonIdentityPoint`, which would be more in line with `NonZeroScalar`.

Fixes #1170.

### DIFF
--- a/elliptic-curve/src/dev.rs
+++ b/elliptic-curve/src/dev.rs
@@ -9,7 +9,7 @@ use crate::{
     ops::{LinearCombination, Reduce},
     pkcs8,
     rand_core::RngCore,
-    sec1::{FromEncodedPoint, ToEncodedPoint},
+    sec1::{CompressedPoint, FromEncodedPoint, ToEncodedPoint},
     subtle::{Choice, ConditionallySelectable, ConstantTimeEq, CtOption},
     zeroize::DefaultIsZeroes,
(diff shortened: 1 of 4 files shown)


tarcieri (approved):
Sorry for the belated review. This looks fine.

PR #1178, opened 2022-12-24
Implement serde for `NonZeroScalar`

daxpedda:
This implements `Serialize` and `Deserialize` for `NonZeroScalar`.

The use-case here is to prevent having to hand-write this for types containing `NonZeroScalar`, which then have to ensure checking for the zero point themselves.


tarcieri:
Thanks!

PR #1179, opened 2022-12-29
Don't unnecessarily initialize a block when unused

CoreyShupe:
There's an unnecessary initializing of a `Block` when it's unused; causing an allocation and immediate de-allocation of the instance when tail is not greater than 0.

This lets us avoid quite a lot of allocations when this cipher is used a lot.


newpavlov:
Note that `Block` does not use any allocations. It resides completely on stack, so in the worst-case scenario we simply needlessly zeroize a bunch of stack bytes, wasting a couple of cycles.

I think the compiler should be smart enough to move `block` initialization into the `if`s, but I guess there is no harm in helping it a bit and making the code a bit less ambiguous.

PR #1186, opened 2023-01-08
elliptic-curve: move `NonIdentity` under `point`

tarcieri:
Reflects that the type represents a non-identity curve point.

cc @daxpedda


daxpedda:
Nice improvement with removing `&self` because of `Copy`.

tarcieri:
@daxpedda `clippy` was complaining about that

daxpedda:
Why did CI pass? I didn't see it locally either.

tarcieri:
I'm a little confused as to why CI passed originally. I don't think there were any interim changes which would've affected the clippy version used in CI.

Issue #641, opened 2021-05-26
Add a `Write` wrapper


piegamesde:
I can pass a hasher where I need a `Write`, and I'll get the digest of the written bytes. I can also have a reader and `std::io::copy` into the hasher to get the digest of all read bytes. In both cases, those bytes go into the void, there is nothing of them left except for the hash.

Sometimes, I want to process data and hash it along the way. So it'd be cool to have `Read` and `Write` wrappers for hashers that can decorate around any IO and give me the hash while passing the data along.

piegamesde:
I've found https://stackoverflow.com/questions/60336419/how-can-i-pipe-bytes-from-a-reader-to-a-writer-with-an-interceptor which has more or less the same problem. The workaround answer suggests to use a `Write` fanout like provided by the `broadcast` crate.

Nevertheless, a direct solution to the problem would be appreciated. I'd be willing to help implement, given some guidance on how an acceptable solution would look like.

newpavlov:
I think the [`digest-io`](https://github.com/RustCrypto/utils/tree/master/digest-io) crate resolves this issue. If anything is missing, feel free to open new PRs against it.

PR #1189, opened 2023-01-13
adding detached tag mode to stream

laudiacay:
Hiiii 

I really wanted detached tags on the streaming encryption mode because I was having to do some really messed up things with buffers that were slowing me down.

I wrote it, it compiles and I'm not sure what tests you'd want... I didn't quite get it working with y'all's macros either but I don't think that made sense bc they were meant to deduplicate code between Encryptor and Decryptor, but it only really makes sense for Encryptors to detach tags... Hope this is cleaned up okay enough, lmk what I should change :)

### DIFF
--- a/aead/src/stream.rs
+++ b/aead/src/stream.rs
@@ -32,7 +32,7 @@
 
 #![allow(clippy::upper_case_acronyms)]
 
-use crate::{AeadCore, AeadInPlace, Buffer, Error, Key, KeyInit, Result};
+use crate::{AeadCore, AeadInPlace, Buffer, Error, Key, KeyInit, Result, Tag};
 use core::ops::{AddAssign, Sub};
 use generic_array::{
     typenum::{Unsigned, U4, U5},
@@ -127,6 +127,15 @@ where
         associated_data: &[u8],
         buffer: &mut dyn Buffer,
     ) -> Result<()>;
+
+    /// Encrypt an AEAD message in-place at the given position in the STREAM.
+    fn encrypt_in_place_detached(
+        &self,
+        position: Self::Counter,
+        last_block: bool,
+        associated_data: &[u8],
+        buffer: &mut dyn Buffer,
+    ) -> Result<Tag<A>>;
 
     /// Encrypt the given plaintext payload, and return the resulting
     /// ciphertext as a vector of bytes.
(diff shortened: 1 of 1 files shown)


tarcieri:
This is OK but it should probably also have the corresponding detached methods on `Decryptor` too

laudiacay:
added- let's see if tests pass now...

tarcieri:
This looks good now, however it's a breaking change, so it will have to be part of the next release cycle (which we'll likely be starting soon)

laudiacay:
Cool- is there anything else I can do around RustCrypto that might be good
for a beginner to this codebase? Always looking for weekend/downtime fun
code besides, like, project euler.

laudiacay:
I also wrote a silly little Writer impl for Encryptor:

can obviously be improved in multiple places (generalizing over ciphers, for one, haha), but it works

here's how i used it to make my code a little nicer...
```
    let mut gzencoder = GzEncoder::new(encrypted_writer, Compression::default());
...
        gzencoder.write_all(&buf_for_input[..n])?;
        gzencoder.flush()?;
...
    let encryptor = gzencoder.finish()?;
    let bytes_written = encryptor.finish()?;
```

and here's the code:
```
pub struct EncryptionWriter<W: Write> {
    buf: Vec<u8>,
    writer: W,
    encryptor: Encryptor<Aes256Gcm, StreamBE32<Aes256Gcm>>,
    bytes_written: usize,
}

impl<W: Write> EncryptionWriter<W> {
    pub fn new(writer: W, key: &[u8]) -> Self {
        let mut nonce = [0u8; 12];
        OsRng.fill_bytes(&mut nonce);
        let cipher = Aes256Gcm::new(key.as_ref().into());
        let encryptor = StreamBE32::from_aead(cipher, nonce.as_ref().into()).encryptor();
        Self {
            buf: Vec::new(),
            writer,
            encryptor,
            bytes_written: 0,
        }
    }

    fn finish(mut self) -> Result<usize> {
        self.flush()?;
        self.encryptor
            .encrypt_last_in_place(b"".as_ref(), &mut self.buf)
            .unwrap();
        self.writer.write_all(&self.buf)?;
        self.bytes_written += self.buf.len();
        Ok(self.bytes_written)
    }
}

impl<W: Write> Write for EncryptionWriter<W> {
    // TODO better buffering? bufwriter?
    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
        self.buf.extend_from_slice(buf);
        Ok(buf.len())
    }

    fn flush(&mut self) -> std::io::Result<()> {
        self.encryptor
            .encrypt_next_in_place(b"", &mut self.buf)
            .unwrap();
        self.writer.write_all(&self.buf)?;
        self.writer.flush()?;
        self.bytes_written += self.buf.len();
        self.buf.clear();
        Ok(())
    }
}
```

laudiacay:
sorta like #641 but not for hashes

tarcieri:
@laudiacay yeah, the buffering is what makes that tricky, but also without either fixed size chunks or some kind of framing protocol you can't write a corresponding decryptor because you don't know the boundaries of the AEAD messages (I personally prefer the fixed-sized chunks approach because it has no size overhead beyond the tags)

If you're looking for other small things to work on, we could use some help here: https://github.com/RustCrypto/hashes/issues/87

kbknapp:
What is the status of this PR? I have a use case where I need to use a `&mut [u8]` instead of `&mut dyn Buffer` (like the standard [`AeadInPlace::*_detached`](https://docs.rs/aead/latest/aead/trait.AeadInPlace.html#tymethod.encrypt_in_place_detached) methods. Would changing this PR to match that signature be a blocker? 

Also re:

> This looks good now, however it's a breaking change, so it will have to be part of the next release cycle (which we'll likely be starting soon)

Not sure if that has already come and gone, or was delayed. So this may be a consideration as well.

tarcieri:
Still waiting on the next breaking release cycle, which will come after #1174

tarcieri:
`aead-stream` has been moved to: https://github.com/RustCrypto/AEADs/tree/master/aead-stream

PR #1198, opened 2023-01-14
elliptic-curve: add `MulByGenerator` trait

tarcieri:
Adds a trait for performing scalar multiplication by the generator point, which may use optimizations (e.g. precomputed tables) when available


tarcieri:
Note: there's a PR open to add a similar feature to the `Group` crate: https://github.com/zkcrypto/group/pull/44/files

This is a stopgap until that gets merged.

PR #1217, opened 2023-01-25
Expand `NonIdentity` usage

daxpedda:
Follow-up to #1176.
IANAC, as far as I understand multiplying a non-zero-scalar with a non-identity-point can never yield the identity-point.

Also realized that I forgot to make the type public :sweat_smile:.

### DIFF
--- a/elliptic-curve/src/lib.rs
+++ b/elliptic-curve/src/lib.rs
@@ -113,6 +113,7 @@ pub use zeroize;
 pub use {
     crate::{
         arithmetic::{CurveArithmetic, PrimeCurveArithmetic},
+        point::NonIdentity,
         public_key::PublicKey,
         scalar::{NonZeroScalar, Scalar},
     },
--- a/elliptic-curve/src/lib.rs
+++ b/elliptic-curve/src/lib.rs
@@ -69,6 +69,7 @@ extern crate alloc;
 extern crate std;
 
 pub mod ops;
+pub mod point;
 pub mod scalar;
 
 #[cfg(feature = "dev")]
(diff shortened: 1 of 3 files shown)


elliptic-curve/src/lib.rs
```diff
@@ -113,6 +113,7 @@ pub use zeroize;
 pub use {
     crate::{
         arithmetic::{CurveArithmetic, PrimeCurveArithmetic},
+        point::NonIdentity,
```
tarcieri, line 116: FYI, it's already re-exported, just as `elliptic_curve::point::NonIdentity` (which I did in #1186): https://github.com/RustCrypto/traits/blob/master/elliptic-curve/src/point.rs#L7

It can potentially be re-exported in both places, although I think `point::NonIdentity` is a bit clearer personally.

daxpedda: Ah, I forgot about #1186.
So the problem is that `mod point` isn't public, so it's still not re-exported:
https://github.com/RustCrypto/traits/blob/778e6f4368422f87ba5d667095673423f8abccda/elliptic-curve/src/lib.rs#L86
I'm happy to remove the new re-export and change `mod point` to `pub mod point`.

tarcieri: Oh whoops. Yeah, I think it's about time we consider making the `point` module public.

daxpedda: Done, let me know if you want me to adjust anything else.

tarcieri (approved):

PR #1175, opened 2022-12-20
Update VOPRF identifier type

daxpedda:
See https://github.com/cfrg/draft-irtf-cfrg-voprf/pull/371.
We should probably at least wait until this is merged upstream.

Cc @kevinlewi.

PR #1238, opened 2023-02-01
elliptic-curve: Allow multiple `dst`s in the hash2curve API

daxpedda:
After https://github.com/RustCrypto/traits/pull/1175 got merged I attempted to go ahead and update voprf and opaque-ke, but quickly hit the same problem that required https://github.com/RustCrypto/traits/pull/876.

The change to the OPRF ID now being a string instead of a byte made it impossible to generate a fixed size array to hold the DST. So I decided to apply the same solution I did in https://github.com/RustCrypto/traits/pull/876.

There are only two changes:
- All hash2curve APIs now take `&[&[u8]]` instead of `&[u8]` as the `dst` parameter.
- `ExpandMsgXmd` now requires `Default + FixedOutput + HashMarker` instead of `Digest`.

### DIFF
--- a/elliptic-curve/src/hash2curve/group_digest.rs
+++ b/elliptic-curve/src/hash2curve/group_digest.rs
@@ -48,10 +48,10 @@ where
     /// [`ExpandMsgXof`]: crate::hash2curve::ExpandMsgXof
     fn hash_from_bytes<'a, X: ExpandMsg<'a>>(
         msgs: &[&[u8]],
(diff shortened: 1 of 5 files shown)


tarcieri:
@daxpedda if you can update https://github.com/RustCrypto/elliptic-curves that'd be great. It's pulling in  `elliptic-curve` from git again

daxpedda:
On it.

PR #1246, opened 2023-02-15
password-hash: Fix invalid operator in the SaltString struct constructor

DmitrijVC:
I am not 100% sure, but I think the length of provided string `s` should be lower or equal to the `MAX_LENGTH`.

When implementing some algorithm that required 64 bytes long salt I couldn't make a new `SaltString` structure due to this issue  and simple change from `<` to the `<=` operator fixed it. What is more, function `Salt::new` allows for the `s` string to be exactly 64 bytes (`MAX_LENGTH`) long. 

I did not find any opened or closed issues related to this matter, so that's why I am not sure if it is a bug or a feature.

### DIFF
--- a/password-hash/src/salt.rs
+++ b/password-hash/src/salt.rs
@@ -198,7 +198,7 @@ impl SaltString {
 
         let length = s.as_bytes().len();
 
-        if length < Salt::MAX_LENGTH {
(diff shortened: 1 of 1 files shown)


tarcieri (commented):
Thanks!

PR #1261, opened 2023-03-03
cipher: use borrowing in AsyncStreamCipher

valaphee:
AsyncStreamCipher unnecessarily takes ownership, even if its underlying methods use borrowing. And makes using it with for example tokio-util's Codec more complicated, as logically the iv changes during encryption, and I would have to save it, and restore it in the next chunk, etc.

This allows to just use mut borrowing.

### DIFF
--- a/cipher/src/stream.rs
+++ b/cipher/src/stream.rs
@@ -11,7 +11,7 @@ use inout::{InOutBuf, NotEqualError};
 /// Marker trait for block-level asynchronous stream ciphers
 pub trait AsyncStreamCipher: Sized {
     /// Encrypt data using `InOutBuf`.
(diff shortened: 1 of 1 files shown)


newpavlov:
This change not only a breaking one, but also incorrect. When we use byte-level encrypt/decrypt methods  of an async cipher we forget part of the final XORing block. So if you are to reuse such instance, it will appear as if the latter part of the message was padded to a multiple of block size.

What you need is a buffered version of encryptor/decryptor, we have them implemented in the `cfb-mode` crate.

valaphee:
Thanks, didn't know that there are also buffered stream ciphers, and yep would only work for cfb8

PR #1264, opened 2023-03-04
password-hash: provide std::error::Error::source() for Error

z33ky:
There's not much in the way of description here, I hope the change is self-explanatory.


tarcieri:
Thanks!

PR #1282, opened 2023-03-31
Add a function to generate a nonce prefix for stream primitives.

jcdyer:
Currently, the way to generate a nonce for an `impl StreamPrimitive<A>` is not obvious unless you understand how typenums work.  This is an attempt to make it analogous to how `AeadCore::generate_nonce` works.


jcdyer:
Then I'll close this, and suggest that we consider providing better documentation for how to set up a stream, preferably with example code.

PR #1287, opened 2023-04-03
elliptic-curve: fix FieldBytesEncoding ranges

wangwillian0:
This should fix methods like SecretKey::from_slice for curves where the number of bytes in their internal representation is larger than it's modulus (e.g. NIST P224)


tarcieri:
Thank you!

PR #1310, opened 2023-05-18
elliptic-curve: faster `PublicKey::from_encoded_point`

tarcieri:
Checks if the SEC1 point is using the identity encoding directly, rather than round tripping through `ProjectivePoint`.


wangwillian0:
Just tested and the runtime 200k ECDH operations with P224 went from 38s to 35s, a performance jump of ~8% as discussed at Zulip 😀

PR #1315, opened 2023-05-26
fix: `missing_debug_implementations`

JonathanWoollett-Light:
Sets `missing_debug_implementations` to warn and adds missing `std::fmt::Debug` implementations.

When tracing/logging/debugging a program it is common to debug print inputs and outputs of function, to do this generally dependencies need to offer `std::fmt::Debug` implementations.

I won't repeat what is written under [`missing_debug_implementations`](https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#missing-debug-implementations) which gives some more explanation.

### DIFF
--- a/cipher/src/block.rs
+++ b/cipher/src/block.rs
@@ -134,9 +134,9 @@ pub trait BlockEncrypt: BlockSizeUser + Sized {
     #[cfg(feature = "block-padding")]
     #[cfg_attr(docsrs, doc(cfg(feature = "block-padding")))]
     #[inline]
-    fn encrypt_padded_inout<'inp, 'out, P: Padding<Self::BlockSize>>(
+    fn encrypt_padded_inout<'out, P: Padding<Self::BlockSize>>(
         &self,
-        data: InOutBufReserved<'inp, 'out, u8>,
+        data: InOutBufReserved<'_, 'out, u8>,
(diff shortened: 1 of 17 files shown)


cipher/src/block.rs
```diff
@@ -250,9 +250,9 @@ pub trait BlockDecrypt: BlockSizeUser {
     #[cfg(feature = "block-padding")]
     #[cfg_attr(docsrs, doc(cfg(feature = "block-padding")))]
     #[inline]
-    fn decrypt_padded_inout<'inp, 'out, P: Padding<Self::BlockSize>>(
+    fn decrypt_padded_inout<'out, P: Padding<Self::BlockSize>>(
```
newpavlov, line 253: Technically, removing lifetime argument is a breaking change.

newpavlov (changes requested):
Similarly to RustCrypto/block-modes/pull/46, you should not derive `Debug` for types which contain sensitive information.

cipher/src/block.rs
JonathanWoollett-Light: Okay, do I need to take any action for this?

newpavlov:
This PR got outdated, so I will close it in favor of the [tracking issue](https://github.com/RustCrypto/traits/issues/1389).

PR #1319, opened 2023-05-30
[WIP] crypto-common: migrate to `hybrid-array`; MSRV 1.65

tarcieri:

`hybrid-array` is designed to be a largely drop-in replacement, and the number of changes required to switch are relatively minimal aside from some idiosyncracies.


crypto-common/src/lib.rs
```diff
@@ -154,7 +154,7 @@ pub trait KeyInit: KeySizeUser + Sized {
         if key.len() != Self::KeySize::to_usize() {
             Err(InvalidLength)
         } else {
-            Ok(Self::new(Key::<Self>::from_slice(key)))
+            Ok(Self::new(Key::<Self>::ref_from_slice(key)))
```
tarcieri, line 157: Note: this method exists to simplify migrating `generic-array`-based code but I'd like to at least deprecate it before the next release.

The code here should be replaced with `TryFrom` which I'll do before removing draft.

crypto-common/src/lib.rs
```diff
@@ -164,7 +164,7 @@ pub trait KeyInit: KeySizeUser + Sized {
     #[inline]
     fn generate_key(mut rng: impl CryptoRng + RngCore) -> Key<Self> {
         let mut key = Key::<Self>::default();
-        rng.fill_bytes(&mut key);
+        rng.fill_bytes(key.as_mut());
```
tarcieri, line 167: It seems `GenericArray` has `Deref`/`DerefMut` impls with a target of `[T]`, whereas `hybrid-array` has a target of `ArraySize::ArrayType`, which is what allowed this to work before without the explicit `as_mut()`.

I think there might be some wisdom to that, and it would simplify updating existing code. Access to the inner array is still available via `AsRef`/`AsMut`, `Borrow`/`BorrowMut`, or `.0` since the inner array field is `pub`.

tarcieri: Updated `hybrid-array` to `Deref` to a slice target

tarcieri:
Since `block-buffer` depends on `crypto-common`, I've opened the next step of this migration in the utils repo: https://github.com/RustCrypto/utils/pull/944

newpavlov:
`generic-array` v1.0 has implemented `Try`/`From` for arrays using const generics and simplified the `ArrayLength` trait. Is this PR still relevant in the light of it?

tarcieri:
Yes, `hybrid-array` is natively built on const generics. `Array` is a newtype of the inner array type, which is `pub`.

`generic-array` instead has an opaque inner type and uses large amounts of unsafe code to achieve its goals.

newpavlov:
Ok. Should we migrate to `generic-array` v1.0 in the next release cycle first or would you prefer to try `hybrid-array` migration without it?

tarcieri:
Personally I'd prefer to go straight to `hybrid-array` and have already done some of the work, see also https://github.com/RustCrypto/utils/pull/944

I can do `cipher` and `digest` next.

tarcieri:
The `minimal-versions` failure seems to be a nightly regression and probably impacts `master` as well

tarcieri:
@newpavlov this is ready now

crypto-common/src/lib.rs
```diff
@@ -59,12 +59,12 @@ impl<T: BlockSizeUser> BlockSizeUser for &mut T {
 }
 
 /// Trait implemented for supported block sizes, i.e. for types from `U1` to `U255`.
-pub trait BlockSizes: ArrayLength<u8> + sealed::BlockSizes + 'static {}
+pub trait BlockSizes: ArraySize + sealed::BlockSizes + 'static {}
```
newpavlov, line 62: I wonder if these `'static` bounds are still needed. I think we introduced them to work around an old issue which may have been already fixed. But I guess we can leave them be.

newpavlov (approved):
Can you also change bound for the `BlockSize` sealed impl from `Self: IsLess<U256> + IsGreater<U1>` to `Self: IsLess<U256> + IsGreater<U0>`? I intended to do it in #1174.

crypto-common/src/lib.rs
tarcieri: Good catch! We don't because `ArraySize` already bounds on `'static`: https://docs.rs/hybrid-array/0.2.0-pre.5/hybrid_array/trait.ArraySize.html

tarcieri: Duplicate `'static` bounds removed in c3ede80

newpavlov (approved):

PR #1346, opened 2023-09-22
cipher: mark set_pos_unchecked private method as unsafe

### LINKED ISSUE #1342: `set_pos_unchecked` should be an `unsafe fn`
https://github.com/RustCrypto/traits/blob/59490b50e88c252d179087072c1559303f8c5872/cipher/src/stream_wrapper.rs#L65-L68

`self.pos` has a safety invariant that `self.pos < T::BlockSize::USIZE`, so `set_pos_unchecked` should be an `unsafe fn`. There is currently no unsoundness - in practice, all calls to `set_pos_unchecked` are sound. This was discovered while [reviewing cipher for  vendoring in Fuchsia](https://fuchsia-review.googlesource.com/c/fuchsia/+/711365/9..13/third_party/rust_crates/vendor/cipher/src/stream_wrapper.rs#b65).

cc @anp

newpavlov:
Closes #1342 

cc @joshlf

### DIFF
--- a/cipher/src/stream_wrapper.rs
+++ b/cipher/src/stream_wrapper.rs
@@ -64,8 +64,13 @@ where
         T::BlockSize::USIZE
     }
 
+    /// Set buffer position without checking that it's smaller
+    /// than buffer size.
+    ///
+    /// # Safety
+    /// `pos` MUST be smaller than `T::BlockSize::USIZE`.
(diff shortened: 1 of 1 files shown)


joshlf:
Oh awesome! Sorry I didn't get around to implementing this myself before now.

PR #1352, opened 2023-10-01
password-hash: add `Base64ShaCrypt`

oganexon:
`base64ct` recent versions introduces `Base64ShaCrypt`, a `Crypt` variant for sha1_crypt, sha256_crypt, sha512_crypt, and md5_crypt.

I added this variant in the Encoding of `password-hash`.
```rust
pub enum Encoding {
    B64,
    Bcrypt,
    Crypt,
    ShaCrypt,
}
```

> This change is necessary to support these algorithms. (incoming PR for these once this is merged)

I only updated `password-hash`'s Cargo.toml file and not the global `Cargo.lock` since there were a lot of dependencies on it. I didn't want this PR to be gigantic.


tarcieri:
This is unfortunately a breaking change as `Encoding` is not marked `#[non_exhaustive]`.

We can add it in the next breaking release of `password-hash`.

oganexon:
I agree with the addition of `#[non_exhaustive]` 👍

PR #1358, opened 2023-10-08
cipher+digest: migrate to hybrid array; MSRV 1.65

tarcieri:
Continuation of #1319.    

`hybrid-array` is designed to be a largely drop-in replacement, and the number of changes required to switch are relatively minimal aside from some idiosyncrasies.

### DIFF
--- a/cipher/src/stream_core.rs
+++ b/cipher/src/stream_core.rs
@@ -1,8 +1,8 @@
 use crate::{ParBlocks, ParBlocksSizeUser, StreamCipherError};
 use crypto_common::{
-    generic_array::{ArrayLength, GenericArray},
+    array::{Array, ArraySize},
     typenum::Unsigned,
-    Block, BlockSizeUser,
+    Block, BlockSizeUser, BlockSizes,
 };
 use inout::{InOut, InOutBuf};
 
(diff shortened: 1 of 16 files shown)


tarcieri:
@newpavlov well, there's the next step

digest/src/digest.rs
```diff
@@ -196,21 +196,15 @@ impl<D: Update + FixedOutputReset + Reset + Clone + 'static> DynDigest for D {
...
     fn finalize_into(self, buf: &mut [u8]) -> Result<(), InvalidBufferSize> {
-        if buf.len() == self.output_size() {
-            FixedOutput::finalize_into(self, Output::<Self>::from_mut_slice(buf));
-            Ok(())
-        } else {
-            Err(InvalidBufferSize)
-        }
+        let buf = <&mut Output<Self>>::try_from(buf).map_err(|_| InvalidBufferSize)?;
+        FixedOutput::finalize_into(self, buf);
+        Ok(())
```
newpavlov, line 201: I think it's better to write it like this:
```rust
let buf: &mut Output<Self> = buf.try_into().map_err(|_| InvalidBufferSize)?;
```
I think you can even skip the type ascription, since compiler should be able to determine the type itself.

Another alternative:
```rust
buf
    .try_into()
    .map_err(|_| InvalidBufferSize)
    .map(|buf| FixedOutput::finalize_into(self, buf))
```

cipher/src/stream_core.rs
```diff
@@ -195,7 +195,7 @@ impl_counter! { u32 u64 u128 }
 /// In case if `N` is less or equal to 1, buffer of arrays has length
 /// of zero and tail is equal to `self`.
 #[inline]
-fn into_chunks<T, N: ArrayLength<T>>(buf: &mut [T]) -> (&mut [GenericArray<T, N>], &mut [T]) {
+fn into_chunks<T, N: ArraySize>(buf: &mut [T]) -> (&mut [Array<T, N>], &mut [T]) {
```
newpavlov, line 198: Maybe we should move this method into `hybrid-array`? I think we had a recent discussion about it.

digest/src/digest.rs
tarcieri: Applied the latter suggestion in 6a2190a

cipher/src/stream_core.rs
tarcieri: Sure, although perhaps that could be a followup?

PR #1363, opened 2023-10-18
signature: add traits to obtain verifier objects

fjarri:
I occasionally encounter situations where I need to get a verifier from a signer, and have to pass both to a method since there's currently no generic mechanism to do that. `ecdsa::SigningKey` has a [`verifying_key()`](https://docs.rs/ecdsa/0.16.8/ecdsa/struct.SigningKey.html#method.verifying_key) method, but it's not in any trait, so if my code is generic over Signer/Verifier traits, I can't use it.

Would it make sense to add a trait for that connection? Are there any situations where a signer object does not allow one to obtain the corresponding verifier? If not, we can even bound `Singer<S>: HasVerifier<S>`.

### DIFF
--- a/signature/src/signer.rs
+++ b/signature/src/signer.rs
@@ -1,12 +1,32 @@
 //! Traits for generating digital signatures
 
 use crate::error::Error;
+use crate::verifier::Verifier;
 
 #[cfg(feature = "digest")]
-use crate::digest::Digest;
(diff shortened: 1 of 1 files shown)


fjarri:
My bad, I think that will cover my use cases. Closing the PR.

PR #1365, opened 2023-10-24
Committing AEAD marker traits and generic constructions

rlee287:
See the documentation comments for an explanation of committing AE and the generic constructions included.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,8 +10,10 @@ dependencies = [
  "blobby",
  "bytes",
  "crypto-common 0.1.6",
+ "digest 0.10.7",
  "generic-array",
  "heapless",
+ "hmac 0.12.1",
  "subtle",
 ]
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12,6 +12,7 @@ dependencies = [
  "crypto-common 0.1.6",
  "generic-array",
  "heapless",
+ "subtle",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


rlee287:
Seems like the test failures are caused by the cipher crate emitting a warning in the minimal version build, as opposed to being related to any of the changes I made.

newpavlov:
I think it will be better to move the wrapper into a different crate in the AEADs repository and link it in the new wrappers trait docs.

tarcieri:
Yeah, I had a similar first thought: this might make sense in a separate crate.

If it were located in the https://github.com/RustCrypto/AEADs/ repo it would be much easier to test the generic implementation against a concrete AEAD algorithm implementation located there.

rlee287:
I'll move the wrappers into a separate PR to the AEADs repo, but I'll also take the time to add some tests using concrete AEADs. However, I'd like to confirm whether you think it'd be better to keep the marker traits in the `aead` crate of this repo or whether I should also move the traits into the separate crate as well.

rlee287:
I opened a PR against the AEADs repo at https://github.com/RustCrypto/AEADs/pull/564, so I'll close this one. I will send in a PR to cross-reference the new crate in the `aead` crate documentation after that one is merged.

tarcieri:
Sounds good. Since committing AEADs can be completely implemented in that crate, it seems like there isn't much value in having upstream marker traits?

We're going to start making breaking changes to `aead` soon, so if there's anything you do need changed, get a PR in ASAP

PR #1368, opened 2023-10-27
cipher: remove needless lifetimes

### LINKED ISSUE #1336: cipher: resolve `clippy::needless_lifetimes` lint
Right now we have methods like this:
```rust
fn encrypt_padded_inout<'inp, 'out, P: Padding<Self::BlockSize>>(
    &self,
    data: InOutBufReserved<'inp, 'out, u8>,
) -> Result<&'out [u8], PadError> { .. }
```
The Clippy lint suggests to write it like this:
```rust
fn encrypt_padded_inout<'out, P: Padding<Self::BlockSize>>(
    &self,
    data: InOutBufReserved<'_, 'out, u8>,
) -> Result<&'out [u8], PadError> { .. }
```
In my understanding, strictly speaking, it's a breaking change, so it should be done as part of cipher v0.5.

tarcieri:
One of the lifetimes is unused, but we couldn't remove it before because it would be a breaking change.

Closes #1336

### DIFF
--- a/cipher/src/block.rs
+++ b/cipher/src/block.rs
@@ -135,9 +135,9 @@ pub trait BlockEncrypt: BlockSizeUser + Sized {
     #[cfg(feature = "block-padding")]
     #[cfg_attr(docsrs, doc(cfg(feature = "block-padding")))]
     #[inline]
-    fn encrypt_padded_inout<'inp, 'out, P: Padding<Self::BlockSize>>(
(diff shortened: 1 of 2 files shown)


newpavlov (commented):
You have missed a bunch of other methods with this issue. I guess Clippy should highlight them, if not, then its version should be bumped.

newpavlov (approved):

PR #1361, opened 2023-10-16
crypto-common: add SerializableState trait

rpiasetskyi:
Trait for saving the internal state of the object and restoring it later.

Original PR: https://github.com/RustCrypto/traits/pull/1078

PR #1369, opened 2023-10-27
crypto-common: add SerializableState trait

rpiasetskyi:
Trait for saving the internal state of the object and restoring it later.

Original PRs: https://github.com/RustCrypto/traits/pull/1078 https://github.com/RustCrypto/traits/pull/1361

Original issue: https://github.com/RustCrypto/hashes/issues/310

### DIFF
--- a/digest/src/core_api/rt_variable.rs
+++ b/digest/src/core_api/rt_variable.rs
@@ -4,8 +4,17 @@ use crate::MacMarker;
 use crate::{HashMarker, InvalidBufferSize};
 use crate::{InvalidOutputSize, Reset, Update, VariableOutput, VariableOutputReset};
 use block_buffer::BlockBuffer;
-use core::fmt;
-use crypto_common::typenum::Unsigned;
+use core::{
+    convert::TryInto,
+    fmt,
+    ops::{Add, Sub},
+};
+use crypto_common::SubSerializedStateSize;
+use crypto_common::{
+    array::{ArraySize, ByteArray},
+    typenum::{Diff, IsLess, Le, NonZero, Sum, Unsigned, U1, U256},
+    AddBlockSize, DeserializeStateError, SerializableState, SerializedState, SubBlockSize,
+};
 
 /// Wrapper around [`VariableOutputCore`] which selects output size
 /// at run time.
@@ -115,6 +124,60 @@ where
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
         T::write_alg_name(f)?;
         f.write_str(" { .. }")
+    }
+}
+
+type RtVariableCoreWrapperSerializedStateSize<T> =
+    Sum<AddBlockSize<Sum<<T as SerializableState>::SerializedStateSize, U1>, T>, U1>;
+
+impl<T> SerializableState for RtVariableCoreWrapper<T>
+where
+    T: VariableOutputCore + UpdateCore + SerializableState,
+    T::BlockSize: IsLess<U256>,
+    Le<T::BlockSize, U256>: NonZero,
+    T::SerializedStateSize: Add<U1>,
+    Sum<T::SerializedStateSize, U1>: Add<T::BlockSize> + ArraySize,
+    AddBlockSize<Sum<T::SerializedStateSize, U1>, T>: Add<U1> + ArraySize,
+    RtVariableCoreWrapperSerializedStateSize<T>: Sub<T::SerializedStateSize> + ArraySize,
+    SubSerializedStateSize<RtVariableCoreWrapperSerializedStateSize<T>, T>: Sub<U1> + ArraySize,
+    Diff<SubSerializedStateSize<RtVariableCoreWrapperSerializedStateSize<T>, T>, U1>:
+        Sub<T::BlockSize> + ArraySize,
+    SubBlockSize<
+        Diff<SubSerializedStateSize<RtVariableCoreWrapperSerializedStateSize<T>, T>, U1>,
+        T,
+    >: ArraySize,
+{
+    type SerializedStateSize = RtVariableCoreWrapperSerializedStateSize<T>;
+
+    fn serialize(&self) -> SerializedState<Self> {
+        let serialized_core = self.core.serialize();
+        let serialized_pos =
+            ByteArray::<U1>::clone_from_slice(&[self.buffer.get_pos().try_into().unwrap()]);
+        let serialized_data = self.buffer.clone().pad_with_zeros();
+        let serialized_output_size =
+            ByteArray::<U1>::clone_from_slice(&[self.output_size.try_into().unwrap()]);
+
+        serialized_core
+            .concat(serialized_pos)
+            .concat(serialized_data)
+            .concat(serialized_output_size)
+    }
+
+    fn deserialize(
+        serialized_state: &SerializedState<Self>,
+    ) -> Result<Self, DeserializeStateError> {
+        let (serialized_core, remaining_buffer) =
+            serialized_state.split_ref::<T::SerializedStateSize>();
+        let (serialized_pos, remaining_buffer) = remaining_buffer.split_ref::<U1>();
+        let (serialized_data, serialized_output_size) =
+            remaining_buffer.split_ref::<T::BlockSize>();
+
+        Ok(Self {
+            core: T::deserialize(serialized_core)?,
+            buffer: BlockBuffer::try_new(&serialized_data[..serialized_pos[0].into()])
+                .map_err(|_| DeserializeStateError)?,
+            output_size: serialized_output_size[0].into(),
+        })
     }
 }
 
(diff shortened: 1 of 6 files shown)


crypto-common/serializable-state-derive/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "serializable-state-derive"
```
tarcieri, line 2: It would probably make more sense to call this `crypto-common-derive`, and make it an optional dependency so we don't have hard dependencies on a proc macro stack.

However, for review purposes I would suggest providing it as a followup PR.

newpavlov:
I am not sure about the derive crate. It adds a significant amount of code and could be dangerous in respect to serialization output stability. I think we can afford to write serialization by hand on a case-by-case basis?

tarcieri:
@rpiasetskyi could you remove the custom derive for now so we can just focus on the core traits? That would help make it easier to review.

rpiasetskyi:
@tarcieri sorry for the delay.

@newpavlov I would still recommend using it in most cases because it helps to avoid unnecessary code duplication and makes it easy and safe. It works as `serde`: serializes members one by one if they implement `SerializableState` trait.

But some for structures it might be better to implement manually.

> could be dangerous in respect to serialization output stability
Could you please explain your concerns?

digest/src/core_api/rt_variable.rs
```diff
@@ -118,6 +135,67 @@ where
...
+            Diff<
+                Sum<Sum<Sum<T::SerializedStateSize, U1>, T::BlockSize>, U1>,
+                T::SerializedStateSize,
+            >,
+            U1,
+        >,
+        T::BlockSize,
+    >: ArraySize,
+{
+    type SerializedStateSize = Sum<Sum<Sum<T::SerializedStateSize, U1>, T::BlockSize>, U1>;
```
tarcieri, line 165: Curious if e.g. the `Sum`-computed types might be clearer with type aliases, similar to something like https://github.com/RustCrypto/signatures/blob/7bf1313/ecdsa/src/der.rs#L45

newpavlov:
>Could you please explain your concerns?

Reordering of fields or changing types (e.g from `[u32; 2]` to `u64`) can break output compatibility of derived serialization. Also, derive macros have a noticeable negative impact on compilation times, so if we are to implement the traits by default, I would prefer to have manual implementations.

digest/src/core_api/rt_variable.rs
rpiasetskyi:
```
type Sum1<T> = Sum<T, U1>;
type Diff1<T> = Diff<T, U1>;
type SumBlockSize<T, B> = Sum<T, <B as BlockSizeUser>::BlockSize>;
type DiffBlockSize<T, B> = Diff<T, <B as BlockSizeUser>::BlockSize>;
type RtVariableCoreWrapperSerializedStateSize<T> =
    Sum1<SumBlockSize<Sum1<<T as SerializableState>::SerializedStateSize>, T>>;

impl<T> SerializableState for RtVariableCoreWrapper<T>
where
    T: VariableOutputCore + UpdateCore + SerializableState,
    T::BlockSize: IsLess<U256>,
    Le<T::BlockSize, U256>: NonZero,
    T::SerializedStateSize: Add<U1>,
    Sum1<T::SerializedStateSize>: Add<T::BlockSize> + ArraySize,
    SumBlockSize<Sum1<T::SerializedStateSize>, T>: Add<U1> + ArraySize,
    RtVariableCoreWrapperSerializedStateSize<T>: Sub<T::SerializedStateSize> + ArraySize,
    Diff<RtVariableCoreWrapperSerializedStateSize<T>, T::SerializedStateSize>: Sub<U1> + ArraySize,
    Diff1<Diff<RtVariableCoreWrapperSerializedStateSize<T>, T::SerializedStateSize>>:
        Sub<T::BlockSize> + ArraySize,
    DiffBlockSize<
        Diff1<Diff<RtVariableCoreWrapperSerializedStateSize<T>, T::SerializedStateSize>>,
        T,
    >: ArraySize,
{
    type SerializedStateSize = RtVariableCoreWrapperSerializedStateSize<T>;
```

What do you think about this one?
But `Sum1`, `Diff1`, `SumBlockSize` and `DiffBlockSize` should go to the `typenum` crate. Or to the `crypto-common`.
Also, something like `SumSerializedState<T, S>` and `DiffSerializedState<T, S>` is possible.

tarcieri: Not sure `Sum1` and `Diff1` really make anything clearer, but the others look good

rpiasetskyi: Done.

tarcieri (approved):

tarcieri:
Going to merge this as it's a frequently requested feature.

Since `crypto-common` is currently a prerelease anyway, there's time to iterate on the API before a final release.

burdges:
Is this hazmat?  It's well beyond the typical analysis of course.

Inherent methods might not be dangerous for specific hash functions, but in general you'd expect serializing the internal state breqaks assumptions when used, or even permits collisions.  At least the warning should read stronger:
```
Analysis of cryptographic primitives rarely if ever consideres serialized state, so employing serialized state would typically break the security of the underlying primitive:  Serialized state would tyically leak sensitive data.  A digest or cipher loses its randomness or collisions resistance properties.

We advise that serialized state only be used in non-advarsarial layers of the protocol.
```

tarcieri:
I'd say there's quite a bit of misuse potential. It could definitely use some warnings.

We could also gate implementations of the trait under a `hazmat` feature in each crate (though thanks to feature unification, such features are easily flipped on by other dependencies)

wiktor-k:
Hi,

As I'm interested in this feature I started testing the `pre` versions.

Are there any examples on how to use `SerializableState` with concrete crates?

I wrote this small example using `sha2 = "0.11.0-pre.2"`:

```rust
fn main() {
    use sha2::digest::crypto_common::SerializableState;
    use sha2::Digest;
    let mut hasher = sha2::Sha256::new();
    hasher.update(b"hello ");

    let result = hasher.serialize();
}
```

And the compilation error looks pretty opaque to me:

```
error[E0599]: the method `serialize` exists for struct `CoreWrapper<CtVariableCoreWrapper<Sha256VarCore, UInt<UInt<UInt<..., ...>, ...>, ...>, ...>>`, but its trait bounds were not satisfied
  --> src/main.rs:7:25
   |
7  |     let result = hasher.serialize();
   |                         ^^^^^^^^^ method cannot be called due to unsatisfied trait bounds
   |
  ::: /home/wiktor/.cargo/registry/src/index.crates.io-6f17d22bba15001f/digest-0.11.0-pre.7/src/core_api/ct_variable.rs:30:1
   |
30 | pub struct CtVariableCoreWrapper<T, OutSize, O = NoOid>
   | ------------------------------------------------------- doesn't satisfy `_: SerializableState`
   |
  ::: /home/wiktor/.cargo/registry/src/index.crates.io-6f17d22bba15001f/digest-0.11.0-pre.7/src/core_api/wrapper.rs:30:1
   |
30 | pub struct CoreWrapper<T: BufferKindUser> {
   | ----------------------------------------- doesn't satisfy `_: SerializableState`
   |
  ::: /home/wiktor/.cargo/registry/src/index.crates.io-6f17d22bba15001f/sha2-0.11.0-pre.2/src/core_api.rs:22:1
   |
22 | pub struct Sha256VarCore {
   | ------------------------ doesn't satisfy `Sha256VarCore: SerializableState`
   |
   = note: the full type name has been written to '/home/wiktor/tmp/cargo/debug/deps/hasher_test-d2968b8c355f072e.long-type-1048706664379593364.txt'
   = note: the following trait bounds were not satisfied:
           `CtVariableCoreWrapper<Sha256VarCore, UInt<UInt<UInt<UInt<UInt<UInt<UTerm, B1>, B0>, B0>, B0>, B0>, B0>, OidSha256>: SerializableState`
           which is required by `CoreWrapper<CtVariableCoreWrapper<Sha256VarCore, UInt<UInt<UInt<UInt<UInt<UInt<UTerm, B1>, B0>, B0>, B0>, B0>, B0>, OidSha256>>: SerializableState`
           `Sha256VarCore: SerializableState`
           which is required by `CoreWrapper<CtVariableCoreWrapper<Sha256VarCore, UInt<UInt<UInt<UInt<UInt<UInt<UTerm, B1>, B0>, B0>, B0>, B0>, B0>, OidSha256>>: SerializableState`
```

Thanks for help! :wave:

burdges:
An option, you could give either the trait or its methods scary names, like `InsecureSerializableState`.

tarcieri:
@burdges for other things like this we've put it in a `hazmat` module with appropriate module-level documentation, which seems like it would work here as well.

We should probably also add documentation about potential misuses to each of the traits as well.

tarcieri:
Here's a PR to place all of the traits inside a `hazmat` module: https://github.com/RustCrypto/traits/pull/1487

PR #1371, opened 2023-10-31
crypto-common: change `generate_*` to support getrandom

tarcieri:
This commit splits the existing `generate_*` functions into two different variants:

- `fn generate_*`: accepts no parameters and uses `OsRng` automatically. Gated on the `getrandom` feature.
- `fn generate_*_with_rng`: accepts a `&mut impl CryptoRngCore` parameter. Gated on the `rand_core`.

Previously all of the `generate_*` methods were parameterized on a `CryptoRngCore`.

Splitting them up like this makes it very easy for users to do the right thing, which is use `getrandom`/`OsRng`, but without the need to document how to import `OsRng` into their code everywhere.

Retaining the parameterized versions as `*_with_rng` allows users to pass a custom RNG where it makes sense, for example `rand::thread_rng`, or potentially an embedded peripheral/entropy pool.

### DIFF
--- a/crypto-common/src/lib.rs
+++ b/crypto-common/src/lib.rs
@@ -20,8 +20,11 @@ pub use hybrid_array::typenum;
 
 use core::fmt;
 use hybrid_array::{typenum::Unsigned, Array, ArraySize, ByteArray};
+
 #[cfg(feature = "rand_core")]
 use rand_core::CryptoRngCore;
+#[cfg(feature = "getrandom")]
+use rand_core::OsRng;
 
 /// Block on which [`BlockSizeUser`] implementors operate.
 pub type Block<B> = ByteArray<<B as BlockSizeUser>::BlockSize>;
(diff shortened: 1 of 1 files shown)


newpavlov (approved):

crypto-common/src/lib.rs
```diff
@@ -156,10 +159,17 @@ pub trait KeyInit: KeySizeUser + Sized {
...
+        Self::generate_key_with_rng(&mut OsRng)
+    }
+
     /// Generate random key using the provided [`CryptoRngCore`].
     #[cfg(feature = "rand_core")]
     #[inline]
-    fn generate_key(mut rng: impl CryptoRngCore) -> Key<Self> {
+    fn generate_key_with_rng(rng: &mut impl CryptoRngCore) -> Key<Self> {
         let mut key = Key::<Self>::default();
         rng.fill_bytes(&mut key);
```
tarcieri, line 174: @newpavlov should we also make these methods fallible and call `try_fill_bytes`?

newpavlov: It may be worth to add separate `try_*` variants and add a note to the `generate_*` method docs about potential (improbable) panics.

Issue #1373, opened 2023-11-08
`async-signature` on executors expecting `!Send` futures


QuinnWilton:
Hey there! I know it's just an experimental API, but I was hoping to make use of the [AsyncSigner](https://docs.rs/async-signature/latest/async_signature/trait.AsyncSigner.html) trait as part of a `wasm-bindgen` project, to support signing payloads using non-extractable `WebCrypto` keys.

I ran into two issues when trying that:
1) The `Send + Sync` bounds on `Self` aren't satisfied by the types in `web-sys`, like `CryptoKeyPair`, which are explicitly `!Send`
2) The `Send` bound on the future itself isn't satisfied by the `!Send` futures used when invoking `SubtleCrypto.sign`

Both of these problems arise because `JsValue` is `!Send` in `wasm-bindgen`, and so it's expected that all futures will also be `!Send`, unlike in more traditional runtimes like `tokio`.

I'm able to kludge around the first problem by manually implementing `Send` for my type that'll implement `AsyncSigner<S>`, which should be fine assuming web-workers aren't being used, but I'm not sure how to get around the second problem, and I suspect that an upstream change will be needed for that.

Do you have any ideas?

This is roughly what I'm trying to do:

```rust
struct JsSigner<K> {
    key_pair: CryptoKeyPair,
    _marker: std::marker::PhantomData<fn() -> K>,
}

#[async_trait]
impl AsyncSigner<rsa::pkcs1v15::Signature> for JsSigner<rsa::pkcs1v15::Signature> {
    async fn sign_async(
        &self,
        msg: &[u8],
    ) -> Result<rsa::pkcs1v15::Signature, async_signature::Error> {
        let subtle = Self::subtle_crypto().map_err(|e| async_signature::Error::from_source(e))?;

        // This can be done without copying using the unsafe `Uint8Array::view` method,
        // but I've opted to stick to safe APIs for now, until we benchmark signing.
        let data = Uint8Array::from(msg).buffer();

        let key = self
            .signing_key()
            .map_err(|e| async_signature::Error::from_source(e))?;

        let algorithm = HashMap::from_iter([("name", "RSASSA-PKCS1-v1_5")])
            .serialize(&serde_wasm_bindgen::Serializer::json_compatible())
            .map_err(|_| async_signature::Error::new())?
            .dyn_into::<js_sys::Object>()
            .map_err(|_| async_signature::Error::new())?;

        let promise = subtle
[...]

PR #1375, opened 2023-11-08
async-signature: remove mandatory Send + Sync bounds

QuinnWilton:
See: https://github.com/RustCrypto/traits/issues/1373

I've confirmed that with these changes I'm able to produce signatures from non-extractable keys in a browser, using an implementation of `AsyncSigner<S>` that calls out to the `SubtleCrypto` APIs.

I have not tested the `AsyncDigestSigner` traits in a wasm project, however I have confirmed that I'm able to build the project with the `digest` feature enabled.

### DIFF
--- a/signature/async/src/lib.rs
+++ b/signature/async/src/lib.rs
@@ -18,12 +18,8 @@ use async_trait::async_trait;
 /// (e.g. client for a Cloud KMS or HSM), returning a digital signature.
 ///
 /// This trait is an async equivalent of the [`signature::Signer`] trait.
-#[async_trait]
-pub trait AsyncSigner<S>
-where
-    Self: Send + Sync,
-    S: Send + 'static,
-{
+#[async_trait(?Send)]
+pub trait AsyncSigner<S> {
     /// Attempt to sign the given message, returning a digital signature on
     /// success, or an error if something went wrong.
     ///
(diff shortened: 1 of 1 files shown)


signature/async/src/lib.rs
```diff
@@ -18,12 +18,8 @@ use async_trait::async_trait;
 /// (e.g. client for a Cloud KMS or HSM), returning a digital signature.
 ///
 /// This trait is an async equivalent of the [`signature::Signer`] trait.
-#[async_trait]
-pub trait AsyncSigner<S>
-where
-    Self: Send + Sync,
-    S: Send + 'static,
```
tarcieri, line 25: Did you mean to remove the `'static` bound here?

tarcieri:
Going to go ahead and merge this as we'd like to do some repo reorganization which impacts this crate

QuinnWilton:
Thanks for merging it, and apologies for not getting back to you on the static bounds yesterday! I did intentionally remove them, but I probably should have split that into a separate commit.

I believe that the 'static bounds in the old code were actually superfluous, because all Send types are already 'static, and without the Send bound, requiring 'static would overconstrain the generic.

tarcieri:
Oof, well it's out with the `'static` bound, though it's an uncommonly used crate and we could probably yank it and cut a replacement release, or just bump the minor version again

QuinnWilton:
No worries at all! I'll verify it still works with my code today or tomorrow, and let you know if the 'static bound gives any issues :) 

It's just on the type of the signature itself, which is probably fine, since that'll mostly just be first-party types from rust-crypto anyway, which shouldn't really be holding any references to the stack anyway.

tarcieri:
When Rust 1.75 drops we can transition away from the `async-trait` crate to Rust's native support for async traits, at which time the `'static` bound can potentially be removed

PR #1376, opened 2023-11-09
elliptic_curve: add `BatchInvert` and `BatchNormalize` traits

ycscaly:
Addresses https://github.com/RustCrypto/elliptic-curves/issues/943


elliptic-curve/src/ops.rs
```diff
@@ -25,6 +26,77 @@ pub trait Invert {
...
+
+    /// Invert a batch of field elements in-place.
+    fn invert_batch_generic<const N: usize>(field_elements: &mut [Self; N]) -> <Self as InvertBatch>::Output;
+
+    /// Invert a batch of field elements in-place.
+    #[cfg(feature = "alloc")]
+    fn invert_batch(field_elements: &mut alloc::vec::Vec<Self>) -> <Self as InvertBatch>::Output;
+}
+
+// TODO: safe to assume here that invert performs inversion and not negation (i.e. that we're in multiplicative notation and `Mul` is the write operator)?
```
ycscaly, line 45:
```suggestion
```
@tarcieri thoughts?

ycscaly:
Also, I'd like to write tests but I couldn't as `dev::Scalar` doesn't implement `Invert`, and importing `k256` doesn't work. What should I do

ycscaly:
@tarcieri what do you think of this design? this allows for both a const-generic implementation and a dynamically `alloc`ed one while reusing the code. If you like it, I'd like to do the same with [`lincomb`](https://github.com/RustCrypto/elliptic-curves/issues/921)

elliptic-curve/src/ops.rs
tarcieri: Yes, `Neg` is the trait for negation

elliptic-curve/src/ops.rs
```diff
@@ -25,6 +26,77 @@ pub trait Invert {
...
+    /// Since inversion is performed in-place, the outputted value doesn't contain inverses of field elements.
+    /// Instead, it should be set to `Choice` when inversion may fail or `()` if it always succeeds.
+    type Output;
+
+    /// Invert a batch of field elements in-place.
+    fn invert_batch_generic<const N: usize>(field_elements: &mut [Self; N]) -> <Self as InvertBatch>::Output;
+
+    /// Invert a batch of field elements in-place.
+    #[cfg(feature = "alloc")]
+    fn invert_batch(field_elements: &mut alloc::vec::Vec<Self>) -> <Self as InvertBatch>::Output;
```
tarcieri, line 42: Why not just:

```suggestion
    fn invert_batch(field_elements: &mut [Self]) -> <Self as InvertBatch>::Output;
```

elliptic-curve/src/ops.rs
```diff
@@ -25,6 +26,77 @@ pub trait Invert {
     }
 }
 
+/// Perform a batched inversion on a sequence of field elements (i.e. base field elements or scalars)
+/// at an amortized cost that should be practically as efficient as a single inversion.
+pub trait InvertBatch: Invert + Sized {
+    /// The output type of batch inversion.
+    /// Since inversion is performed in-place, the outputted value doesn't contain inverses of field elements.
+    /// Instead, it should be set to `Choice` when inversion may fail or `()` if it always succeeds.
+    type Output;
```
tarcieri, line 35: Maybe just call this `Choice`? Or always use `subtle::Choice`? What is the benefit of making it generic?

I guess it could be `()` for `NonZeroScalar` but what does that actually buy you?

ycscaly: That definitely is an option. It buys me a clear API and no need to pattern match in the case of NonZeroScalar for example. I wasn't sure about it, I'll change it as you suggest

elliptic-curve/src/ops.rs
```diff
@@ -25,6 +26,116 @@ pub trait Invert {
...
 
+/// Perform a batched inversion on a sequence of field elements (i.e. base field elements or scalars)
+/// at an amortized cost that should be practically as efficient as a single inversion.
+pub trait InvertBatch: Invert + Sized {
+    /// Invert a batch of field elements.
+    fn invert_batch_generic<const N: usize>(field_elements: &[Self; N]) -> CtOption<[Self; N]>;
+
+    /// Invert a batch of field elements.
+    #[cfg(feature = "alloc")]
+    fn invert_batch<B: FromIterator<Self>>(field_elements: &[Self]) -> CtOption<B>;
```
tarcieri, line 37: Perhaps `invert_batch_array` and `invert_batch_slice` for naming? It at least clearly defines the distinction.

It would probably be good to also note why what is currently `invert_batch` needs allocation, e.g. "uses heap allocations internally", noting the other alternative does not have that restriction.

elliptic-curve/src/ops.rs
```diff
@@ -25,6 +26,116 @@ pub trait Invert {
...
+}
+
+impl<
+        T: Invert<Output = CtOption<Self>>
+            + Mul<Self, Output = Self>
+            + Copy
+            + Default
+            + ConditionallySelectable,
+    > InvertBatch for T
```
tarcieri, line 46:
```suggestion
impl<T> InvertBatch for T
where
    T: Invert<Output = CtOption<Self>>
        + Mul<Self, Output = Self>
        + Copy
        + Default
        + ConditionallySelectable,
```

tarcieri:
I'm reconsidering whether `InvertBatch` being a trait is a good idea. It seems like it could just be two free functions, e.g.:

```rust
pub fn invert_batch_generic<T, const N: usize>(field_elements: &[T; N]) -> CtOption<[T; N]> {
where
    T: Invert<Output = CtOption<Self>>
        + Mul<Self, Output = Self>
        + Copy
        + Default
        + ConditionallySelectable,
```

It just seems a little weird that there's `invert_batch` which is gated on the trait by `alloc`, but it's really the implementation that requires `alloc`.

tarcieri:
As another option: they could potentially be provided methods on the `Invert` trait

ycscaly:
> As another option: they could potentially be provided methods on the `Invert` trait

Can't do that, as the output from the `Invert` trait is a generic type `Output`, I need to work with `CtOption<Self>`

ycscaly:
> I'm reconsidering whether `InvertBatch` being a trait is a good idea. It seems like it could just be two free functions, e.g.:
> 
> ```rust
> pub fn invert_batch_generic<T, const N: usize>(field_elements: &[T; N]) -> CtOption<[T; N]> {
> where
>     T: Invert<Output = CtOption<Self>>
>         + Mul<T, Output = T>
>         + Copy
>         + Default
>         + ConditionallySelectable,
> ```
> 
> It just seems a little weird that there's `invert_batch` which is gated on the trait by `alloc`, but it's really the implementation that requires `alloc`.

I'm not sure that I follow your reasoning, but I've implemented what you suggested

tarcieri:
> Can't do that, as the output from the Invert trait is a generic type Output, I need to work with CtOption<Self>

You can use the exact same bounds you're already using while defining provided methods on `Invert`, including `Self: Invert<Output = CtOption<Self>>`, even if that's a bit oddly self-referential.

tarcieri:
Here's an example: https://github.com/RustCrypto/traits/commit/05bce018151bbc5712236d794e7629e86eac901e

ycscaly:
> > Can't do that, as the output from the Invert trait is a generic type Output, I need to work with CtOption
> 
> You can use the exact same bounds you're already using while defining provided methods on `Invert`, including `Self: Invert<Output = CtOption<Self>>`, even if that's a bit oddly self-referential.

Ok done

elliptic-curve/src/ops.rs
```diff
@@ -23,6 +24,117 @@ pub trait Invert {
         // Fall back on constant-time implementation by default.
         self.invert()
     }
+
+    /// Perform a batched inversion on a sequence of field elements (i.e. base field elements or scalars)
+    /// at an amortized cost that should be practically as efficient as a single inversion.
+    /// This variation takes a const-generic array and thus does not require `alloc`.
+    fn invert_batch_array<const N: usize>(field_elements: &[Self; N]) -> CtOption<[Self; N]>
```
tarcieri, line 31: Another name possibility to consider: `batch_invert_array` and `batch_invert_slice`.

I like that because "batch invert" is how I typically hear this process described.

ycscaly: I was considering both too. I'll change. Anything besides that?

tarcieri: Might have some feedback on `ToAffineBatch` but it seems mostly OK

elliptic-curve/src/arithmetic.rs
```diff
@@ -84,3 +84,21 @@ pub trait PrimeCurveArithmetic:
...
+    /// Converts a batch of points in their projective representation into the affine ones.
+    /// /// This variation takes a const-generic array and thus does not require `alloc`.
+    fn to_affine_batch_array<const N: usize>(
+        points: &[Self::ProjectivePoint; N],
+    ) -> [Self::AffinePoint; N];
+
+    /// Converts a batch of points in their projective representation into the affine ones.
+    /// However, this also requires to make dynamic allocations and as such requires `alloc`.
+    #[cfg(feature = "alloc")]
+    fn to_affine_batch_slice<B: FromIterator<Self::AffinePoint>>(
```
tarcieri, line 101: I think `group` already has a method for this: [`group::Curve::batch_normalize`](https://docs.rs/group/latest/group/trait.Curve.html#method.batch_normalize)

A trait for `to_affine_batch_array` still seems interesting, but should probably be similarly defined in terms of `group::Curve` so you can impl it for `ProjectivePoint`.

elliptic-curve/src/arithmetic.rs
```diff
@@ -84,3 +84,21 @@ pub trait PrimeCurveArithmetic:
...
+
+/// Perform a batched conversion to affine representation on a sequence of projective points
+/// at an amortized cost that should be practically as efficient as a single conversion.
+/// Internally, implementors should rely upon `InvertBatch`.
+pub trait ToAffineBatch: CurveArithmetic {
+    /// Converts a batch of points in their projective representation into the affine ones.
+    /// /// This variation takes a const-generic array and thus does not require `alloc`.
+    fn to_affine_batch_array<const N: usize>(
+        points: &[Self::ProjectivePoint; N],
+    ) -> [Self::AffinePoint; N];
```
tarcieri, line 96: FWIW I opened an issue on `group` to suggest a trait like this: https://github.com/zkcrypto/group/issues/50

elliptic-curve/src/arithmetic.rs
ycscaly: I'm ok with this, but then better change `batch_invert_slice` back to `batch_invert` to keep the same line

ycscaly: I also renamed to `Normalize` and added a `to_affine()` method, so we'll really be in line with `Invert`

elliptic-curve/src/arithmetic.rs
```diff
@@ -84,3 +84,27 @@ pub trait PrimeCurveArithmetic:
...
 }
+
+/// Normalize point(s) in projective representation by converting them to their affine ones.
+pub trait Normalize: Sized {
+    /// Elliptic curve point in affine coordinates.
+    type AffinePoint;
+
+    /// Converts the point to its affine representation.
+    fn to_affine(&self) -> Self::AffinePoint;
+
```
tarcieri, line 95: The associated type for an affine point (`AffineRepr`) and `to_affine` are already provided by [`group::Curve`](https://github.com/RustCrypto/formats/pull/810/files) which is already impl'd for all of the `ProjectivePoint` types.

```suggestion
/// Normalize point(s) in projective representation by converting them to their affine ones.
pub trait Normalize: group::Curve {
```

It also provides a [`batch_normalize`](https://docs.rs/group/latest/group/trait.Curve.html#method.batch_normalize) function, albeit with a slightly different API.

I was suggesting leveraging this trait, rather than duplicating its functionality. It should be possible to conditionally impl it when the `alloc` feature is available, falling back to its default implementation if it isn't available.

ycscaly: I know it's provided. But I wanted to have a single trait that holds it all. 

I'm not sure I understand your  requested then. Let's decide on something and merge it, the naunces are becoming a hardel at this point imo. Should we just remove to_affine() and leave everything else as is?

tarcieri: I'd like to avoid duplication of functionality, and especially having two methods with the same name that do the same thing defined on two different traits with different type signatures.

If you bound on `group::Curve`, you can use `group::Curve::AffineRepr` instead of having to redefine a separate associated type for that.

ycscaly: Ok but still we should expose both the array based and alloc based api ourselves, and also implement the group one?

tarcieri: Though I do like your API better than `group::Curve::batch_normalize`, I think the best way to proceed would be to implement that, add a new trait for `batch_normalize_array` (with `group::Curve` as a supertrait), and try to change the upstream API in `group` in the next breaking release.

I also opened an issue for having a similar trait to `batch_normalize_array` upstream in `group` as well: https://github.com/zkcrypto/group/issues/50

If you'd really like to have something like `batch_normalize` but with your original API, I guess that's okay, but I'd suggest giving it a different name to prevent confusion (though I can't think of a good one offhand).

ycscaly: But I don't even understand how I can implement `Cruve`'s `batch_normalize`? The batch inversion algorithm requires allocations to be made, and that's the reason I didn't go with this interface in the first place.

I understand that you don't want to have two conflicting APIs, but their API cannot be made to work. Therefore, I suggest to ignore it, and "override" it with ours. 

I now pushed a commit that takes `AffineRepr` from `Curve` though; I think this code should be approved.

tarcieri: > But I don't even understand how I can implement Cruve's batch_normalize? The batch inversion algorithm > requires allocations to be made

You can gate your implementation on `#[cfg(feature = "alloc")]`. It's a provided method, so when you don't provide an optimized version it will fall back onto the default implementation, which just loops over the inputs calling `to_affine`.

elliptic-curve/src/arithmetic.rs
```diff
@@ -84,3 +84,21 @@ pub trait PrimeCurveArithmetic:
...
+    fn batch_normalize_array<const N: usize>(points: &[Self; N]) -> [Self::AffineRepr; N];
+
+    /// Perform a batched conversion to affine representation on a sequence of projective points
+    /// at an amortized cost that should be practically as efficient as a single conversion.
+    /// Internally, implementors should rely upon `InvertBatch`.
+    /// This variation takes a (possibly dynamically allocated) slice and returns `FromIterator<Self::AffinePoint>`
+    /// allowing it to work with any container.
+    /// However, this also requires to make dynamic allocations and as such requires `alloc`.
+    #[cfg(feature = "alloc")]
+    fn batch_normalize_slice<B: FromIterator<Self::AffineRepr>>(points: &[Self]) -> B;
```
tarcieri, line 103: Just a thought...

```suggestion
    fn batch_normalize_vec(points: &[Self]) -> Vec<Self::AffineRepr>
    where
        Self::AffineRepr: Copy + Default
    {
        let mut ret = vec![Self::AffineRepr::default(); points.len()];
        Self::batch_normalize(points, &mut ret);
        ret
    }
```

Not quite as flexible, but probably the main use case and justifies a hard dependency on `alloc`.

ycscaly: `FromIterator` justifies `alloc` anyways, and it allows users to work with e.g. `Box`. I'm ok with switching, but I don't really see the reason

elliptic-curve/src/arithmetic.rs
ycscaly: I did as you requested, but this really became a little cumbersome; the conditional implementation doesn't work as swiftly as I'd hope.

> The associated type for an affine point (`AffineRepr`) and `to_affine` are already provided by [`group::Curve`](https://github.com/RustCrypto/formats/pull/810/files) which is [already impl'd for all of the `ProjectivePoint` types](https://docs.rs/k256/latest/k256/struct.ProjectivePoint.html#impl-Curve-for-ProjectivePoint).
> 
> It also provides a [`batch_normalize`](https://docs.rs/group/latest/group/trait.Curve.html#method.batch_normalize) function, albeit with a slightly different API.
> 
> I was suggesting leveraging this trait, rather than duplicating its functionality. It should be possible to conditionally impl `group::Curve::batch_normalize` when the `alloc` feature is available, falling back to its default implementation if it isn't available.

Please see https://github.com/RustCrypto/elliptic-curves/pull/971 and let me know if this is acceptable

elliptic-curve/src/arithmetic.rs
```diff
@@ -84,3 +84,21 @@ pub trait PrimeCurveArithmetic:
...
+    fn batch_normalize_array<const N: usize>(points: &[Self; N]) -> [Self::AffineRepr; N];
+
+    /// Perform a batched conversion to affine representation on a sequence of projective points
+    /// at an amortized cost that should be practically as efficient as a single conversion.
+    /// Internally, implementors should rely upon `InvertBatch`.
+    /// This variation takes a (possibly dynamically allocated) slice and returns `FromIterator<Self::AffinePoint>`
+    /// allowing it to work with any container.
+    /// However, this also requires to make dynamic allocations and as such requires `alloc`.
+    #[cfg(feature = "alloc")]
+    fn batch_normalize_vec(points: alloc::vec::Vec<Self>) -> alloc::vec::Vec<Self::AffineRepr>;
```
tarcieri, line 103: Seems like the input could still be a slice?
```suggestion
    fn batch_normalize_vec(points: &[u8]) -> alloc::vec::Vec<Self::AffineRepr>;
```

elliptic-curve/src/arithmetic.rs
ycscaly: I pushed a commit that does this, but it breaks https://github.com/RustCrypto/elliptic-curves/pull/971.
The reason is that Rust forces me to have an impl for the function even when `not(feature = "alloc")`: https://github.com/RustCrypto/elliptic-curves/pull/971/files#r1391796578

tarcieri: I'm confused, if you use the provided method I suggested above, you don't need to write a downstream impl at all since it's provided

elliptic-curve/src/arithmetic.rs
ycscaly: It can be, but the output still hard-requires `alloc`, so same problem

elliptic-curve/src/arithmetic.rs
ycscaly: Fixed

tarcieri (approved):
This looks good now, although I'd like to see https://github.com/RustCrypto/elliptic-curves/pull/971 green before merging

elliptic-curve/src/arithmetic.rs
```diff
@@ -84,3 +87,28 @@ pub trait PrimeCurveArithmetic:
...
+    #[cfg(feature = "alloc")]
+    fn batch_normalize_to_vec(points: &[Self]) -> Vec<Self::AffineRepr>
+    where
+        Self::AffineRepr: Copy + Default,
+    {
+        let mut ret = vec![Self::AffineRepr::default(); points.len()];
+        Self::batch_normalize(points, &mut ret);
+        ret
+    }
+}
```
tarcieri, line 114: Perhaps it's possible to use a generic trait to abstract over the difference between these two, e.g. something generic over the input or output types:

```rust
pub trait Normalize<Points>: group::Curve {
    type Output: AsRef<Self::AffineRepr>;

    // TODO: name that doesn't clash with `group::Curve::batch_normalize`
    fn batch_normalize(&Points) -> Self::Output;
}
```

Then you could have something like:

```rust
#[cfg(feature = "alloc")]
impl<Point: group::Curve> Normalize<[Point]> for Point {
    type Output = Vec<Point::AffineRepr>;

    fn batch_normalize(&[Point]) -> Vec<Point::AffineRepr> {
        let mut ret = vec![Self::AffineRepr::default(); points.len()];
        Self::batch_normalize(points, &mut ret);
        ret
    }
} 
```

ycscaly: I love this, I refactored. I don't agree that we should have a different name than `group::Curve::batch_normalize`; it happens that there are colliding names for different names, this is the appropriate name. Users can choose which API they wish to use, and the implementation is going to default to the same code anyways. In any case, I'll rename if you wish, but I really need to wrap this up soon as it became much more work than originally intended and I have other priorities. Thank you

elliptic-curve/src/arithmetic.rs
```diff
@@ -84,3 +84,14 @@ pub trait PrimeCurveArithmetic:
     /// Prime order elliptic curve group.
     type CurveGroup: group::prime::PrimeCurve<Affine = <Self as CurveArithmetic>::AffinePoint>;
 }
+
+/// Normalize point(s) in projective representation by converting them to their affine ones.
+pub trait Normalize<Points>: group::Curve {
```
tarcieri, line 89: If you want the method to be `batch_normalize`, then the trait name should probably match if it's the only method:

```suggestion
pub trait BatchNormalize<Points>: group::Curve {
```

tarcieri:
What issues did you run into with the `AsRef` bound? It seems handy for writing generic code

ycscaly:
> What issues did you run into with the `AsRef` bound? It seems handy for writing generic code

The Rust compiler complained that it didn't provided it. We can try to fight it if its important

elliptic-curve/src/arithmetic.rs
```diff
@@ -88,10 +88,10 @@ pub trait PrimeCurveArithmetic:
 /// Normalize point(s) in projective representation by converting them to their affine ones.
 pub trait BatchNormalize<Points>: group::Curve {
     /// The output of the batch normalization; a container of affine points.
-    type Output;
+    type Output: AsRef<Self::AffineRepr>;
```
tarcieri, line 91: Try:

```rust
type Output: AsRef<[Self::AffineRepr]>;
```

...but it would also have to be the following below:

```rust
fn batch_normalize(points: &Points) -> <Self as BatchNormalize<Points>>::Output;
```

ycscaly: Perfect this is working. https://github.com/RustCrypto/elliptic-curves/pull/971 should be good now too

ycscaly:
Thanks!

PR #1377, opened 2023-11-11
crypto-common: add fallible `try_generate_*_with_rng` functions

tarcieri:
Adds a fallible `try_*` variant to each `generate_*_with_rng` function, changing usages of `fill_bytes` to `try_fill_bytes`, and returning a `Result` with `rand_core::Error` in the event an RNG error has occurred.

### DIFF
--- a/crypto-common/src/lib.rs
+++ b/crypto-common/src/lib.rs
@@ -170,9 +170,18 @@ pub trait KeyInit: KeySizeUser + Sized {
     #[cfg(feature = "rand_core")]
     #[inline]
     fn generate_key_with_rng(rng: &mut impl CryptoRngCore) -> Key<Self> {
+        Self::try_generate_key_with_rng(rng).expect("RNG failure")
+    }
+
+    /// Generate random key using the provided [`CryptoRngCore`], returning an error on RNG failure.
+    #[cfg(feature = "rand_core")]
+    #[inline]
+    fn try_generate_key_with_rng(
+        rng: &mut impl CryptoRngCore,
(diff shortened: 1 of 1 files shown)


tarcieri:
We could potentially split `generate_*_with_rng` into a `try_generate_*_with_rng` variant, but I would suggest against doing the same for the `getrandom`-based APIs, which are trying to abstract over `rand_core`-related concerns.

crypto-common/src/lib.rs
```diff
@@ -163,16 +163,16 @@ pub trait KeyInit: KeySizeUser + Sized {
     #[cfg(feature = "getrandom")]
     #[inline]
     fn generate_key() -> Key<Self> {
-        Self::generate_key_with_rng(&mut OsRng)
+        Self::generate_key_with_rng(&mut OsRng).expect("RNG failure")
```
tarcieri, line 166: Could potentially use a constant for "RNG failure" to ensure it's consistent everywhere, but otherwise I think this is fine

newpavlov (commented):
I think it's a bit weird to have panicking `generate_*` methods and `Result`-based `generate_*_with_rng`. Most RNGs either don't panic or may panic only during reseeding (i.e. on `getrandom` call). If we don't want to have `try_*` methods, then i think it will be better to expose errors in `generate_*` methods as well.

tarcieri:
Okay, I'll go ahead and split up the `*_with_rng` methods too

newpavlov (approved):

PR #1378, opened 2023-11-11
crypto-common: make all `generate_*` functions fallible

tarcieri:
Redux of #1377: removes the `try_` methods in lieu of having all `generate*` methods consistently return `Result<T, rand_core::Error>`.


newpavlov:
Heh, I implemented #1379 in parallel. I think the ony difference is that it uses `getrandom` directly without relying on `rand_core`.

tarcieri:
Okay, I do like that.

PR #1379, opened 2023-11-11
Return `Result` from all `generate_*` methods

newpavlov:


tarcieri:
Beat you to it 😉  #1378

crypto-common/Cargo.toml
```diff
@@ -17,10 +17,10 @@ hybrid-array = "=0.2.0-pre.5"
 
 # optional dependencies
 rand_core = { version = "0.6.4", optional = true }
+getrandom = { version = "0.2", optional = true }
 
 [features]
 std = []
-getrandom = ["rand_core/getrandom"]
```
tarcieri, line 23: Perhaps:

```toml
getrandom = ["dep:getrandom", "rand_core?/getrandom"]
```

...which would ensure `crypto_common::rand_core::OsRng` is still available when `getrandom` and `rand_core` are both enabled.

tarcieri:
Weird (unrelated) `kem` failure there: https://github.com/RustCrypto/traits/actions/runs/6836704964/job/18591915685?pr=1379

tarcieri (approved):

tarcieri:
Of course it breaks right after I approve!

Ugh, yeah workspaces with mixed pre-and-post 1.60 crates aren't fun. FWIW I've put the other crates which use weak feature activation and/or namespacing in an `exclude`: https://github.com/RustCrypto/traits/blob/master/Cargo.toml#L15

I guess at some point we might consider bumping all of the crates in the workspace to at least 1.60 because otherwise it's an enormous pain.

.github/workflows/aead.yml
```diff
@@ -34,6 +34,8 @@ jobs:
         with:
           toolchain: ${{ matrix.rust }}
           targets: ${{ matrix.target }}
+      # Isolate this crate from workspace which is otherwise MSRV 1.60 due to namespaced features
+      - run: rm ../Cargo.toml
```
tarcieri, line 38: Instead of this you could invert the `exclude` in Cargo.toml, placing all of the 1.60+ crates in `members`, and the <1.60 crates in `exclude`:

https://github.com/RustCrypto/traits/blob/master/Cargo.toml#L15

Cargo.toml
```diff
@@ -13,7 +13,10 @@ members = [
 ]
 # re-add when all crates are MSRV 1.60+
 exclude = [
+    "aead",
+    "async-signature",
     "crypto",
+    "signature",
     "elliptic-curve"
```
tarcieri, line 20: You can remove `crypto` and `elliptic-curve` from this and add them to `members`. Likewise `aead` et al need to be removed from `members`.

I think you also need `universal-hash` in `exclude`?

tarcieri:
@newpavlov if you remove my suggested `getrandom` line, I'd be happy to take care of adding it back, which would also simplify this PR

newpavlov:
@tarcieri 
It's probably better to simply update `aead` to use `crypto-common v0.2.0-pre`.

tarcieri:
@newpavlov you'll also have to update `universal-hash` which has the same problem.

I opened #1380 to swap how `members` and `exclude` are currently being used.

newpavlov:
@tarcieri 
I am not sure why, but `universal-hashes` CI passes without issues.

newpavlov:
Duh... Updating `aead` to `crypto-common v0.2` will be slightly more involved than I thought, so I will follow your recomendation and will remove the use of namespaced features.

tarcieri:
Aah, `universal-hash` is already using the `rm ../Cargo.toml` trick

PR #1380, opened 2023-11-11
Exclude pre-1.60 crates from workspace

tarcieri:
Pre-1.60 crates can't be used in workspaces with other crates that use namespaced/weak features.

This commit places all MSRV 1.60+ crates into `members` and moves the ones with earlier MSRVs into `exclude`.


tarcieri:
Okay, that should hopefully make it green.

This will result in `clippy` and `rustfmt` being skipped for the crates in `exclude`, but I think we can just rapidly iterate on getting those crates upgraded to be MSRV 1.60+ (e.g. by doing `crypto-common` upgrades)

.github/workflows/async-signature.yml
```diff
@@ -39,10 +39,9 @@ jobs:
...
       - uses: actions/checkout@v4
       - uses: RustCrypto/actions/cargo-cache@master
-      - uses: dtolnay/rust-toolchain@master
-        with:
-          toolchain: nightly
+      - uses: dtolnay/rust-toolchain@nightly
+      - run: cargo update -Z minimal-versions
+      - uses: dtolnay/rust-toolchain@stable
       - uses: RustCrypto/actions/cargo-hack-install@master
       - run: rm ../../Cargo.toml
```
tarcieri, line 46: Can't use the shared workflow due to this, FYI (it's nested a level deeper than usual)

newpavlov: Maybe it's worth to move `async-signature` to the top level?

tarcieri:
Going to go ahead and merge this so I can PR the `getrandom` feature for `crypto-common` (which would be helpful for bumping `crypto-common` in `aead`)

newpavlov (approved):

.github/workflows/async-signature.yml
tarcieri: Yeah, it would help with things like this. There's some open PRs though, so we'd need to get those merged first.

PR #1381, opened 2023-11-12
crypto-common: weakly activate `rand_core/getrandom`

tarcieri:
When both `getrandom` and `rand_core` are enabled, activates the `rand_core/getrandom` feature, which makes `OsRng` available as `crypto_common::rand_core::OsRng`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -286,6 +286,7 @@ dependencies = [
 name = "crypto-common"
 version = "0.2.0-pre"
 dependencies = [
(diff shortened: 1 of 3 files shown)


newpavlov (approved):
Can you also add `getrandom` re-export?
```rust
#[cfg(feature = "getrandom")]
pub use getrandom;
```
It may be needed for properly handling `getrandom::Error`. I forgot to do it in the previous PR.

PR #1382, opened 2023-11-12
crypto-common: add `RngError` type

tarcieri:
Adds a common error type which can be used with both `getrandom` and `rand_core`.

This eliminates the need to have `getrandom` as part of the public API.

### DIFF
--- a/crypto-common/src/lib.rs
+++ b/crypto-common/src/lib.rs
@@ -21,6 +21,9 @@ pub use hybrid_array::typenum;
 use core::fmt;
 use hybrid_array::{typenum::Unsigned, Array, ArraySize, ByteArray};
 
+#[cfg(any(feature = "getrandom", feature = "rand_core"))]
+use core::num::NonZeroU32;
+
 #[cfg(feature = "rand_core")]
 use rand_core::CryptoRngCore;
 
@@ -162,7 +165,7 @@ pub trait KeyInit: KeySizeUser + Sized {
     #[inline]
     fn generate_key() -> Result<Key<Self>, RngError> {
         let mut key = Key::<Self>::default();
-        getrandom::getrandom(&mut key).map_err(|_| RngError)?;
+        getrandom::getrandom(&mut key).map_err(getrandom::Error::code)?;
         Ok(key)
     }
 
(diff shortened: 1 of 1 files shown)


crypto-common/src/lib.rs
```diff
@@ -370,3 +368,19 @@ impl fmt::Display for InvalidLength {
 
 #[cfg(feature = "std")]
 impl std::error::Error for InvalidLength {}
+
+/// The error type returned when a random number generator fails.
+#[cfg(any(feature = "getrandom", feature = "rand_core"))]
+#[derive(Copy, Clone, Eq, PartialEq, Debug)]
+pub struct RngError;
```
tarcieri, line 375: Potentially it could carry along an inner code e.g. the `raw_os_error` as reported by either `getrandom::Error` or `rand_core::Error`, although I don't think it's bad for it to be completely opaque

newpavlov (commented):
I don't think we should replace the errors in such way. They may carry crucial information for finding why an error has occurred. If you don't want to expose `getrandom` in public API, you could convert `getrandom::Error` into `rand_core::Error` using `From`/`Into`, but it would mean that you have to pull `rand_core` when only the `getrandom` feature is enabled.

Personally, I think the current code is fine.

newpavlov:
BTW it may be worth to add `std = ["rand_core?/std", "getrandom?/std"]` to `crypto-common`'s Cargo.toml, so the errors from those crates would have the `Error` trait implemented when `std` is enabled.

tarcieri:
> They may carry crucial information for finding why an error has occurred. 

Do you want anything more than the `code`? That's easy enough to propagate. `rand_core::Error` is just a thin wrapper for `getrandom::Error` in that regard: https://github.com/rust-random/rand/blob/ef89cbefaf484270dc3936d5d32fc2a73314173c/rand_core/src/error.rs#L178

The downside of the current approach is there are two different error types to deal with depending on which API you use.

newpavlov:
>That's easy enough to propagate. rand_core::Error is just a thin wrapper for getrandom::Error in that regard

Note the `#[cfg(feature = "std")]` part. User-defined RNGs may define their own errors, without associated error code.

>The downside of the current approach is there are two different error types to deal with depending on which functions you're calling.

I don't think `generate_*` and `generate_*_with_rng` will be mixed much in practice. And even if they are, you can easily convert `Result<T, getrandom::Error>` into `Result<T, rand_core::Error>` by using `.map_err(Into::into)`. In some cases it even may be done by `?`.

tarcieri:
Alright, I guess I'll close this

PR #1384, opened 2023-11-12
aead: bump `crypto-common` to v0.2.0-pre; MSRV 1.65

tarcieri:


aead/Cargo.toml
```diff
@@ -13,11 +13,10 @@ documentation = "https://docs.rs/aead"
 repository = "https://github.com/RustCrypto/traits"
 keywords = ["crypto", "encryption"]
 categories = ["cryptography", "no-std"]
-rust-version = "1.56"
+rust-version = "1.65"
 
 [dependencies]
-crypto-common = "0.1.4"
-generic-array = { version = "0.14", default-features = false }
+crypto-common = "=0.2.0-pre"
```
newpavlov, line 19: I wonder if we should use `crypto-common = { version = "=0.2.0-pre", path = "../crypto-common" }` here.

newpavlov (approved):

aead/Cargo.toml
tarcieri: It's pulling in that version via the `[patch.crates-io]` directive in the toplevel `Cargo.toml`, which avoids repeating it in every crate

aead/src/lib.rs
```diff
@@ -140,13 +142,15 @@ pub trait AeadCore {
     ///
     /// [NIST SP 800-38D]: https://csrc.nist.gov/publications/detail/sp/800-38d/final
     #[cfg(feature = "rand_core")]
-    fn generate_nonce(mut rng: impl CryptoRng + RngCore) -> Nonce<Self>
+    fn generate_nonce_with_rng(
```
newpavlov, line 145: Maybe we also should add `generate_nonce` method similarly to `crypto-common`?

tarcieri: Added in [963a8cf](https://github.com/RustCrypto/traits/pull/1384/commits/963a8cf5aee061effb5541c2d95c5b3c25e98f96)

PR #1385, opened 2023-11-12
universal-hash: bump crypto-common to v0.2.0-pre; MSRV 1.65

tarcieri:


universal-hash/src/lib.rs
```diff
@@ -177,10 +179,10 @@ impl std::error::Error for Error {}
 /// Split message into slice of blocks and leftover tail.
 // TODO: replace with `slice::as_chunks` on migration to const generics
 #[inline(always)]
-fn to_blocks<T, N: ArrayLength<T>>(data: &[T]) -> (&[GenericArray<T, N>], &[T]) {
+fn to_blocks<T, N: ArraySize>(data: &[T]) -> (&[Array<T, N>], &[T]) {
```
newpavlov, line 182: IIRC we already have this function in `hybrid-array`.

newpavlov (approved):

PR #1392, opened 2023-11-12
async-signature: re-add 'static bound

tarcieri:
Accidentally removed in #1375


tarcieri:
It's questionable if these bounds add value. It might still be worth removing these all together as notating them in generic cases can be a little annoying.

However, I'm adding it back for now just to keep things consistent. If we do remove these bounds, they should all be removed.

PR #1403, opened 2023-11-14
elliptic-curve: remove references from batch generic params

tarcieri:
Previously the traits were generic around a reference type. The problem with that is reference types have lifetimes, which requires notating them in generic code, in this case requiring an HRTB per generic parameter.

HRTBs work because the output has no lifetime dependencies on the input, but if that's the case we can just explicitly borrow in the function signature in trait, which means no HRTB is needed in generic code, which makes notating the trait bounds significantly simpler.

### DIFF
--- a/elliptic-curve/src/arithmetic.rs
+++ b/elliptic-curve/src/arithmetic.rs
@@ -86,12 +86,12 @@ pub trait PrimeCurveArithmetic:
 }
 
 /// Normalize point(s) in projective representation by converting them to their affine ones.
-pub trait BatchNormalize<Points>: group::Curve {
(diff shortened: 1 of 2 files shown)


tarcieri:
@ycscaly I got a generic implementation of `batch_normalize` that abstracts across the various storage types working, but it was quite unwieldy without this change

ycscaly:
Just notice that it requires us to pass a reference to a vector in the alloc case, which is a bit odd. But it works

tarcieri:
I don't understand the concern. This had no actual effect on what was being passed as a reference or not: only the bounds, which are now reference/lifetime free. All of the inputs were references already.

PR #1405, opened 2023-11-15
elliptic-curve: add `LinearCombinationExt` trait

tarcieri:
Adds a trait which is generic around its input parameters for computing linear combinations.

This allows for different strategies depending on the input parameter type, for example using stack-allocated intermediates sized to match an input array, which works on `no_std` targets.

Or, if a slice is passed, intermediate `Vec`s can be used when the `alloc` feature is available.

Also adds a blanket impl for the old `LinearCombination` trait which calls the new trait.

### DIFF
--- a/elliptic-curve/src/ops.rs
+++ b/elliptic-curve/src/ops.rs
@@ -160,6 +160,34 @@ pub trait LinearCombination: Group {
     }
 }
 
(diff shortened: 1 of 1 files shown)


ycscaly:
This is good

PR #1419, opened 2023-12-14
async-signature: move to AFIT

baloo:
`AFIT` is expected in the upcoming rust-1.75 release (scheduled for 2023/12/28).

`#[allow(async_fn_in_trait)]` is required until a solution to RPITIT is merged in rust.
This put responsability on the implementor of `async_signature::AsyncSigner` to make sure their future is `Send`able.
see https://github.com/rust-lang/rust/pull/115822#issuecomment-1731149475 for more context.

### DIFF
--- a/async-signature/src/lib.rs
+++ b/async-signature/src/lib.rs
@@ -17,6 +17,8 @@ pub use signature::{self, Error};
 #[cfg(feature = "digest")]
 pub use signature::digest::{self, Digest};
 
+#[cfg(feature = "rand_core")]
+use signature::rand_core::CryptoRngCore;
 
 /// Asynchronously sign the provided message bytestring using `Self`
 /// (e.g. client for a Cloud KMS or HSM), returning a digital signature.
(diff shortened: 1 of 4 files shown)


baloo:
yeah, but when 1.75 is released right?
I guess I can run that on beta for now until it's released.

tarcieri:
Yeah, you should be able to switch to `beta` and it should work (in theory!)

tarcieri:
Woohoo, it works(-ish)!

async-signature/src/lib.rs
```diff
@@ -33,7 +34,6 @@ pub trait AsyncSigner<S: 'static> {
     async fn sign_async(&self, msg: &[u8]) -> Result<S, Error>;
 }
 
-#[async_trait(?Send)]
 impl<S, T> AsyncSigner<S> for T
```
baloo, line 37: I'd much rather have a `impl<D, S, T> AsyncSigner<S> for T where T: AsyncDigestSigner<D, S>` instead of this blanket.

tarcieri: First, you cannot write such a blanket impl, because `D` is unconstrained (and indeed working around that issue is the whole purpose of the [signature-derive](https://github.com/RustCrypto/traits/blob/master/signature_derive/src/lib.rs):

```
error[E0207]: the type parameter `D` is not constrained by the impl trait, self type, or predicates
```

(And though `PrehashSigner` can define a default digest, you still won't be able to write a valid blanket impl with it)

Second, these traits are intended to correspond 1:1 with the traits in `signature`. If there isn't a blanket impl between `Signer` and `DigestSigner`, it doesn't make sense to have one here because that would be inconsistent.

Third, these blanket impls exist so that you can use a `Signer` as an `AsyncSigner`, i.e. if you have an API that accepts `AsyncSigner`, anyone can plug in a `Signer`, and the same goes for all of the other traits, eliminating the need to duplicate APIs everywhere just for async.

baloo: I oversimplified the previous comment a bit, but yeah I intended:
``` rust
impl<S, T> AsyncSigner<S> for T
where
    S: PrehashSignature + 'static,
    T: AsyncDigestSigner<S::Digest, S>,
{
    async fn sign_async(&self, msg: &[u8]) -> Result<S, Error> {
        self.sign_digest_async(S::Digest::new_with_prefix(msg))
            .await
    }
}
```

This is essentially what's in the `#[derive(signature::Signer)]` I guess I could add that to `signature_derive` then?

tarcieri: That does indeed seem to work, but `signature` has no such blanket impl, so I think it'd be weird for them to be inconsistent.

It could potentially be included in a hypothetical future `signature` 3.0, although the devil is in the details for these sort of blanket impls, and they may preclude overlapping impls with a valid use case.

tarcieri:
Rust 1.75 is out 🎉

tarcieri:
Merged in #1428

baloo:
Thanks!

I was in a plane yesterday and couldn't follow up.

tarcieri:
@baloo let me know if v0.5.0-pre.0 works for you and I can cut a final release

PR #1448, opened 2024-01-03
signature: define a `RandomizedSignerMut` trait

tjade273:
Extends #734 with a randomized version. This is useful for [LMS / LM-OTS](https://datatracker.ietf.org/doc/html/rfc8554) signatures, which are stateful and randomized.


tarcieri (approved):
This looks reasonable enough

PR #1451, opened 2024-01-03
digest v0.11.0-pre.2

tarcieri:


newpavlov:
Note that docs fail to build: https://docs.rs/crate/digest/0.11.0-pre.2/builds/1079051

PR #1468, opened 2024-01-10
crypto-common: bump `hybrid-array` to v0.2.0-rc.0

tarcieri:
Relaxes the `hybrid-array` requirement so we can release additional compatible versions without having to upgrade `crypto-common` each time.

Also bumps `crypto-common` to v0.2.0-pre.4 (for release)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,7 +19,7 @@ dependencies = [
  "arrayvec",
  "blobby",
  "bytes",
- "crypto-common 0.2.0-pre.3 (registry+https://github.com/rust-lang/crates.io-index)",
+ "crypto-common 0.2.0-pre.3",
  "heapless",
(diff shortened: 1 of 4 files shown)


tarcieri:
Due to circular dependency requirements between this repo and https://github.com/RustCrypto/utils we need to release `crypto-common` first, upgrade the `utils` crates, then we can bump the remaining trait crates

newpavlov (approved):
Maybe it's worth to move `hybrid-array` into its own repo, similarly to `crypto-bigint`?

tarcieri:
@newpavlov sure we could do that.

But also at play here is the `block-buffer` dependency on `crypto-common`, and `digest`'s dependency on `block-buffer`, which is circular between the two repos.

PR #1479, opened 2024-01-20
async-signature: bump `signature` to v2.3.0-pre.2

baloo:
This would be required for https://github.com/RustCrypto/formats/pull/1280/ as I'm pulling both `signature` and `async-signature`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,3 +14,6 @@ members = [
     "signature_derive",
     "universal-hash",
 ]
+
+[patch.crates-io]
+signature = { path = "./signature" }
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -14,3 +14,6 @@ members = [
     "signature_derive",
     "universal-hash",
 ]
+
+[patch.crates-io]
+signature = { path = "./signature" }
```
baloo, line 19: this I'm not sure.

I had this error otherwise:
```
error: failed to select a version for `signature`.
    ... required by package `crypto v0.5.1 (traits/crypto)`
versions that meet the requirements `^2` (locked to 2.2.0) are: 2.2.0

all possible versions conflict with previously selected packages.

  previously selected package `signature v2.3.0-pre.2`
    ... which satisfies dependency `signature = "=2.3.0-pre.2"` of package `async-signature v0.6.0-pre.0 (traits/async-signature)`

failed to select a version for `signature` which could resolve this conflict
```

tarcieri: That's fine for now

tarcieri (approved):

PR #1482, opened 2024-01-21
cipher: rename `BlockCipher*`/`BlockMode*`

tarcieri:
Renames the following traits:

- `BlockEncrypt` => `BlockCipherEncrypt`
- `BlockDecrypt` => `BlockCipherDecrypt`
- `BlockEncryptMut` => `BlockModeEncrypt`
- `BlockDecryptMut` => `BlockModeDecrypt`

As originally suggested in this comment:

https://github.com/RustCrypto/block-modes/pull/14#issuecomment-1073304690

This better reflects how these traits are actually used, i.e. `BlockCipher*` is used by ciphers, and `BlockMode*` is used by their modes of operation.

### DIFF
--- a/cipher/src/block.rs
+++ b/cipher/src/block.rs
@@ -317,39 +317,39 @@ pub trait BlockCipherDecrypt: BlockSizeUser {
 /// underlying hardware peripheral.
 pub trait BlockModeEncrypt: BlockSizeUser + Sized {
     /// Encrypt data using backend provided to the rank-2 closure.
-    fn encrypt_with_backend_mut(&mut self, f: impl BlockClosure<BlockSize = Self::BlockSize>);
+    fn encrypt_with_backend(&mut self, f: impl BlockClosure<BlockSize = Self::BlockSize>);
 
     /// Encrypt single `inout` block.
     #[inline]
-    fn encrypt_block_inout_mut(&mut self, block: InOut<'_, '_, Block<Self>>) {
-        self.encrypt_with_backend_mut(BlockCtx { block });
+    fn encrypt_block_inout(&mut self, block: InOut<'_, '_, Block<Self>>) {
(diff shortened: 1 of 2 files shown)


newpavlov (approved):
Maybe it's worth to remove the `_mut` suffixes from `BlockMode*` method names? It may cause issues when both `BlockMode*` and `BlockCipher*` imported simultaneously, but I think it's a relatively rare situation in user code.

tarcieri:
Hmm, guess I should've prepared a branch of https://github.com/rustcrypto/block-ciphers first because this is causing all sorts of wacky breakages including macro invocation failures like:

```
error: expected one of `!`, `(`, `+`, `::`, `<`, `for`, `where`, or `{`, found `:`
   --> sm4/src/lib.rs:127:1
    |
127 | / cipher::impl_simple_block_encdec!(
128 | |     Sm4, U16, cipher, block,
129 | |     encrypt: {
130 | |         let b = block.get_in();
...   |
174 | |     }
175 | | );
    | | ^
    | | |
    | |_expected one of 8 possible tokens
    |   in this macro invocation
```

cipher/src/block.rs
```diff
@@ -665,7 +653,7 @@ macro_rules! impl_simple_block_encdec {
             }
         }
 
-        impl<$($N$(:$b0$(+$b)*)?),*> $crate::BlockDecrypt for $cipher<$($N),*> {
+        impl<$($N$(:$b0$(+$b)*)?),*> $crate:BlockDecryptt for $cipher<$($N),*> {
```
tarcieri, line 656: Aha, here's the culprit re: macro invocation failures

tarcieri: Opened #1483 with fixes

tarcieri:
#1483 contains fixes to errors/omissions in this PR

PR #1487, opened 2024-01-26
crypto-common: rename `serializable_state` module to `hazmat`

tarcieri:
Also removes the toplevel re-exports for the traits, forcing users to go through the now `pub hazmat` module.

There's a lot of misuse potential with these traits, which are intended to make it possible to serialize/deserialize the internal state of hash functions. However previously they were presented side-by-side with other traits, which made that unclear.

This commit deliberately doesn't make other changes to the file so git will preserve its history:

    diff --git a/crypto-common/src/serializable_state.rs b/crypto-common/src/hazmat.rs
    similarity index 100%
    rename from crypto-common/src/serializable_state.rs
    rename to crypto-common/src/hazmat.rs

It would be very good for a followup commit to massively expand the documentation around both the module and the traits, especially to spell out the various ways they can be misused.


crypto-common/src/lib.rs
```diff
@@ -12,6 +12,9 @@
 #[cfg(feature = "std")]
 extern crate std;
 
+/// Hazardous materials.
```
tarcieri, line 15: I put this here for now to avoid modifying the now `hazmat.rs` but it would be good to move the comment into there in a followup commit

tarcieri:
Hmm, perhaps we should move the documentation checks into the "Workspace" workflow.

Running `cargo doc` on the toplevel of the repo addresses the error.

newpavlov (approved):
I wonder if it may be worth to move the serialization traits into a separate crate and make serialization support feature-gated and disabled by default.

tarcieri:
I could potentially add a `hazmat` feature and gate on that. A separate crate is also something to consider.

Issue #1494, opened 2024-01-31
Keystream buffers aren't zeroized in `StreamCipherCore`


nstilt1:
I noticed that `StreamCipherCoreWrapper` zeroizes its buffer on drop, but there are larger temporary buffers that are created in `stream_core.rs` that do not get zeroized. If the main buffer that contains a block of the keystream gets zeroized, then it would make sense for larger buffers containing the keystream to be zeroized as well.

There are at least 2 solutions to add the zeroizing buffers for `stream_core.rs`.

1) Initialize the buffers earlier in the function, outside of loops, and then zeroize them at the end if `zeroize` is enabled.
- this solution is fairly simple, and it would only add a small overhead for zeroizing. The extra `zeroize` call should only be called at most once per `apply_keystream()`

2) Make the `StreamCipherCoreWrapper`'s buffer a variable-sized buffer, or sized to the maximum size for the given cipher.
- this solution is not so simple, but it would remove the need for creating temporary buffers and it would only zeroize the single buffer when the `StreamCipherCoreWrapper` is done with.
- variable-sized buffers seem to be near impossible to pick at runtime without using the heap/`alloc`. Hybrid array might work if it is able to work with array sizes that aren't known at compile time. Otherwise, easiest solution is either solution 1, or picking the largest buffer for the cipher's backends, which would kind of defeat the purpose of `ParBlocksSize`

There may be more solutions, but I have a PR for solution 1 if that is acceptable.

newpavlov:
We intentionally do not deal with stack spilling. Zeroizing stack-based variables can significantly hurt performance and not as useful as you may think, since compiler is free to create as many copies as it wants (and often does). Stack should be bleached separately after you finished working with secret data.

nstilt1:
Is there a safe, portable way to perform stack bleaching? I have looked into it a little, but I have only found some unresolved issues and some of your x86 assembly, and I am planning on using aarch64... and I don't fully understand the stack.

But I think I have found a way to try to prevent stack spilling though, using the second solution. By removing temporary buffers from `stream_core.rs` and instead using buffers located on the `StreamCipherCoreWrapper`, or perhaps on a `StreamBackend` struct, the cipher should be flat and own all of its sensitive data. Then, an end user could wrap their cipher with `Box` or `secrecy::SecretBox` to ensure that all buffers are located on the heap. With that approach, users would not be forced to use the heap.

Adjusting `cipher` for that might be a little tricky

newpavlov:
[...]

PR #1495, opened 2024-01-31
adding zeroize for stream_core.rs (#1494)

nstilt1:
Regarding #1494

### DIFF
--- a/cipher/src/stream_core.rs
+++ b/cipher/src/stream_core.rs
@@ -38,7 +38,7 @@ pub trait StreamClosure: BlockSizeUser {
 }
 
 /// Block-level synchronous stream ciphers.
-pub trait StreamCipherCore: BlockSizeUser + Sized {
+pub trait StreamCipherCore: BlockSizeUser + ParBlocksSizeUser + Sized {
     /// Return number of remaining blocks before cipher wraps around.
     ///
     /// Returns `None` if number of remaining blocks can not be computed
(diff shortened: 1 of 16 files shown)


cipher/src/stream_core.rs
```diff
@@ -35,7 +38,7 @@ pub trait StreamClosure: BlockSizeUser {
 }
 
 /// Block-level synchronous stream ciphers.
-pub trait StreamCipherCore: BlockSizeUser + Sized {
+pub trait StreamCipherCore: BlockSizeUser + ParBlocksSizeUser + Sized {
```
newpavlov, line 41: As I wrote in the issue, this approach is incorrect. `StreamCipherCore` does not "know" at compile time the number of blocks which can be generated in parallel at runtime. It's known only by backends which can be selected using runtime target feature detection. For example, `chacha20` has SSE2 backend which generates only [one block](https://github.com/RustCrypto/stream-ciphers/blob/master/chacha20/src/backends/sse2.rs#L45-L47) and AVX2 which generates [4 blocks](https://github.com/RustCrypto/stream-ciphers/blob/master/chacha20/src/backends/avx2.rs#L59-L61). By default on x86 targets the crate uses runtime autodetection to select between those.

newpavlov (changes requested):

cipher/src/stream_core.rs
nstilt1: The approach is not ideal, but it would still work if the x86 `ParBlocksSize` for `ChaChaCore` was set to `U4`.

Is there a better approach for choosing the size at runtime? The last time I tried to pick a variable-sized `[T; N]` at runtime, the compiler was yelling at me because it did not know `N` at compile time. Is this possible with the `array` crate, or something similar?

PR #1501, opened 2024-02-02
elliptic-curve: `LinearCombinationExt` => `LinearCombination`

tarcieri:
Replaces the old, significantly less flexible `LinearCombination` trait with the newer `LinearCombinationExt` trait, removing the old trait completely and renaming the new one.

The bounds for `CurveArithmetic::ProjectivePoint` have also been updated accordingly.


tarcieri:
cc @ycscaly

Issue #1508, opened 2024-02-16
Comments on KEM traits


bifurcation:
As part of the work to [add an ML-KEM implementation](https://github.com/RustCrypto/KEMs/pull/2), I took a look at the KEM traits in this repo (in the `kem` directory).  Given that crates.io shows only a few dependents, I thought some comments might still be acceptable.  Mostly fairly minor things.

* The core point in this API is `EncappedKey`, in the sense that that's the type that knows about all of the types involved in this KEM.  That seems odd to me.
    * The encapsulated key seems more like a "plain data" type than an "intelligent" type.
    * It's not clear why the encapped key needs to know the public key types anyway, since once you have an encapped key, you mainly want to know how to *decapsulate* it.
    * I would refactor this to be more like `Signer<S>` and `Verifier<S>`, where the things the two sides need to agree on are in the generic parameters.
    * The analogy here would be `Encapsulator<EK, SS>` and `Decapsulator<EK, SS>`, since the encapsulation key and shared secret are the two things that the sender and receiver need to agree on.
    * That means it will be up to implementors of the traits to define `EK` and `SS`, but these are just byte arrays.

* `Encapsulator::try_encap` method is incorrect to take a public key argument. `self` should be represent the encapsulation key, just as in `Decapsulator::try_decap`, `self` represents the decapsulation key.

* I would drop the `AuthDecapsulator` trait for now.  It's inappropriate to have that without a corresponding `AuthEncapsulator` trait.  And AFAIK the only known implementation of these operations is DHKEM, so it doesn't seem worth generalizing right now.

* In `Encapsulator::try_encap`, you don't need to be generic over the RNG type, you can just take `&mut impl CryptoRngCore`, as in [`RandomizedDigestSigner`](https://docs.rs/signature/latest/signature/trait.RandomizedDigestSigner.html).

* Unless you're going to provide an infallible version like `Signer` does (`try_sign` vs. `sign`), I would remove the `try_` from the encap/decap methods.

* I note that there are no `as_bytes` / `from_bytes` methods on the traits for encapsulation / decapsulation keys.  That seems consistent with other traits in this repo, but just wanted to check it was intended, since at least for encapsulation keys, one will want to ship them around.

[...]

Issue #1148, opened 2022-11-02
meta: type signature for `CryptoRng`s


tarcieri:
In #1087 we discussed various possible type signatures to use for RNGs.

We currently use `impl CryptoRng + RngCore`, leveraging a blanket impl of `RngCore` for `&mut R where R: RngCore`.

This has some ergonomic problems in that confuses users who aren't aware of the blanket impl about what exactly to pass. The flipside is it permits some nice syntax like directly passing `OsRng` instead of `&mut OsRng` as a caller.

However, for the receiver it requires reborrowing the `rng` as `&mut rng` everywhere if the `rng` is used more than once. The `mut` also needs to be present on the local variable binding, so really the only thing it gets rid of is `&`.

`rand_core` v0.6.4 introduced a `CryptoRngCore` marker trait with supertrait bounds on `CryptoRng + RngCore`, as well as a blanket impl for any types which impl `CryptoRng + RngCore`.

This allows writing `impl CryptoRngCore` instead of `impl (CryptoRng + RngCore)`.

In #1147 I'm trying out `&mut impl CryptoRngCore` which minimizes syntax as well as the drawbacks for the receiver (this will be in a prerelease with the option to tweak it prior to a final v2.0 release).

Regardless of what we pick as a new convention, I think we should use the same convention everywhere.

tarcieri:
There's an additional problem with `&mut impl CryptoRngCore`: it doesn't permit RNGs passed as trait objects.

For that to work, we need to do one of:

1. `&mut (impl CryptoRngCore + ?Sized)`
2. `&mut dyn CryptoRngCore`
3. `<R: CryptoRngCore + ?Sized>(rng: &mut R)`

Of these options, I'm now leaning towards #2, and now understand why the `rand`/`rand_core` crates documented their APIs this way a little better.

I think this is one of those cases where the overhead of virtual dispatch is insignificant compared to the operation being invoked, which will involve something like a system call (for `getrandom`) or a cryptographic operation (for `thread_rng`).

newpavlov:
Personally, I prefer the third option and would be surprised to find `&mut dyn` signature in code which does not rely on dynamic dispatch in any way.

But I think, the root issue is on the `rand` side and it should be fixed in the upcoming v0.9 release. Though I haven't done the necessary design work yet.

tarcieri:
Yeah I'd be fine with the third option too.

We need to make a call before we release `signature` v2.0

garbageslam:
[...]

PR #1509, opened 2024-02-16
Simplify KEM API

rozbb:
This is neat. So per our conversation in https://github.com/RustCrypto/traits/issues/1508, I simplified everything down to two traits which are highly generic. Some benefits:

1. Way clearer what's going on
2. No need to make our own `SharedSecret` type, or any type for that matter
3. Removed need for `std` feature
4. Removed need for `generic_array` at all (now we don't need to migrate to hybrid-array at this level)
5. You can still do authenticated encap/decap! All you need is to make you `Encapsulate` struct have an identity privkey, and your `Decapsulate` struct have an identity pubkey. The X3DH example in `tests/` shows this isn't hard.

One small nit, because I know we've talked about it before: what's the now preferred way to do RNGs? It was between `mut rng: impl CryptoRngCore` or `rng: &mut impl CryptoRngCore + ?Sized`.

cc @bifurcation


kem/src/lib.rs
```diff
@@ -9,11 +9,24 @@
...
 #![warn(missing_docs, unused_qualifications, missing_debug_implementations)]
 
-#[cfg(feature = "std")]
-extern crate std;
+/// A value that can be encapsulated to. Often, this will just be a public key. However, it can
+/// also be a bundle of public keys, or it can include a sender's private key for authenticated
+/// encapsulation.
+pub trait Encapsulate<EK, SS> {
+    /// Encapsulation error
+    type Error;
```
bifurcation, line 17: It would be nice to bound this by `Debug` (`type Error: Debug;`) so that `Result::unwrap()` is happy.

kem/src/lib.rs
```diff
@@ -9,11 +9,24 @@
...
 
-mod errors;
-mod kem;
+    /// Encapsulates a fresh shared secret
+    fn encapsulate(&self, rng: impl rand_core::CryptoRngCore) -> Result<(EK, SS), Self::Error>;
```
bifurcation, line 20: It seems to me that `&mut impl rand_core::CryptoRngCore` is the right answer here.  All the methods that let you get any use out of `CryptoRngCore` take `&mut self`, so at the very least, you need a mutable thing here.  And there's no need to pass ownership of the RNG to this function.

kem/src/lib.rs
```diff
@@ -9,11 +9,24 @@
...
+}
 
-pub use crate::{errors::*, kem::*};
-pub use generic_array;
+/// A value that can be used to decapsulate an encapsulated key. Often, this will just be a secret
+/// key. But, as with [`Encapsulate`], it can be a bundle of secret keys, or it can include a
+/// sender's private key for authenticated encapsulation.
+pub trait Decapsulate<EK, SS> {
+    /// Decapsulation error
+    type Error;
```
bifurcation, line 28: Same comment here re `Debug`.

bifurcation (approved):
Thanks @rozbb, this refactor looks good to me modulo a couple of small things.

kem/src/lib.rs
tarcieri: Yeah, `&mut impl CryptoRngCore` is the convention we've selected to use in #1148

rozbb:
Addressed comments

tarcieri (approved):

tarcieri:
@bifurcation does this look good to you?

bifurcation:
LGTM!  Thanks for the nudge @tarcieri

PR #1511, opened 2024-02-20
cipher: fixup macro `block_mode_{dec,enc}_test`

baloo:
Those were throwing errors in subcrates consuming the macros.

I believe that was forgotten in #1482


tarcieri:
Thanks!

PR #1525, opened 2024-02-29
build(deps) update stable from 2.4.0 to 2.5.0

nbaztec:
Update stable from 2.4.0 to 2.5.0

Plenty of other project, specifically `rust-tls` has ported to `2.5.0` causing version conflicts with any users of the `digest` crate. 

Perhaps a minor patch can be released for it.


tarcieri:
This is `subtle`, not `stable`.

These are also compatible releases, so there’s no reason for `digest` to bump versions here. We’d only need to do that if we were using new features.

I’m guessing you mean `rustls` by `rust-tls`, but regardless these should be version-compatible.

We’ll bump the version number when we start using new features from `subtle`.

Issue #1247, opened 2023-02-20
JWK Serialization for RFC7638 Thumbprints


alexrudy:
The [JwkEcKey](https://docs.rs/elliptic-curve/latest/elliptic_curve/struct.JwkEcKey.html) struct supports serialization to JSON, but does so in a manner where the fields are always specified to match the human-readable output (i.e. with `kty` first). Unfortunately, this makes the serialization incompatible with [RFC7638](https://www.rfc-editor.org/rfc/rfc7638), JSON Web Key Thumbprints, which requires that fields are serialized in lexicographic order.

It would be great to have either (a) the default serialization output lexicographic order, so that it can be used as the input to a JWK thumbprint, or (b) a wrapper struct for serializing a JWK in lexicographic order to create a JWK thumbprint.

I guess, ideally, this would end up in the JOSE crate, but I'm opening the issue here b/c this is where I discovered the issue, and it looks like the JOSE crate isn't yet implemented.

tarcieri:
We can order the keys alphabetically, that's fine. Would you care to open a PR?

alexrudy:
Will do!

tarcieri:
@alexrudy if this is still relevant, can you refile against https://github.com/rustcrypto/jose? Thanks!

Issue #1016, opened 2022-05-26
elliptic-curve: `JwkEcKey` is too strict in what it accepts


Erik1000:
As per [RFC7517 section 4](https://datatracker.ietf.org/doc/html/rfc7517#section-4) (fourth paragraph)
> Additional members can be present in the JWK; if not understood by
   implementations encountering them, they MUST be ignored.

But currently, if you try to parse this key for example (using the `p256` crate; key created by [step-cli](https://github.com/smallstep/cli)):
```json
{
  "use": "enc",
  "kty": "EC",
  "kid": "UYa89vgc4u_lpcbbmDQfYJQRAUD4AED8H8FUNjk5KyQ",
  "crv": "P-256",
  "alg": "ECDH-ES",
  "x": "hFc6OfbgRsYFOWyhGbWH0sS5DZBjJLGABJvPttVZfA4",
  "y": "tnXB8ks0-AZJKOgbMWJrE5Jm3nTFy0UiqQugmx9jku4"
}
```

It will fail, because it does not recognize the `use` parameter (it doesn't recognize some other parameters either, but that is the first parameter it sees).
In fact, you have to reduce it to the following parameters (so only the ones from [`Field`](https://docs.rs/elliptic-curve/latest/src/elliptic_curve/jwk.rs.html#370-376) are present)

```json
{
  "kty": "EC",
  "crv": "P-256",
  "x": "hFc6OfbgRsYFOWyhGbWH0sS5DZBjJLGABJvPttVZfA4",
  "y": "tnXB8ks0-AZJKOgbMWJrE5Jm3nTFy0UiqQugmx9jku4"
}
```

There's a registry for all key parameters at [IANA](https://www.iana.org/assignments/jose/jose.xhtml#web-key-parameters) but note that this list  1) may grow with new entries and 2) implementations are allowed to define custom parameters not in the registry.

My proposed solution would be to ignore unknown parameters just like the spec dictates it.

Erik1000:
I think the whole parsing from Json Web Keys and the abstraction over different EC key types does not belong into this crate, tbh.
You can have a look how we implemented JsonWebKeys in `jose` https://docs.rs/jose/latest/jose/struct.JsonWebKey.html

ubamrein:
Yeah I was wondering, how one then would use e.g. the `p256` crate. Because until now I loved the ease of access to get a verification key from a JWK.

ubamrein:
[...]

PR #1547, opened 2024-04-07
Extends JWK implementation to include optional fields

martynp:
The intent is to provide access to the additional fields that a JWK may contain to avoid re-processing the JWK file to get at that data.

This MR will need to wait for the next release of serdect (for serde derive features) and I think elliptic-curves needs updating to make it usable? (at least the master branch emits a load of primeorder related missing trait implementations).

The main change is to add the optional fields to the JwkEcKey struct and make all fields public. I replaced the custom (de)serializer to use the derive based one, it should now be order agnostic but still enforce "EC" key types. It should also cope with any additional fields gracefully - non-standard fields would still need to be decoded separately.

I have also added in a thumbprint method which is very useful.

### DIFF
--- a/elliptic-curve/Cargo.toml
+++ b/elliptic-curve/Cargo.toml
@@ -33,13 +33,14 @@ hex-literal = { version = "0.4", optional = true }
 pem-rfc7468 = { version = "=1.0.0-pre.0", optional = true, features = ["alloc"] }
 pkcs8 = { version = "=0.11.0-pre.0", optional = true, default-features = false }
 sec1 = { version = "=0.8.0-pre.1", optional = true, features = ["subtle", "zeroize"] }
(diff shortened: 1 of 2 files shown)


martynp:
https://github.com/RustCrypto/traits/issues/1247
https://github.com/RustCrypto/traits/issues/1016

tarcieri:
I wonder if perhaps we should try to migrate from this built-in implementation to https://github.com/RustCrypto/JOSE/tree/master/jose-jwk

Erik1000:
I think we shoud not include optional fields but just allow unknown fields. Users can then use #[serde(flatten)] and a container type to get additional parameters from the JSON Object/JWK

tarcieri:
Oh sorry, I didn't notice this was open before migrating the implementation.

@martynp can you reopen this against https://github.com/RustCrypto/JOSE/blob/master/jose-jwk/src/legacy.rs ?

PR #1556, opened 2024-04-17
Add kem trait impls when ECDH is supported

incertia:
This will allow easy integration of the generic TLS 1.3 KEM combiner, since the current plan is to take any two KEMs that support Encapsulate/Decapsulate and combine them piecewise, and this turns the DH implementations into a KEM as described by TLS. I wonder if this should go into KEMs instead, however.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -552,6 +552,7 @@ dependencies = [
  "hex-literal",
  "hkdf 0.13.0-pre.3",
  "hybrid-array",
+ "kem 0.3.0-pre.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "pem-rfc7468 1.0.0-pre.0",
  "pkcs8 0.11.0-pre.0",
  "rand_core 0.6.4",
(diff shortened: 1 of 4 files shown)


tarcieri:
> ...KEM as described by TLS.

It would be good to specifically call it out as such, and add test vectors.

> I wonder if this should go into KEMs instead, however.

Yes, that'd probably be a good idea. I went ahead and snagged the `dhkem` crate if you'd like to open a PR there.

incertia:
> The impl would have to be proxied.

One upside here is that we might be able to bundle the DH-KEMs from a variety of providers, probably through an enum?

In particular since X25519-dalek does not implement the `elliptic-curve` traits to my knowledge, at least one proxy implementation must be written to get `X25519Kyber768Draft00` implemented as a generic combiner.

tarcieri:
>  The impl would have to be proxied. 

Yes, you'd need newtypes.

> One upside here is that we might be able to bundle the DH-KEMs from a variety of providers, probably through an enum?

We generally use traits, not enums, to support multiple providers. Enums have a number of drawbacks, such as preventing downstream crates from integrating, and also always linking all of the associated code for all providers whether they're used or not.

> In particular since X25519-dalek does not implement the elliptic-curve traits to my knowledge

`curve25519-dalek` supports traits from the `group` crate, although they're probably only impl'd for `EdwardsPoint` and not `MontgomeryPoint`.

That's a problem that could potentially be remedied upstream.

PR #1559, opened 2024-04-19
Add SimpleKEM and FullKEM traits to kem

incertia:
This adds a `SimpleKEM` trait, representing KEM models where the public and private keys from key generation are equivalent to the encapsulating and decapsulating keys. We also add a `FullKEM` trait, which is more general, which allows for KEM models where this is not the case, such as when trying to model authenticated X3DH as a KEM.

Motivation for this PR is mostly just promoting the `DhKem` trait from RustCrypto/KEMs#16 into the kem crate itself.

I think it is also good to have some code that forwards the calls from `SimpleKEM::encapsulate` and `FullKEM::encapsulate` to the actual `Encapsulate<EK, SS>` implementation. That way, a user can directly write `KemModel::encapsulate` and ensure that the types are correct, in the case that there are separate models that use very similar setups. e.g. unauthenticated vs authenticated modes of key exchange.

### DIFF
--- a/kem/Cargo.toml
+++ b/kem/Cargo.toml
@@ -18,7 +18,7 @@ zeroize = { version = "1.7", default-features = false }
 
 [dev-dependencies]
 hpke = "0.10"
-p256 = { version = "0.9", features = ["ecdsa"] }
+p256 = { version = "0.9", features = ["ecdh", "ecdsa"] }
 pqcrypto = { version = "0.15", default-features = false, features = [
     "pqcrypto-saber",
 ] }
--- a/kem/src/lib.rs
+++ b/kem/src/lib.rs
@@ -41,6 +41,9 @@ pub trait KEM {
     /// The type that will implement [`Decapsulate`]
     type DecapsulatingKey: Decapsulate<Self::EncapsulatedKey, Self::SharedSecret>;
 
+    /// The public key produced by [`random_keypair`]
+    type PublicKey;
+
     /// The type that will implement [`Encapsulate`]
     type EncapsulatingKey: Encapsulate<Self::EncapsulatedKey, Self::SharedSecret>;
 
@@ -51,5 +54,5 @@ pub trait KEM {
     type SharedSecret;
 
     /// Generates a new (decapsulating key, encapsulating key) keypair for the KEM model
-    fn random_keypair(rng: &mut impl CryptoRngCore) -> (Self::DecapsulatingKey, Self::EncapsulatingKey);
+    fn random_keypair(rng: &mut impl CryptoRngCore) -> (Self::DecapsulatingKey, Self::PublicKey);
 }
--- a/kem/src/lib.rs
+++ b/kem/src/lib.rs
@@ -34,14 +34,37 @@ pub trait Decapsulate<EK, SS> {
     fn decapsulate(&self, encapsulated_key: &EK) -> Result<SS, Self::Error>;
 }
 
-/// This is a trait that all KEM models should implement, and should probably be promoted to the kem
-/// crate itself. It specifies the types of encapsulating and decapsulating keys created by key
-/// generation, the shared secret type, and the encapsulated key type
-pub trait KEM {
+/// This trait represents a simplified KEM model, where the encapsulating key and public key are the
+/// same type.
+pub trait SimpleKEM {
     /// The type that will implement [`Decapsulate`]
     type DecapsulatingKey: Decapsulate<Self::EncapsulatedKey, Self::SharedSecret>;
 
-    /// The public key produced by [`random_keypair`]
+    /// The type that will implement [`Encapsulate`]
+    type EncapsulatingKey: Encapsulate<Self::EncapsulatedKey, Self::SharedSecret>;
+
+    /// The type of the encapsulated key
+    type EncapsulatedKey;
+
+    /// The type of the shared secret
+    type SharedSecret;
+
+    /// Generates a new (decapsulating key, encapsulating key) keypair for the KEM model
+    fn random_keypair(
+        rng: &mut impl CryptoRngCore,
+    ) -> (Self::DecapsulatingKey, Self::EncapsulatingKey);
+}
+
+/// This is a trait that all KEM models should implement. It represents all the stages and types
+/// necessary for a KEM, from `KeyGen() -> (DecapsKey, PublicKey)`, `Encaps(EncapsulatingKey) ->
+/// (EncappedKey, SharedSecret)`, and `Decaps(EncappedKey) -> SharedSecret`. Promotion from
+/// [`PublicKey`](FullKEM::PublicKey) to [`EncapsulatingKey`](FullKEM::EncapsulatingKey) is context
+/// dependent.
+pub trait FullKEM {
+    /// The type that will implement [`Decapsulate`]
+    type DecapsulatingKey: Decapsulate<Self::EncapsulatedKey, Self::SharedSecret>;
+
+    /// The public key produced by [`random_keypair`](FullKEM::random_keypair)
     type PublicKey;
 
     /// The type that will implement [`Encapsulate`]
@@ -55,4 +78,16 @@ pub trait KEM {
 
     /// Generates a new (decapsulating key, encapsulating key) keypair for the KEM model
     fn random_keypair(rng: &mut impl CryptoRngCore) -> (Self::DecapsulatingKey, Self::PublicKey);
+}
+
+impl<K: SimpleKEM> FullKEM for K {
+    type DecapsulatingKey = K::DecapsulatingKey;
+    type PublicKey = K::EncapsulatingKey;
+    type EncapsulatingKey = K::EncapsulatingKey;
+    type EncapsulatedKey = K::EncapsulatedKey;
+    type SharedSecret = K::SharedSecret;
+
+    fn random_keypair(rng: &mut impl CryptoRngCore) -> (Self::DecapsulatingKey, Self::PublicKey) {
+        Self::random_keypair(rng)
+    }
 }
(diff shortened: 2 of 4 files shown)


tarcieri:
Note that somewhat similarly shaped traits used to exist but were deliberately removed in #1509.

cc @rozbb

incertia:
This is also very similar to `KemCore` from `ml-kem`

```rust
pub trait KemCore {
    type SharedKeySize: ArraySize;
    type CiphertextSize: ArraySize;
    type DecapsulationKey: Decapsulate<Ciphertext<Self>, SharedKey<Self>> + EncodedSizeUser + Debug + PartialEq;
    #[cfg(not(feature = "deterministic"))]
    type EncapsulationKey: Encapsulate<Ciphertext<Self>, SharedKey<Self>> + EncodedSizeUser + Debug + PartialEq;

    #[cfg(feature = "deterministic")]
    type EncapsulationKey: Encapsulate<Ciphertext<Self>, SharedKey<Self>> + EncapsulateDeterministic<Ciphertext<Self>, SharedKey<Self>> + EncodedSizeUser + Debug + PartialEq;

    fn generate(rng: &mut impl CryptoRngCore) -> (Self::DecapsulationKey, Self::EncapsulationKey);

    #[cfg(feature = "deterministic")]
    fn generate_deterministic(d: &B32, z: &B32) -> (Self::DecapsulationKey, Self::EncapsulationKey);
}
```

rozbb:
I agree ml-kem probably shouldn't be defining its own traits. I guess I'm not sure what functionality is needed by users. The primary addition I see in `SimpleKEM` is the ability to generate keypairs. As we've discovered,  that small choice adds a lot of complexity. Do you have a strong take on whether that should be supported?

Thinking through alternatives: suppose a function is generic over a KEM and needs to be able to generate a fresh ephemeral keypair. Then perhaps it should be of the form
```rust
fn foo<EK, SS, E, F>(gen: F)
where
    E: Encapsulate<EK, SS>,
    F: Fn(&mut impl CryptoRngCore) -> (E, [u8; 32])
{
    let rng = rand::thread_rng();
    let (ek, dk) = gen(&mut rng);
    // ...
}
```
This is nice because `gen` can have whatever state the caller needs. In the simple Kyber case, this is totally stateless and returns a plain keypair. In the X3DH, `gen` can be a lambda containing an ID key and a prekey.

Thoughts?

incertia:
> Do you have a strong take on whether that should be supported?

I don't have an opinion per se, but the goal of the trait is so that one can precisely express what a specific KEM model does. i.e.

1. The key generation process generates keys of `PrivateKey` and `PublicKey`
2. The encapsulation process requires input material of type `EncapsulatingKey`, producing an `EncapsulatedKey` and `SharedSecret`
3. The decapsulation process requires input material of type `DecapsulatingKey` and `EncapsulatedKey`, producing a `SharedSecret`

As for how these types are specified, it is ultimately up to the user. The [X3DH test](https://github.com/trailofbits/traits/blob/2850d082e4dc6b218a4dc41c7dec6e43d2ebeef3/kem/tests/x3dh.rs) includes an example of how `FullKEM` would be written in the style of the test itself.

```rust
impl FullKEM for X3Dh {
    type PrivateKey = X3DhPrivkeyBundle;
    type PublicKey = X3DhPubkeyBundle;
    type DecapsulatingKey = DecapContext;
    type EncapsulatingKey = EncapContext;
    type EncapsulatedKey = EphemeralKey;
    type SharedSecret = SharedSecret;

    fn random_keypair(_: &mut impl CryptoRngCore) -> (Self::PrivateKey, Self::PublicKey) {
        let sk = Self::PrivateKey::gen();
        let pk = sk.as_pubkeys();

        (sk, pk)
    }
}

fn test_kemtrait_x3dh() {
    let mut rng = rand::thread_rng();

    let sk_ident_a = IdentityKey::default();
    let pk_ident_a = sk_ident_a.strip();
    let (sk_bundle_b, pk_bundle_b) = X3Dh::random_keypair(&mut rng);

    let encap_context = EncapContext(pk_bundle_b, sk_ident_a);
    let decap_context = DecapContext(sk_bundle_b, pk_ident_a);

    // Now do an authenticated encap
    let (encapped_key, ss1) = X3Dh::encapsulate(&encap_context, &mut rng).unwrap();
    let ss2 = X3Dh::decapsulate(&decap_context, &encapped_key).unwrap();

    assert_eq!(ss1, ss2);
}
```

rozbb:
I see what you mean. Though I might be missing the point on the example API and impl for X3DH you give.

If you defined a function that generically created a keypair and did some operations, it's very unlikely that it would generate an identity key AND a prekey AND an ephemeral key. For that reason, it seems like there's no concrete use case for a `random_keypair(rng)` function that returns a privkey bundle. Either the function should take more state (pre-existing long-term keys) or it should return only a subcomponent of the bundle. Hence why I suggest the lamba-based keygen method.

incertia:
I do agree that key generation can probably be moved outside of the trait. I think the main reason I included it is due to it being in `KemCore` and trying to do something similar. We can probably wait for @tarcieri and @bifurcation to chime in before settling on what the trait should be exactly.

tarcieri:
I think having a trait for capturing these details is good.

I'm unclear why we need two traits and why they can't build on each other. It feels like a lot of duplication.

The names should follow [RFC430](https://rust-lang.github.io/api-guidelines/naming.html), i.e. `*Kem`. It would probably be good to just define a `Kem` trait as opposed to having a prefix like `Simple*`.

bifurcation:
My initial impression is that this seems like a regression over the simplification we did in #1509.  It looks to me like the only details this interface exposes is key generation, which has not been exposed in, e.g., the traits in the `signature` crate.  If we're going to expose key generation, I would do only that, something like:

``` rust
trait KemKeyGenerate {
    type EncapKey: Encapsulator<EK>,
    type DecapKey: Decapsulator<EK>,
    fn generate(rng: &mut CryptoRng) -> Result<(Self::DecapKey, Self::EncapKey), Error>;
}
```

Having `encapsulate` and `decapsulate` on the reified KEM just seems duplicative.

If we're copying patterns from `ml-kem`, the error is probably in that crate.

tarcieri:
@incertia having a ZST to hang the overall scheme off of seems fine to me, but I would think it would only have associated types for the `Encapsulator` and `Decapsulator` so you can look them up ala `DhKem::<X25519>::Encapsulator`

bifurcation:
@tarcieri what value would that add over say `dhkem::Encapsulator<X25519>`?

tarcieri:
Alternatively, `Encapsulate` could have an associated `Decapsulator`, which would be a bit closer to how things are in `kem` v0.2

tarcieri:
@bifurcation being able to write generic code that can locate the type which performs decapsulation, similar to `signature::Keypair`

incertia:
> only have associated types for the `Encapsulator` and `Decapsulator` so you can look them up ala `DhKem::<X25519>::Encapsulator`

This also makes sense

bifurcation:
`kem::Keypair` then?  Basically, seems like we can follow the good example of `signature` here, unless there are reasons to be different.

tarcieri:
As a more concrete example of a `signature`-shaped API for this sort of thing, `rsa` defines the following traits:

- [`Decryptor`](https://docs.rs/rsa/0.9.6/rsa/traits/trait.Decryptor.html)
- [`EncryptingKeypair`](https://docs.rs/rsa/0.9.6/rsa/traits/trait.EncryptingKeypair.html)
- [`RandomizedDecryptor`](https://docs.rs/rsa/0.9.6/rsa/traits/trait.RandomizedDecryptor.html)
- [`RandomizedEncryptor`](https://docs.rs/rsa/0.9.6/rsa/traits/trait.RandomizedEncryptor.html)

(sidebar: it would be nice to support RSA-KEM eventually)

tarcieri:
@bifurcation `kem::Keypair` sounds fine to me, and unlike `signature` where there's a proliferation of possible signer traits, it could probably be bounded on `Decapsulate`

rozbb:
Ok, to sketch out at bit: we have as a starting point
```rust
pub trait signature::Keypair {
    type VerifyingKey: Clone;
    fn verifying_key(&self) -> Self::VerifyingKey;
}
```
Replacing everything with the appropriate types, we get
```
pub trait kem::Keypair<EK, SS> {
    type EncapsulationKey: Encapsulation<EK, SS>;
    fn encapsulation_key(&self) -> Self::EncapsulationKey;
}
```
Did we want to support getting the decap key from this? If so, do we know why the `signature::Keypair` trait doesn't support that?

tarcieri:
> Did we want to support getting the decap key from this? If so, do we know why the signature::Keypair trait doesn't support that?

@rozbb in `signature`, the `Keypair` trait is intended to be impl'd on signing key types themselves.

The reason `signature::Keypair` doesn't have a bound on `Signer` is because there are multiple `*Signer` traits depending on the nature of the underlying signature algorithm.

As I mentioned in this earlier, since that complication doesn't exist here, `kem::Keypair` could be bounded on `Decapsulate`.

tarcieri:
@rozbb of those, the struct looks better to me.

Either are a bit different from `signature`, which allows computing the public key from the secret key.

incertia:
> Actually, by this logic, why can't we just do:
> 
> ```rust
> struct kem::Keypair<EK, SS, E, D>
> where
>     E: Encapsulation<EK, SS>,
>     D: Decapsulation<EK, SS>,
> {
>     pub encap_key: E,
>     pub decap_key: D,
> }
> ```

following in this vein, perhaps `dhkem` will define something like `type DhKemKeypair<C: Curve> = kem::Keypair<PublicKey<C>, SharedSecret<C>, DhKemEncapsulator<C>, DhKemDecapsulator<C>>`?

tarcieri:
The `Kem` trait we ended up shipping in `kem` v0.3 looks sort of like these, but anyway, this PR is now quite stale

PR #1561, opened 2024-04-28
Add lifetime to CustomizedExtArg

sylvainpelissier:
Add a explicit lifetime for `CustomizedExtArg`. It seems to be needed. See https://github.com/RustCrypto/hashes/pull/573.

### DIFF
--- a/digest/src/core_api/wrapper.rs
+++ b/digest/src/core_api/wrapper.rs
@@ -165,7 +165,7 @@ impl<T> CustomizedInit for CoreWrapper<T>
 where
     T: BufferKindUser + CustomizedInit,
 {
-    type CustomizedExtArg = T::CustomizedExtArg;
+    type CustomizedExtArg<'a> = T::CustomizedExtArg<'a>;
 
     #[inline]
     fn new_customized(customization: &[u8]) -> Self {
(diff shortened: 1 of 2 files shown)


tarcieri (approved):

newpavlov (approved):
Hm, this is a bit unfortunate, but I guess it's the most practical solution with the current version of Rust.

tarcieri:
Hmm, this sort of API seems problematic for building generic abstractions over customized hashes. I think customization should probably be limited to a single bytestring in order to promote generic abstractions.

The cSHAKE function name seems unrelated to customization, or rather to the extent it's related it's very much cSHAKE-specific.

sylvainpelissier:
I agree It would make the trait more generic. Function name is used to build TupleHash, ParallelHash and KMAC from cSHAKE beside that I don't see other usage. For Blake2 it would handle the Salt parameter in addition to the Personalization but I don't know if it is used in practice. Do you want me to remove `CustomizedExtArg` ?

tarcieri:
@sylvainpelissier even if both constructions took 2-tuples of byte slices, the parameters aren't interchangeable: the cSHAKE function name is something very specific to cSHAKE.

> Do you want me to remove CustomizedExtArg ?

That's my vote. I think for the purposes of generic abstractions for customization the best we can do is a single byte slice.

newpavlov:
I am also fine with having simple byte slice customization only.

sylvainpelissier:
I agree, I removed it in https://github.com/RustCrypto/traits/pull/1564.

PR #1575, opened 2024-05-22
digest: expose `AssociatedAlgorithmIdentifier` through `CoreWrapper`

baloo:
This allows implementation in consumer crates like:
https://github.com/RustCrypto/hashes/pull/586

### DIFF
--- a/digest/Cargo.toml
+++ b/digest/Cargo.toml
@@ -29,7 +29,6 @@ core-api = ["block-buffer"] # Enable Core API traits
 mac = ["subtle"] # Enable MAC traits
 rand_core = ["crypto-common/rand_core"] # Enable random key generation methods
 oid = ["const-oid"]
-spki = ["oid", "dep:spki"]
 zeroize = ["dep:zeroize", "block-buffer?/zeroize"]
 alloc = []
 std = ["alloc", "crypto-common/std"]
--- a/digest/Cargo.toml
+++ b/digest/Cargo.toml
@@ -20,6 +20,7 @@ block-buffer = { version = "=0.11.0-pre.5", optional = true }
 subtle = { version = "2.4", default-features = false, optional = true }
 blobby = { version = "0.3", optional = true }
 const-oid = { version = "=0.10.0-pre.2", optional = true }
+spki = { version = "=0.8.0-pre.0", optional = true }
 zeroize = { version = "1.7", optional = true, default-features = false }
 
 [features]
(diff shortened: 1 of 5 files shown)


digest/Cargo.toml
```diff
@@ -20,6 +20,7 @@ block-buffer = { version = "=0.11.0-pre.5", optional = true }
 subtle = { version = "2.4", default-features = false, optional = true }
 blobby = { version = "0.3", optional = true }
 const-oid = { version = "=0.10.0-pre.2", optional = true }
+spki = { version = "=0.8.0-pre.0", optional = true }
```
tarcieri, line 23: Ugh, this has a circular dependency: `spki` has an optional dependency on `sha2` for computing SPKI fingerprints.

tarcieri:
As a meta comment: I would once again suggest making newtypes, rather than `CoreWrapper` type aliases, for each of the digest algorithms we support, possibly using a macro to write them.

This seems like a lot of unnecessary coupling through `digest` simply because the digest crates don't control their own types.

The type aliases are not semantically meaningful. It feels like internals leaking out.

It's also very, very confusing from a diagnostics perspective. See #1069.

digest/Cargo.toml
baloo: Ha, I missed this one.

Newtype then ? :D 
And we get the `AssociatedOid` and `AssociatedAlgorithmIdentifier` attached to those ? :D

newpavlov:
It may be worth to experiment with newtype macros. I think we should still expose block-level types and provide ways to do conversion in both directions, but for most users plain newtypes should be indeed easier to understand. We also need to decide what should be done with variable size hashes. For example, I don't think we should introduce `Blak2b512` newtype, instead it should be an alias for `Blake2b<U512>`.

digest/Cargo.toml
tarcieri: If we had newtypes, it would at least get rid of any coupling between `spki` and `digest`, though the problem would remain for `sha2`.

We could potentially split off an e.g. `spki-fingerprint` crate so `spki` doesn't depend on `sha2`. It is something of a niche feature.

tarcieri:
Would suggest closing this for now.

> instead it should be an alias for Blake2b<U512>.

@newpavlov that seems fine to me. Curious with `hybrid-array` if we could even make that `Blake2b<512>`

newpavlov:
With `hybrid-array` we still have to use `Blake2b<U512>`.

tarcieri:
@newpavlov you could potentially use the `AssocArraySize` to get `U512` from a const generic `usize`: https://docs.rs/hybrid-array/0.2.0-rc.8/hybrid_array/trait.AssocArraySize.html

baloo:
That would be `Blake2b<U64>` though? Or am I missing something?

newpavlov:
Yes, I meant `Blake2b<U64>`.

PR #1584, opened 2024-06-02
async-signature: remove `AsyncDigestSigner` blanket implementation

baloo:
This introduces a basic check to make sure our blanket implementation would not block manual implementation of the traits in this crate.

Before this commit `AsyncDigestSigner` was causing a conflict:
```
  --> async-signature/tests/mock_impl.rs:14:1
   |
14 | / impl<D> async_signature::AsyncDigestSigner<D, Signature> for MockSigner
15 | | where
16 | |     D: async_signature::Digest,
   | |_______________________________^
   |
   = note: conflicting implementation in crate `async_signature`:
           - impl<D, S, T> AsyncDigestSigner<D, S> for T
             where D: Digest, T: DigestSigner<D, S>;
   = note: downstream crates may implement trait `async_signature::signature::DigestSigner<_, Signature>` for type `MockSigner`
```

### DIFF
--- a/async-signature/src/lib.rs
+++ b/async-signature/src/lib.rs
@@ -58,17 +58,6 @@ where
     async fn sign_digest_async(&self, digest: D) -> Result<S, Error>;
 }
 
-#[cfg(feature = "digest")]
-impl<D, S, T> AsyncDigestSigner<D, S> for T
-where
-    D: Digest,
-    T: signature::DigestSigner<D, S>,
-{
-    async fn sign_digest_async(&self, digest: D) -> Result<S, Error> {
-        self.try_sign_digest(digest)
-    }
-}
-
 /// Sign the given message using the provided external randomness source.
 #[cfg(feature = "rand_core")]
 #[allow(async_fn_in_trait)]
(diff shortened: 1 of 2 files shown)


tarcieri (approved):

tarcieri (approved):
Hmm, that's unfortunate. I wonder if we should reconsider the other blanket impls as well.

tarcieri:
@baloo perhaps we should bump the version number as well? I think this is a technically a breaking change in that someone could previously use a `DigestSigner` as an `AsyncDigestSigner`

baloo:
I think other blankets were fine (I believe I tested them here, but they might still cause issues, I don't know).

This is a breaking change. I'll submit a follow-up PR to bump it.
I think the only consumer for now are in formats.git and those are not tagged yet.

baloo:
Actually, the async-signature is currently sitting at `0.6.0-pre.1`, I think I'm alright with such an API break in the next `-pre` or `-rc` release.

tarcieri:
Aah right! Seems good then

PR #1589, opened 2024-06-10
elliptic-curve: fix build failure on nightly

lucab:
This tweaks the import of `core::mem::size_of` in order to fix a build failure due to `unused_qualifications` lint, as the function has now been moved to Rust prelude (in 1.80.0 nightly).

Ref: https://github.com/rust-lang/rust/pull/123168


tarcieri:
Thanks!

PR #1598, opened 2024-06-24
Fix Clippy lints

newpavlov:


elliptic-curve/src/public_key.rs
```diff
@@ -503,6 +503,7 @@ where
 }
 
 #[cfg(feature = "pem")]
+#[allow(clippy::to_string_trait_impl)]
```
newpavlov, line 506: @tarcieri
Maybe it's better remove `ToString` implementations? Implementing `Display` on top of `to_public_key_pem` looks wrong and the trait docs state the following:

>This trait is automatically implemented for any type which implements the [Display](https://doc.rust-lang.org/std/fmt/trait.Display.html) trait. As such, ToString shouldn’t be implemented directly: [Display](https://doc.rust-lang.org/std/fmt/trait.Display.html) should be implemented instead, and you get the ToString implementation for free.

tarcieri: There is currently no way to use `core::fmt::Formatter` to serialize PEM, so `ToString` is the only option here.

You could potentially open an issue to add such support, but `pem_rfc7468` is designed around constant-time operation (not really important here due to it being a public key).

newpavlov: I mean, why have this implementation in the first place? Users should prefer using the `to_public_key_pem` method since it explicitly states how serialization is performed. IIUC `ToString` intended for user-facing serialization and PEM serialization arguably barely fits here.

newpavlov: I will leave the implementation as is, but I think it may be worth to remove it in a separate PR.

PR #1641, opened 2024-08-06
signature_derive: adds support for lifetimes

baloo:
This allows to writes things like:
```
#[derive(Debug, Signer)]
pub struct EcSigner<'ctx, C>
where
    C: PrimeCurve + CurveArithmetic,
{
    context: Mutex<&'ctx mut TransientKeyContext>,
}
```

Usecase: https://github.com/parallaxsecond/rust-tss-esapi/pull/537


tarcieri:
Why'd you close this?

baloo:
because I realized it was easy enough to implement manually.

PR #1657, opened 2024-08-28
password-hash: Add `Output::with_encoding` method

LunarLambda:
Allows `Display`/`Debug` printing `Output` with a specific `Encoding` without needing to go through `encode`/`b64_encode`, if e.g. it was created using `init_with`.

Maybe a version of `init_with` that allows specifying an encoding (similar to `new_with_encoding`) would be better, I'm not sure. This is a smaller, more general change.

### DIFF
--- a/password-hash/src/output.rs
+++ b/password-hash/src/output.rs
@@ -182,6 +182,11 @@ impl Output {
         self.encoding
     }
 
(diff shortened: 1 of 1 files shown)


tarcieri (approved):

tarcieri:
Thanks! I'd agree an alternative constructor that allows directly setting the `Encoding` might make sense too.

PR #1660, opened 2024-09-05
crypto-common: use `core::error::Error` trait; MSRV 1.81

tarcieri:
The `core::error::Error` trait is now stable, meaning we no longer need to gate impls for it on a `std` feature.


tarcieri:
Also, perhaps we should consider removing the `std` feature entirely from these crates unless it's actually used for things like filesystem access that truly need `std`.

Right now `crypto-common` has the following in its `[features]` section of Cargo.toml:

```toml
std = ["getrandom?/std", "rand_core?/std"]
```

I didn't remove it to keep the change minimal for now, but it's something we could do potentially. I don't think the transitive activation of the `std` features of those crates really helps when using them with `crypto-common`.

newpavlov:
I would really like to migrate to `core::error::Error`, but MSRV 1.81 is too aggressive in my opinion, so I think we should wait until the next release cycle.

tarcieri:
Okay, will leave this as-is and we can take a look again when 1.82 is released (October 17th)

sorairolake:
If this is merged, will you make similar changes to crates in other repositories owned by [RustCrypto](https://github.com/RustCrypto)?

newpavlov:
Yes, we will apply similar changes (including MSRV bump) to our downstream crates as well. Since https://github.com/RustCrypto/hybrid-array/pull/85 has already landed, we are likely to move forward with this change.

newpavlov (approved):

PR #1661, opened 2024-09-06
elliptic-curve: publish pre-release 0.14.0-rc.0

baloo:


tarcieri:
I guess we can go to an `rc` here. I'm expecting some possible breaking changes by way of dependencies which haven't themselves been released as `rc`.

Issue #1311, opened 2023-05-19
aead: In place API but with different buffers


ia0:
Would it make sense to add the following methods to `AeadInPlace` (similarly for `AeadMutInPlace`) with default implementations based on the existing methods?

```rust
fn encrypt_mut(&self, nonce: &Nonce, aad: &[u8], plain: &[u8], cipher_tag: &mut [u8]) -> Result<()> { ... }
fn encrypt_mut_detached(&self, nonce: &Nonce, aad: &[u8], plain: &[u8], cipher: &mut [u8]) -> Result<Tag> { ... }

fn decrypt_mut(&self, nonce: &Nonce, aad: &[u8], cipher_tag: &[u8], plain: &mut [u8]) -> Result<()> { ... }
fn decrypt_mut_detached(&self, nonce: &Nonce, aad: &[u8], cipher: &[u8], tag: &Tag, plain: &mut [u8]) -> Result<()> { ... }
```

This would add support (without extra copies or allocation) for implementations that don't support aliasing input and output. They would implement the `_mut` version and use a copy to implement the `_in_place` version using the `_mut` version.

tarcieri:
We could potentially add something like this, but I'll note the API surface is already pretty complicated.

I'm not sure it makes sense to add these to `AeadInPlace`: they're not in-place operations. But we could potentially add yet another trait with a blanket impl for types which impl `AeadInPlace`.

ia0:
I agree the current API is complicated but may I suggest to make it more complicated to possibly make it simpler:
- There are 6 "implementer" traits: one per type of implementation (in-place, 2 buffers, or alloc on one side and stateful or stateless on the other, so 6 traits). Those traits would only be used by implementers.
- There are 6 "user" traits: one per type of usage (similar to the implementer traits but with a different name). Those traits would only be used by users.
- A blanket impl from each implementer trait to each user trait when possible. Note that some user traits may not be implementable by any implementer trait.

tarcieri:
Pretty much all of our AEADs impl `AeadCore`+`AeadInPlace`, and there's already a blanket impl of `Aead` for `AeadCore`.

The `*Mut` traits are pretty much there to support hardware accelerators and we don't otherwise impl stateful AEADs.

We don't really have a distinction of "implementer" versus "user" traits: each trait is independently useful to end users.

Such a distinction exists in the `digest` crate e.g. `digest::core_api` but I'm not sure that would be something we want to carry over here.

ia0:
[...]

PR #1663, opened 2024-09-10
Add trait AeadInto so callers can provide separate input and output buffers

Cel-Service:
And identical counterpart `AeadIntoMut` for completeness' sake.

Resolves issue #1311.

## Rationale

`trait Aead` allocates a new `Vec` every time a message is encrypted or decrypted.

`trait AeadInPlace` forces the caller to use one buffer for input and output.

It's not currently possible to write an optimal _and_ specification-compliant implementation of the [Noise Protocol Framework](https://noiseprotocol.org/noise.html) using either of the existing traits.

### DIFF
--- a/aead/src/lib.rs
+++ b/aead/src/lib.rs
@@ -263,6 +263,89 @@ macro_rules! impl_decrypt_in_place {
     }};
 }
 
+
+/// Caller-provided input and output stateless AEAD trait.
+///
+/// This trait is both object safe and has no dependencies on `alloc` or `std`.
+pub trait AeadInto: AeadCore {
+    /// Encrypt the given plaintext, writing the ciphertext and authentication
+    /// tag into the given buffer.
+    ///
+    /// The buffer must have sufficient capacity to store the ciphertext
+    /// message, which will always be larger than the original plaintext.
+    /// The exact size needed is cipher-dependent, but generally includes
+    /// the size of an authentication tag.
+    ///
+    /// Returns an error if the buffer has insufficient capacity to store the
+    /// resulting ciphertext message.
+    ///
+    /// The default implementation assumes a postfix tag (ala AES-GCM,
+    /// AES-GCM-SIV, ChaCha20Poly1305). [`AeadInto`] implementations which do
+    /// not use a postfix tag will need to override this to correctly assemble
+    /// the ciphertext message.
+    fn encrypt_into(
+        &self,
+        nonce: &Nonce<Self>,
+        plaintext: &[u8],
+        associated_data: &[u8],
+        ciphertext: &mut [u8],
+    ) -> Result<()> {
+        if ciphertext.len() != plaintext.len().checked_add(Self::TagSize::to_usize()).ok_or(Error)? {
+            return Err(Error);
+        }
+        let (ciphertext, t) = ciphertext.split_at_mut(plaintext.len());
+        let tag = self.encrypt_into_detached(nonce, plaintext, associated_data, ciphertext)?;
+        t.copy_from_slice(tag.as_slice());
+        Ok(())
+    }
+
+    /// Encrypt the data into the given buffer, returning the authentication
+    /// tag.
+    fn encrypt_into_detached(
+        &self,
+        nonce: &Nonce<Self>,
+        plaintext: &[u8],
+        associated_data: &[u8],
+        ciphertext: &mut [u8],
+    ) -> Result<Tag<Self>>;
+
+    /// Decrypt the message into the given buffer, returning an error in the
+    /// event the provided authentication tag does not match the given
+    /// ciphertext (i.e. ciphertext is modified/unauthentic
+    ///
+    /// The default implementation assumes a postfix tag (ala AES-GCM,
+    /// AES-GCM-SIV, ChaCha20Poly1305). [`AeadInto`] implementations which do
+    /// not use a postfix tag will need to override this to correctly assemble
+    /// the ciphertext message.
+    fn decrypt_into(
+        &self,
+        nonce: &Nonce<Self>,
+        ciphertext: &[u8],
+        associated_data: &[u8],
+        plaintext: &mut [u8],
+    ) -> Result<()> {
+        if ciphertext.len() != plaintext.len().checked_add(Self::TagSize::to_usize()).ok_or(Error)? {
+            return Err(Error);
+        }
+        let (ciphertext, t) = ciphertext.split_at(plaintext.len());
+        let tag = t.try_into().map_err(|_| Error)?;
+        self.decrypt_into_detached(nonce, ciphertext, associated_data, plaintext, tag)?;
+        Ok(())
+    }
+
+    /// Decrypt the message into the given buffer, returning an error in the
+    /// event the provided authentication tag does not match the given
+    /// ciphertext (i.e. ciphertext is modified/unauthentic)
+    fn decrypt_into_detached(
+        &self,
+        nonce: &Nonce<Self>,
+        ciphertext: &[u8],
+        associated_data: &[u8],
+        plaintext: &mut [u8],
+        tag: &Tag<Self>,
+    ) -> Result<()>;
+}
+
 /// In-place stateless AEAD trait.
 ///
 /// This trait is both object safe and has no dependencies on `alloc` or `std`.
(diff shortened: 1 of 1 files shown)


Cel-Service:
Addendum: It's perfectly acceptable to replace the pregenerated test data if so preferred. I only wanted non-zero, non-empty test data to confirm that `encrypt_into` and `decrypt_into` would work as expected with realistic parameters.

newpavlov:
I think a better path forward is to use [`inout` types](https://docs.rs/inout) in the `AeadInPlace` trait. `cipher` already uses it, so it should be relatively easy to do.

Cel-Service:
> It might make sense to change the existing detached APIs to use `inout` types, rather than adding yet more traits. We already have a confusingly large variety of traits as-is.

To clarify -- are you proposing a backwards-incompatible change to the signature of `AeadInPlace::encrypt_in_place_detached` and `AeadInPlace::decrypt_in_place_detached` to accept `InOutBuf` instead of `&mut [u8]`?

I do still see value in having both APIs -- from a strict resource efficiency perspective, `AeadInPlace` has the lowest resource requirement. That said, I'm personally fine either way; I just didn't want to destabilize your API out of the blue.

> Having implemented parts of the Noise CipherState protocol using these crates, I don't see the issue here. Noise should work fine with in-place encryption.

It probably would have been more accurate for me to phrase my last point as "an optimal, specification-compliant, and _pedantic_ implementation of the Noise Protocol Framework" -- at least according to my own personal reading of the spec, which is obviously not the be-all, end-all of interpretations.

Of course, from a cryptographic perspective, there's no meaningful difference between an encryption function which reuses one buffer versus one which uses separate buffers.

newpavlov:
>are you proposing a backwards-incompatible change

Yes. We are in the process of preparing the next release cycle, so it's a good time to introduce it.

>merging the in-place and in-out APIs might be troublesome in Rust

See the `inout` crate docs. It provides types which allow construction from `&mut T` and `(&T, &mut T)` which under the hood use raw pointers and safe API to work with them. Ergonomic issues are usually solved by introducing separate methods for "in-place" and "buffer-to-buffer" applications with blanket impls based on methods which operate over inout types. For example, see the [`BlockEncrypt`](https://docs.rs/cipher/latest/cipher/trait.BlockEncrypt.html) trait.

Cel-Service:
I've taken a closer look at [`InOutBuf`](https://docs.rs/inout/latest/inout/struct.InOutBuf.html) and I now see that (unless I'm mistaken) it can be created from either [a pair of buffers](https://docs.rs/inout/latest/inout/struct.InOutBuf.html#method.new) _or_ [a single buffer](https://docs.rs/inout/latest/inout/struct.InOutBuf.html#method.from_mut). I can work on making that change when I have some spare time.

That said -- if breaking API changes are on the table, perhaps we should rename `AeadInPlace` to `AeadInOut`?

tarcieri:
`AeadInPlace` should stay the same (it’s generic over a `Buffer` trait which provides an abstraction over various buffer types for adding/removing the tag in-place)

Cel-Service:
> `AeadInPlace` should stay the same (it’s generic over a `Buffer` trait which provides an abstraction over various buffer types for adding/removing the tag in-place)
> 
> Edit: sorry, that is to say, there are two sets of methods, and only the `_detached` methods should be updated.
> 
> Separately I thought we split `AeadInPlace` into an `AeadDetached` or thereabouts trait already, but apparently not.

Can do. I haven't had much time to spare on this, but I've been in progress on an updated proposal based on the above conversation. I'm almost done; I've just been trying to provide a higher quality implementation for the `aes-siv` and `ocb3` crates. (The quick option would be to just copy the input buffer to the output buffer, but I'd like to avoid that where possible.)

One thing I'd like to ask: I found it quite surprising that there's no `Buffer` implementation for `&mut [u8]`; I would imagine others would as well.

It appears that the primary purpose of `trait Buffer` (as used in the `AEADs` repo) is to abstract over extending and truncating the provided buffer (in other words, offering Vec-like functionality to other types); however, only `aes-siv` makes use of this functionality.

The order (ciphertext then tag or tag then ciphertext) is already algorithm-dependent; it doesn't look like `trait Buffer` helps with the caller-side ergonomics.

Would it not make more sense to simply accept an `&mut [u8]` in its place, and explicitly specify that callers of `AeadInPlace::encrypt_in_place` should provide `AeadCore::TagSize` additional bytes of space beyond their plaintext, while callers of `AeadInPlace::decrypt_in_place` should ignore `AeadCore::TagSize` bytes at the end of their buffer?

tarcieri:
> One thing I'd like to ask: I found it quite surprising that there's no Buffer implementation for &mut [u8]; I would imagine others would as well.

`&mut [u8]` can't impl `Buffer` because it can't impl `Buffer::extend_from_slice`, i.e. it isn't growable, as you identified (seems like you answered your own question?)

> however, only aes-siv makes use of this functionality.

Every AEAD makes use of it via the default provided implementations of `AeadInPlace::{encrypt_in_place, decrypt_in_place}`, which implement the far more common postfix-appended tags, which is why you're seeing little custom logic dealing with `Buffer`:

https://docs.rs/aead/0.5.2/src/aead/lib.rs.html#287

It's also used by the blanket impl of `Aead` for `AeadInPlace` to handle populating a returned `Vec`: https://docs.rs/aead/0.5.2/src/aead/lib.rs.html#384-407

These are the primary user-facing APIs for these crates, so yes, they're very much used!

> The order (ciphertext then tag or tag then ciphertext) is already algorithm-dependent; it doesn't look like trait Buffer helps with the caller-side ergonomics.
>
> Would it not make more sense to simply accept an &mut [u8] in its place

Actually, `Buffer` and its usage in `AeadInPlace::{encrypt_in_place, decrypt_in_place}` makes it possible to abstract over the algorithm-specific details like whether or not the tag is appended or prepended.

Someone using the current API doesn't have to worry about such details, because the current API abstracts over them, allowing truly generic usage over AEAD modes regardless of how they structure the AEAD message.

Just accepting a slice as input means that the user needs to include additional space for the AEAD tag, and its position may vary depending on the algorithm.

`AeadInPlace::{encrypt_in_place, decrypt_in_place}` and `Buffer` trait are very much based on first-hand experience with the bad API ergonomics of such an approach.

Cel-Service:
After some rumination, I think I understand what you're trying to convey.

`decrypt_in_place` and `encrypt_in_place` being generic over `Buffer` means that callers can plug any implementation into `&dyn AeadInPlace` (or `impl AeadInPlace`, or so on) and expect it to Just Work, so long as they provide a growable type -- `Vec` for convenience, or `ArrayVec` (with adequate space) for control.

From that perspective, I can see how this would be very convenient in a trait that (for example) a TLS library would want to consume -- clearly, it should stay in!

---

That said, if we're open to further API changes, it would be ideal if we could _also_ make the "I have a simple `&mut [u8]` and I, too, don't care where the authentication tag goes, I just want to encrypt/decrypt this data in place" use case more ergonomic, as well.

From _that_ perspective -- the one I was seeing from before the above realization -- it seems quite surprising that `AeadInPlace::encrypt_in_place_detached` accepts `&mut [u8]`, but `AeadInPlace::encrypt_in_place` does not. As a caller, I would just want my AEAD library to take care of packing (and unpacking) the authentication tag for me -- it makes little sense, from that caller's perspective, to preclude them from using Rust-standard byte slices if they want that boilerplate taken care of.

From the caller's perspective, the simplest option I could see would likely be to add a separate pair of methods (e.g. `encrypt_slice_in_place` and `decrypt_slice_in_place`) with explicitly-defined semantics as to where the plaintext is expected to be before calling `encrypt`, and where it _will_ be after calling `decrypt`. The old methods keep their semantics, so there are still no annoying surprises for API consumers after updating their package list.

Alternatively, one could include an ergonomic route by which a `&mut [u8]` can implement `Buffer` (or be wrapped) in order to be passed directly into the existing methods. The semantics would necessarily be different, so I'm not sure this is a great option, but I figured I would mention it as an option.

tarcieri:
>  it seems quite surprising that AeadInPlace::encrypt_in_place_detached accepts &mut [u8], but AeadInPlace::encrypt_in_place does not.

If the `*detached` methods are changed to use `InOut` they should probably be renamed to `encrypt_detached` and `decrypt_detached`, since `InOut` supports separate input and output buffers and is therefore no longer necessarily "in place".

We can potentially add some explicit methods that work directly on slices too but the `InOut` APIs can also operate on slices by using `slice.into()` to convert to `InOut`. Documenting that conversion on each of the methods might be a good place to start.

Note that there are two cases for slice-based APIs to consider: in-place and working over separate input and output buffers. `InOut` abstracts over these two cases, but in considering a slice-based design you'll need to decide which to support or both.

Also it would probably be good to open a new issue for this.

PR #1679, opened 2024-09-30
cipher: bump `inout` to v0.2.0-rc.1

tarcieri:
Notably this version removes the `std` feature, a breaking change from the previous `rc.0`

### DIFF
--- a/cipher/Cargo.toml
+++ b/cipher/Cargo.toml
@@ -14,7 +14,7 @@ categories = ["cryptography", "no-std"]
 
 [dependencies]
 crypto-common = "0.2.0-rc.1"
-inout = "0.2.0-rc.0"
+inout = "0.2.0-rc.1"
 
 # optional dependencies
 blobby = { version = "0.3", optional = true }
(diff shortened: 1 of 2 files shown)


cipher/Cargo.toml
```diff
@@ -14,15 +14,15 @@ categories = ["cryptography", "no-std"]
 
 [dependencies]
 crypto-common = "0.2.0-rc.1"
-inout = "0.2.0-rc.0"
```
tisonkun, line 17: Actually this is a incompatible change. I just noticed that 0.2.0-rc.0 can be "cargo update" to 0.2.0-rc.3 which introduce https://github.com/RustCrypto/AEADs/pull/650

tarcieri: It literally says it’s a breaking change in the commit message. These are prereleases and they frequently contain breaking changes. If you want to avoid breaking changes, use stable releases.

tisonkun: Reasonable. Downgrade to aes-gcm=0.10.3 works.

PR #1694, opened 2024-10-09
Add compatibility section to the `SerializableState` docs

### LINKED ISSUE #1693: Question about the stability of `SerializableState`
Hi,

I'd like to ask about the binary format of `crypto-common::hazmat::SerializableState` and most specifically about the stability guarantees (if any, if there are none it's good to know too)?

Just for the record I'm treating is as an opaque byte array but I'm wondering if it may change and if so would that change be accompanied by a minor/major version change etc.? That is, whether I need to pin the crate versions on the receiver and sender.

For comparison, I've checked and the Go `hash` package can also serialize internal state and the compatibility is spelled out explicitly [in the docs](https://pkg.go.dev/hash#Hash) (emphasis mine):

> Compatibility: Any future changes to hash or crypto packages will endeavor to **maintain compatibility with state encoded using previous versions**. That is, any released versions of the packages should be able to decode data written with any previously released version, subject to issues such as security fixes. See the Go compatibility document for background: https://golang.org/doc/go1compat 

Thanks for your time! :wave: 

CC: @baloo @dvzrv
Related ticket: https://github.com/RustCrypto/hashes/issues/310

newpavlov:
Closes #1693

### DIFF
--- a/crypto-common/src/hazmat.rs
+++ b/crypto-common/src/hazmat.rs
@@ -28,6 +28,16 @@ impl core::error::Error for DeserializeStateError {}
 
 /// Types which can serialize the internal state and be restored from it.
 ///
+/// # Compatibility
+///
+/// Serialized state can be assumed to be stable across backwards compatible
+/// versions of an implementation crate, i.e. any `0.x.y` version of a crate
+/// should be able to decode data serialized with any other `0.x.z` version,
+/// but it may not be able to correctly decode data serialized with a non-`x`
+/// version.
+///
+/// This guarantee is a subject to issues such as security fixes
+///
 /// # SECURITY WARNING
 ///
 /// Serialized state may contain sensitive data.


crypto-common/src/hazmat.rs
```diff
@@ -28,6 +28,16 @@ impl core::error::Error for DeserializeStateError {}
...
 ///
+/// # Compatibility
+///
+/// Serialized state can be assumed to be stable across backwards compatible
+/// versions of an implementation crate, i.e. any `0.x.y` version of a crate
+/// should be able to decode data serialized with any other `0.x.z` version,
+/// but it may not be able to correctly decode data serialized with a non-`x`
+/// version.
+///
+/// This guarantee is a subject to issues such as security fixes
```
tarcieri, line 39:
```suggestion
/// This guarantee is a subject to issues such as security fixes.
```

tarcieri (approved):

PR #1695, opened 2024-10-12
CI: bump `cargo-audit` to v0.20

tarcieri:


tarcieri:
Note: a lot of this configuration is redundant now and can probably be consolidated

PR #1708, opened 2024-10-23
Use full range of the stream message counter for StreamLE31

Eosis:
I believe that we the `StreamPrimitive` for `StreamLE31` in the library is not using the correct value for `COUNTER_MAX`, choosing `0x0FFF_FFFF` rather than `0x7FFF_FFFF`, resulting in a reduced space for the message counter. This commit changes the value to use the full range.

I do not believe this is a security issue, but it does reduce the functionality of the library.

I have just committed this off the tag of the 0.5.2 release of AEAD, rather than master, as this is currently released version.

Let me know if my understanding of the problem is incorrect or if I'm attempting to commit this to the wrong library.

### DIFF
--- a/aead/src/stream.rs
+++ b/aead/src/stream.rs
@@ -490,7 +490,7 @@ where
     type NonceOverhead = U4;
     type Counter = u32;
     const COUNTER_INCR: u32 = 1;
-    const COUNTER_MAX: u32 = 0xfff_ffff;
+    const COUNTER_MAX: u32 = 0x7fff_ffff;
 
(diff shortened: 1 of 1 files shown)


Eosis:
Sister PR in the AEADs crate where this bug persists: https://github.com/RustCrypto/AEADs/pull/641

newpavlov:
Closing in favor of https://github.com/RustCrypto/AEADs/pull/641. Since this change is not critical, I don't think it's worth to do a backport for it.

Issue #1666, opened 2024-09-17
aead: marker trait + simpler encryption API for randomized nonces


tarcieri:
We currently require an explicit nonce for all encryption APIs, including ones which automatically encode the nonce as part of the AEAD message.

For some AEADs we can use purely randomized nonces safely: AES-GCM-SIV, the XChaCha* variants, and the new XAES-256-GCM are all examples of AEAD algorithms which are safe to use with nonces derived from a CSPRNG.

For these constructions, it would be nice if we had APIs that completely delegate nonce generation to a CSPRNG so we could more easily avoid nonce reuse.

tarcieri:
> Are there any protocols used in practice which append nonces to ciphertext?

I'm not aware of any. It seems generally disadvantageous because the nonce is needed before the tag, so putting it at the end is the reverse order you actually need it in (although we don't provide any APIs that would let you perform computation with the nonce without the tag)

newpavlov:
>putting it at the end is the reverse order you actually need it in

I guess it matters in the case of online application of an AEAD mode (which we currently do not expose). When you have the whole ciphertext it should be trivial to slice postfix nonce/tag.

Issue #177, opened 2020-06-08
Trait for tweakable block ciphers


newpavlov:
Threefish is an example of [tweakable](http://people.csail.mit.edu/rivest/LiskovRivest-TweakableBlockCiphers.pdf) block cipher, thus ideally we need an appropriate trait for it.

One option is to use something like this:
```Rust
pub trait TweakableBlockCipher {
    type BlockSize: ArrayLength<u8>;
    type TweakSize: ArrayLength<u8>;

    fn encrypt(&self, b: &mut Block<Self::BlockSize>,  t: &Block<Self::TweakSize>);

    fn decrypt(&self, b: &mut Block<Self::BlockSize>, t: &Block<Self::TweakSize>);
}
```
Additional context can be found in the [previous issue](https://github.com/RustCrypto/block-ciphers/issues/6).

tarcieri:
There was some discussion on https://github.com/RustCrypto/block-ciphers/issues/459 about whether to model a tweak as an additional input to the block cipher encryption/decryption operation, or having a separate method that can mutate the block cipher state prior to performing an encrypt/decrypt operation.

I thought I'd just note here that in the traditional definition of a tweakable block cipher, it's definitely modeled as a separate parameter, and is specifically described as a "new (second) input":

https://people.eecs.berkeley.edu/~daw/papers/tweak-crypto02.pdf

<img width="540" alt="Screenshot 2024-10-27 at 3 44 39 PM" src="https://github.com/user-attachments/assets/25caf9f1-2aff-4cc3-a0fe-36e461db670f">

In all applications of tweakable block ciphers I'm familiar with, the tweak changes between invocations of the block cipher, so having it be a mutable part of the state seems like it's inviting accidental "tweak reuse".

newpavlov:
Unfortunately, there are several unresolved issues with this design. How many of the block cipher trait methods we want to duplicate in new tweak traits? Should tweakable block cipher implement both `BlockCipherEnc/Dec` and `TweakBlockCipherEnc/Dec`? Or should the latter have a blanket impl for the former? Having blankets impls can cause conflicts with other blanket impls.

tarcieri:
It would impact encrypt and decrypt methods.

I don't think we should use blanket impls for these traits. I'm not even sure how they would work. To me they're just completely separate interfaces.

newpavlov:
>It would impact encrypt and decrypt methods.

[...]

PR #1713, opened 2024-11-15
aead: rework traits

### LINKED ISSUE #1672: aead: Proposal for revision to API surface for trait AeadInPlace
Migrated from PR #1663 for discussion.

@tarcieri -- following on from where we left off: I've been ruminating on the issue since my last message yesterday.

As @newpavlov [proposed](https://github.com/RustCrypto/traits/pull/1663#issuecomment-2339972548), I've been working on modifying the [algorithm crates](https://github.com/RustCrypto/AEADs) so that we can revise the API surface of `AeadInPlace` (or whatever we might rename or split it out to) to be implemented in terms of methods which consume `InOutBuf`.

---

## Fully separating authentication tag placement from the algorithm itself

As you already know, currently, the `AeadInPlace` API is generally implemented in terms of `self.encrypt_in_place_detached` and `self.decrypt_in_place_detached`. If we want to change this so that the API is implemented in terms of methods consuming `InOutBuf`, then we need to make some surrounding changes to make that work.

Implementing in-place, tag-position-agnostic decryption in terms of a call to `self.decrypt_inout_detached` (or whatever we call it) would be very simple: the API could be revised to `fn decrypt_in_place<'a>(&self, nonce: &Nonce<Self>, ad: &[u8], buffer: &'a mut [u8]) -> Result<&'a mut [u8], Error>`, very similar to plain `fn decrypt`.

The provided implementation would split the provided buffer, converting part of it into an `InOutBuf<'_, '_, u8>` and part of it into an `&Nonce<Self>` in order to pass it to `self.decrypt_inout_detached`; it would then return the subslice containing just the decrypted plaintext to the caller (thereby retaining all of the convenience that current API consumers enjoy).

The trouble is, in order to implement its counterpart, `encrypt_in_place`, in terms of `encrypt_inout_detached` _specifically for tag-prepending AEADs_, something needs to give:

- If we were to just convert a subslice of the buffer into an `InOutBuf` (as above), the caller would need to put their plaintext at an algorithm-dependent offset -- this would be VERY surprising and strange breaking behavior, so it's not an option.

- If we want to sand down the above rough edge but keep the concept, `InOutBuf` would need to support an offset (so that `.get_in()` would return just the plaintext region, and `.get_out()` would return just the ciphertext region).

- If we just throw our hands up and accept the status quo, then the `aes-siv` crate (and any implementation of `AeadInPlace` which prepends its authentication tag) needs to implement BOTH an in-place, authentication-tag-packing-aware encryption method (one which has to read plaintext from the start and write ciphertext at an offset) AND in-out, authentication-tag-packing-unaware encryption method (one which simply reads plaintext from the in-buffer and write ciphertext to the out-buffer at the same position).

Honestly, I'm personally of the mind that the last option might be the most sensible one in terms of time spent, but I haven't yet dug deep into the `aes-siv` crate to gauge how much work would be required and whether I'm confident enough to implement the necessary changes optimally. That said, I think that adjusting `InOutBuf` to suit this use case might end up being the more flexible and elegant option.

What do you think? Am I overlooking something?

## Simplifying wrappers

On a lesser note, as I was mentioning last night in the above PR, I'd very much like to see the specialist buffer types entirely abstracted away from callers in wrapper methods where possible.

As above, whatever we end up doing to `encrypt_in_place`, it would be ideal if we could also have an in-place encryption method that foregoes specialist types in exchange for slightly more complex semantics (needing to provide at least `Algorithm::TagSize::to_usize()` additional bytes of unused space).

_Once we've resolved the conundrum in the section above_, so long as adding a separate method is acceptable, it should be a fairly straightforward implementation.

newpavlov:
This PR introduced the unified `Aead` trait which replaces the old `Aead*` traits. It adds `inout` support and helper methods for buffer-to-buffer encryption and decryption. The trait also provides a set of methods for postfix encryption/decryption.

Tag position in ciphertext is defined by the `IS_POSTFIX` associated constant used by the `en/decrypt_to_vec` and `en/decrypt_to_buffer` methods.

The `Buffer` trait is tweaked slightly in accordance to operations which are used in the `Aead` trait methods.

Closes #1672


tarcieri:
FWIW I have an incomplete local branch trying to do similar things (mostly implementing the ideas I talked about in #1672)

newpavlov:
@tarcieri 
WDYT about usefulness of the `arrayvec`, `bytes`, and `heapless` bridges based on `Buffer`? I don't know how widely these bridges are used in practice, but to me it seems like the new APIs can be easily used in embedded contexts on top of those crates with a small amount of glue code. So I am somewhat inclined towards completely removing the `Buffer`-based methods.

tarcieri:
@newpavlov as I said in #1672, I am a user of those APIs and know of other embedded users of those APIs. The "small amount of glue code" involves length calculations on slices which those APIs abstract away for users of e.g. `arrayvec`. Dealing with the fallout of removing them would be rather annoying for me personally.

I also don't think we should capriciously remove features and break downstream users' code. Features should only be removed if they're a misfeature or they don't have any users, neither of which is the case here. And I especially don't think we should be comingling the removal of major user-facing features in large PRs which are making a large number of unrelated changes such as this one.

aead/src/lib.rs
```diff
@@ -68,96 +75,220 @@ impl fmt::Display for Error {
...
-
-/// Authenticated Encryption with Associated Data (AEAD) algorithm core trait.
-///
-/// Defines nonce, tag, and overhead sizes that are consumed by various other
-/// `Aead*` traits.
-pub trait AeadCore {
+/// Authenticated Encryption with Associated Data (AEAD) algorithm trait.
+pub trait Aead {
     /// The length of a nonce.
     type NonceSize: ArraySize;
```
newpavlov, line 81: Maybe we should use `IvSizeUser` here instead (after renaming it to `NonceSizeUser`)? We will get the RNG methods and `Nonce` alias "for free". It also would work a bit nicer with the stream traits.

aead/src/dyn_aead.rs
```diff
@@ -0,0 +1,201 @@
...
+use alloc::vec::Vec;
+
+mod sealed {
+    pub trait Sealed {}
+}
+
+/// Object-safe variant of the [`Aead`] trait.
+///
+/// This trait is implemented automaticlly for all types which implement the [`Aead`] trait.
+pub trait DynAead: sealed::Sealed {
```
tarcieri, line 15: This seems unrelated to the `inout` changes and makes the PR harder to review due to the extra complexity.

I would suggest saving it for a followup PR in order to keep this PR better focused.

aead/src/lib.rs
```diff
@@ -68,96 +75,269 @@ impl fmt::Display for Error {
...
+        let tag_len = Self::TagSize::USIZE;
+        let pt_len = ciphertext.len().checked_sub(tag_len).ok_or(Error)?;
+        let pt_dst = buffer.get_mut(..pt_len).ok_or(Error)?;
+        let (ct, tag) = ciphertext.split_at(pt_len);
+        let tag = tag.try_into().expect("tag has correct length");
+        let buf = InOutBuf::new(ct, pt_dst).expect("buffers have the same length");
+        self.detached_decrypt_inout(nonce, associated_data, buf, tag)?;
+        Ok(pt_dst)
+    }
+
```
tarcieri, line 308: These impls are all incredibly redundant and suggest to me that this PR isn't working at a good abstraction level.

aead/src/lib.rs
```diff
@@ -68,96 +75,269 @@ impl fmt::Display for Error {
...
-    /// The upper bound amount of additional space required to support a
-    /// ciphertext vs. a plaintext.
-    type CiphertextOverhead: ArraySize + Unsigned;
+    /// Constant which defines whether AEAD specification appends or prepends tags.
+    ///
+    /// It influences the behavior of the [`Aead::encrypt_to_vec`], [`Aead::decrypt_to_vec`],
+    /// [`Aead::encrypt_to_buffer`], and [`Aead::decrypt_to_buffer`] methods.
+    ///
+    /// If the specification does not explicitly specify tag kind, we default to postfix tags.
+    const IS_POSTFIX: bool = true;
```
tarcieri, line 92: I definitely prefer the approach in #1714 of using a marker trait for postfix-tagged AEADs and using that to gate a blanket impl, especially since we can't provide a good efficient implementation of prefix-tagged AEADs.

tarcieri:
As a general statement, this PR contains a lot of breaking and a lot of irrelevant, unrelated changes, especially with its stated goal of addressing #1672.

I would generally prefer to see smaller, more focused PRs that try to solve one-problem-at-a-time, and avoid making unnecessary breaking changes.

aead/src/dyn_aead.rs
newpavlov: I added this trait because the `Aead` trait is no longer object-safe even after fixing the associated parameters. But we can leave it for a later PR if you want.

aead/src/lib.rs
newpavlov: A separate marker trait would result in a more complex API surface and artificial (in my opinion) restriction on using the postfix methods. Right now we have everything in one trait with easy-to-read docs.

newpavlov:
Please consider it a ground up redesign of the traits. I was not happy with the existing API for quite some time and it's a good opportunity to work on it.

If we are to exclude the `DynAead` trait, this PR contains less than 400 lines of changes. I don't think it makes sense to split it into separate PRs since the changes are connected to each other.

aead/src/lib.rs
tarcieri: This PR adds hundreds of lines of code and dozens of new methods without adding any real new functionality, other than `inout` integration which should only take a few dozen lines of code.

Compared to #1714, the API surface is vastly more complex.

aead/src/lib.rs
tarcieri: The current implementation is based on abstractions which eliminate this boilerplate. Your version is very much not DRY.

newpavlov: The current implementation does not deal with `inout`. I could've just removed the `inplace` and `to_buf` methods to comply with DRY, but I doubt users would like such API.

aead/src/lib.rs
newpavlov: Most of the added methods are ergonomic `inplace` and `to_buf` methods. 

>few dozen lines of code

Can you show these "few dozen lines"? Because unless you want to remove the helper methods, I don't see how exactly you plan to do this.

tarcieri:
> Please consider it a ground up redesign of the crate. I was not happy with the existing API for quite some time and it's a good opportunity to work on it.

I don't think "ground up redesigns" make sense at this point in the project: the `aead` crate has 50 million downloads. Completely redesigning the API and asking all the existing users to switch to it will break everyone's code. To justify that, there must be something very fundamentally wrong with the current design.

I don't think that's the case here and I think the proposed new API is very much a regression.

The current design is based on being simple to understand and use. The `Aead` trait is targeted at the crate's foremost users: those with a working `alloc` implementation. The `AeadInPlace` trait is targeted at embedded/power users. These traits each have a minimal number of methods: just `encrypt` and `decrypt` for `Aead`, which follows the design for an AEAD interface you will see in most other programming language environments.

You have combined these into a "god object" trait with dozens of methods. In doing so you removed the blanket impls that allow the previous traits to abstractly build on one another and replaced them with copy-and-paste code. The crate was previously `forbid(unsafe_code)` and now contains unsafe code for reasons I'm uncertain of.

aead/src/lib.rs
tarcieri: The changes needed to support `inout` are minimal. I can continue them in #1714 (or a separate PR) for comparison.

tarcieri:
> Should we add something for automatically prepending/appending nonces to ciphertexts?

I think a higher-level API that wraps up calling `generate_nonce` and prepending a random nonce would be good, as commonly seen in other AEAD APIs (e.g. Tink). A random nonce is the foremost use case for a prepended nonce. See #1666.

newpavlov:
>Completely redesigning the API and asking all the existing users to switch to it will break everyone's code.

No, it will not break user's code without manually updating dependencies. Changes required in user code will be quite small, far smaller than some of API changes in our other popular crates. Arguably, migration to `hybrid-array` will cause users more annoyance than changes in this PR.

We are still far from 1.0 release and I don't think it's a right time to go full ossification route.

>there must be something very fundamentally wrong with the current design

There was nothing "fundamentally" wrong with `generic-array` and yet here we are migrating all our crates to `hybrid-array`.

The question is whether the new API is a better long-term or not. I believe the answer is yes, it has one clear AEAD trait instead of splitting functionality into several different traits. It's more consistent with the `cipher` APIs. It does not rely on dynamic dispatch for the `Buffer` support.

You call the new `Aead` trait a "god object", but it's no more godly than for example `Iterator`. One trait is easier for users to understand and browse docs for than several smaller traits. You need to provide arguments for _why_ a split could be beneficial. I think the dynamic dispatch use case is better served by a different trait with `&[u8]` nonces or with methods which opaquely add nonce to ciphertext. For example, in the `digest` crate I would've liked to have less traits, but their existence was dictated by difference in functionality between hash algorithms.

One potential split which could be beneficial is between a "core" trait (`AeadCore`) and the main sealed "extension" trait (`Aead`) containing all helper methods. The seal will be used to prevent users from potentially overwriting methods with blanket implementations.

We can have a "minimal" number of methods by removing most helper methods, but I doubt users will thank us for that.

>There seem to be many methods defined on the trait that aren't intended for end-users but are internal implementation details.

All methods in the PR are intended for potential use by users. The trait just has a trifecta of `inout`, `inplace`, and `_to_buf` methods, which creates an initial impression of API complexity. All 3 can be useful in different contexts. What methods do you think can be removed?

>The crate was previously forbid(unsafe_code) and now contains unsafe code for reasons I'm uncertain of, but which feels like mixing abstraction levels. 

As mentioned in the TODO comment it's a temporary change until the required methods will be added to the `inout` crate.

tarcieri:
> Arguably, migration to hybrid-array will cause users more annoyance than changes in this PR.

`hybrid-array` has a full migration guide, which is something that's been lacking from similar changes like this in the past: https://docs.rs/hybrid-array/latest/hybrid_array/#migrating-from-generic-array

> It's a bit off-topic, but that exact trouble are you talking about? 

If you have separate traits for standard block ciphers versus tweakable block ciphers, under the current design you would also have to duplicate all of the methods for using backends as well, which didn't exist in previous versions of the trait:

https://github.com/RustCrypto/traits/issues/177#issuecomment-2442162213

aead/src/lib.rs
newpavlov: I have reduced the amount of code duplication in this [commit](https://github.com/RustCrypto/traits/pull/1713/commits/1e55bec6044b5c32a45143024a25b24f2064d58b). Does this address your concerns?

aead/src/lib.rs
newpavlov: I have reduced the amount of code duplication in this [commit](https://github.com/RustCrypto/traits/pull/1713/commits/1e55bec6044b5c32a45143024a25b24f2064d58b). Does this address your concerns?

tarcieri:
The current design of the `aead` crate was the result of one of the most commented on PRs in the history of this project: #40

This PR is re-litigating literally everything in the current design 5 years later, in absence of that previous discussion.

There was no planning or discussion, or even a prose description of what's wrong with the current design and what the goals of a rewrite are. So far you haven't even identified those things except to say you were "not happy with the existing API". What was wrong? Why does it warrant a ground up rewrite? What is being done differently here to address those issues?

I think if you took a step back and tried to actually plan things out by writing down goals, you would discover many of the changes you want don't require a ground-up rewrite but can be done incrementally, which would make them much easier to understand than a single PR that changes everything, especially for anyone reviewing the source history.

Perhaps in the process you could avoid capricious breaking changes which don't add new functionality or other improvements but merely generate churn for users of the library.

newpavlov:
As I've written above, I do not consider the ossification arguments to be valid. We are right in the middle of a major breaking release cycle, so it's a good time to reconsider the existing APIs without being tied to backward compatibility. Therefore, I believe the APIs should be evaluated on their own merits, regardless of how long they have existed.

5 years ago we did not have the `inout` crate, we made some wrong assumptions like `CiphertextOverhead` and using the `*Mut` traits as HAL (remember, earlier we removed it from `cipher`). Requiring to fix tag and nonce size for object safety is also quite inconvenient and inflexible. It also introduces a small inefficiency around `&mut dyn Buffer` which is not present with `impl Buffer`. The split between `AeadCore` and `AeadInPlace` looks simply redundant. What types implement the former, but not the latter?

I have mostly left the `aead` crate to you and did not participate much in its development outside of discussions. While these changes may seem sudden, I hope you will review them with an open mind.

Let's review the new methods step-by-step to find where exactly our disagreements start (let's consider only their essence, we can bikeshed the names and in which traits they should be defined separately):
1) The fundamental methods: `detached_encrypt_inout`, `detached_encrypt_inout`. You may have some different ideas about "`inout` integration which should only take a few dozen lines of code", but I don't see how you can get more fundamental than these methods and you haven't provided a draft of your ideas yet. Note that an `inout` support on top of `&mut [u8]` methods will be simply less efficient.
2) The helper detached methods: `detached_encrypt_inplace`, `detached_decrypt_inplace`, `detached_decrypt_inplace`, `detached_decrypt_to_buf`. These methods could be removed, but they are consistent with the `cipher` traits and quite helpful in practice, especially the `inplace` methods.
3) The basic postifx methods: `postfix_encrypt_inout`, `postfix_decrypt_inout`. These methods could be used even for AEADs which specify the prefix mode of operation. You've argued that this may result in "non-standard" algorithms, but I think the method name clearly indicates what happens, so users are unlikely to use it accidentally. We can add a big fat warning about it to the trait docs to reduce this probability even further.
4) The helper postfix methods: `detached_encrypt_inplace`, `detached_decrypt_inplace`, `detached_decrypt_inplace`, `detached_decrypt_to_buf`. These methods are not controversial iff you agree with the steps 2 and 3.
5) The postfix `Buffer` methods: `postfix_encrypt_buffer`, `postfix_decrypt_buffer`. As discussed in #1672, we can provide an efficient in-place implementation only for the postfix mode. And since we already defined `Buffer` for methods in the next step, I think it's worth to use it for postfix methods as well.
6) High-level `Buffer` methods: `encrypt_to_buffer`, `decrypt_to_buffer`. Straightforward use of `Buffer` which is useful in practice as argued by you above. 
7) High-level `Vec` specialization: `encrypt_to_vec`, `decrypt_to_vec`. I think these methods are useful for clarity and easier to use, since most users will encrypt message to vectors instead of other `Buffer` implementers.

I also have some ideas for an `inout` type which reserves a fixed amount from head and tail, which would allow to implement an efficient set of prefix methods, but I will postpone their implementation until we finish discussing the existing methods.

tarcieri:
> As I've written above, I do not consider the ossification arguments to be valid. 

It's not an "ossification argument", it's asking you to justify why your changes are an improvement over the old code.

If breaking changes aren't actually improvements, if they're just moving things around, if they're not better but just different, then it would be better to leave things as they are, because otherwise you're just making chores for everyone and wasting their time without actually improving the library.

> I have mostly left the aead crate to you and did not participate much in its development outside of discussions. While these changes may seem sudden, I hope you will review them with an open mind.

It feels as though your first major contribution to the library is to steamroll over everything, changing literally every method name.

> 5 years ago we did not have the inout crate, we made some wrong assumptions like CiphertextOverhead and using the *Mut traits as HAL (remember, earlier we removed it from cipher). 

These all seem like things that can be addressed in self-contained PRs, rather than having to rewrite everything at once.

> Let's review the new methods step-by-step to find where exactly our disagreements start 

Our disagreements start with how to proceed with these changes.

There's an inherent lack of focus in trying to change literally everything at once. I don't think we can drill down on that list productively in this issue. It's too much going on at once. I think it would be better to make separate issues/PRs to discuss each of those changes.

You haven't even enumerated everything yet. Just as one example, you've removed the `generate_nonce` method and all of its associated documentation. I mean, really you've removed *all* of the documentation.

Documentation, or lack thereof, is one of the biggest complaints about this project. I really, really hope this PR doesn't turn into a huge documentation regression.

newpavlov:
>If breaking changes aren't actually improvements, if they're just moving things around

You haven't addressed what exactly you consider "moving around" and what not. I've compiled the list specifically for that purpose. The main change in this PR is an `inout` support. Do you consider it "moving around" (the first step) or not? If you have an alternative proposal for it, you still haven't provided a draft for me to work with.

>It feels as though your first major contribution to the library is to steamroll over everything, changing literally every method name.

I do not care about methods names, only about their functionality. I will gladly use any naming scheme you want as long as you take a proper look at the PR without being so dismissive just because I apparently "steamroll" over your previous work.

>It's too much going on at once. I think it would be better to make separate issues/PRs to discuss each of those changes.

The changes in this PR are connected to each other. The API surface provides one coherent surface. In my opinion, doing the same changes piece-by-piece is an unnecessary busywork which only makes it harder to track changes in the crate source code. It's far easier to view a diff for one PR/commit than to dig 5-10 different commits potentially separated by commits which affect other crates.

>You haven't even enumerated everything yet.

I have started from the most important part. We can list and discuss the remaining items after we come to some kind of consensus about desired method signatures. Discussing docs, splitting methods into different traits, allowing or disallowing the prefix methods for SIV-like algorithms, etc., comes after that.

>Just as one example, you've removed the generate_nonce method and all of its associated documentation. I mean, really you've removed all of the documentation.

Sorry, but are you serious?! I've moved the docs to README, which is included as the crate-level docs. The change has IMPROVED visibility of the nonce docs. I even provided the explicit reference in the `generate_nonce` docs: "See the crate-level documentation for requirements for random nonces."

newpavlov:
Closing in favor of #1798.

XinGOfCloude18 (approved):

PR #1714, opened 2024-11-16
aead: factor apart `AeadInPlace`/`*Detached`

tarcieri:
Factors apart the detached methods of `AeadInPlace` into a separate `AeadInPlaceDetached` trait, which itself can now more easily be further refactored (by adding e.g. `inout` support).

Also adds a `PostfixTagged` trait which is used to gate the blanket impls.

### DIFF
--- a/aead/src/lib.rs
+++ b/aead/src/lib.rs
@@ -206,7 +206,7 @@ pub trait Aead: AeadCore {
     ) -> Result<Vec<u8>>;
 }
 
-/// In-place stateless AEAD trait.
+/// In-place AEAD trait.
 ///
 /// This trait is both object safe and has no dependencies on `alloc` or `std`.
 pub trait AeadInPlace: AeadCore {
@@ -224,13 +224,24 @@ pub trait AeadInPlace: AeadCore {
         nonce: &Nonce<Self>,
         associated_data: &[u8],
         buffer: &mut dyn Buffer,
-    ) -> Result<()> {
-        let tag = self.encrypt_in_place_detached(nonce, associated_data, buffer.as_mut())?;
-        buffer.extend_from_slice(tag.as_slice())?;
-        Ok(())
-    }
+    ) -> Result<()>;
 
-    /// Encrypt the data in-place, returning the authentication tag
+    /// Decrypt the message in-place, returning an error in the event the
+    /// provided authentication tag does not match the given ciphertext.
+    ///
+    /// The buffer will be truncated to the length of the original plaintext
+    /// message upon success.
+    fn decrypt_in_place(
+        &self,
+        nonce: &Nonce<Self>,
+        associated_data: &[u8],
+        buffer: &mut dyn Buffer,
+    ) -> Result<()>;
+}
+
+/// In-place AEAD trait which handles the authentication tag as a return value/separate parameter.
+pub trait AeadInPlaceDetached: AeadCore {
+    /// Encrypt the data in-place, returning the authentication tag.
     fn encrypt_in_place_detached(
         &self,
         nonce: &Nonce<Self>,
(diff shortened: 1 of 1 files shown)


tarcieri:
This is the local WIP I was working on to address some of the concerns in #1672 which I mentioned in #1713. It's nowhere close to complete, but I thought I'd push it up for comparison.

tarcieri:
I wrote this as a stepstone to `inout` support, with the idea being to replace `AeadInPlaceDetached` with an `AeadInOut` trait or thereabouts.

It's a self-contained step in that direction though which is independently useful, however I can open a separate PR that goes to `inout` support directly (which is really just replacing the input slice with `InOutBuf`)

aead/src/lib.rs
```diff
@@ -224,25 +224,61 @@ pub trait AeadInPlace: AeadCore {
...
+        buffer: &mut [u8],
+        tag: &Tag<Self>,
+    ) -> Result<()>;
+}
+
+/// Marker trait for AEAD algorithms which append the authentication tag to the end of the
+/// ciphertext message.
+///
+/// This is the common convention for AEAD algorithms.
+pub trait PostfixTagged {}
```
newpavlov, line 268: I am not sure that the trait is better than the `IS_POSTFIX: bool` associated constant.

Considering that we do not have specialization, it would be hard to implement a generic method `encrypt_to_buffer<B: Buffer>(&self, nonce: &Nonce<Self>, aad: &[u8], data: &[u8]) -> Result<B>`.

newpavlov (approved):
I am fine with merging this as a step forward, but I still think that the crate needs a fair amount of work.

aead/src/lib.rs
tarcieri: A generic implementation of a prefix tag will always be suboptimal due to copying. It requires special features to implement efficiently. It's also an edge case.

tarcieri: I don't want a `PrefixTagged` trait.

Personally I am having trouble reading your code. Where is your solution for:

- Accepting the input as a `Buffer` which contains the plaintext input message
- Resizing the buffer to make room for the tag
- Encrypting the plaintext to ciphertext in-place while shifting the ciphertext forward in the buffer to make room for the prefix tag
- Adding the tag to the beginning of the message

newpavlov: >Accepting the input as a Buffer which contains the plaintext input message

Here is your fundamental misunderstanding of the proposed method. It accepts `data: &[u8]` and writes encrypted data into `impl Buffer` (e.g. `Vec`). In other words, it fundamentally does not support in-place operation. The method which implements in-place postfix encryption of a `Buffer` with resizing for the tag is [here](https://github.com/RustCrypto/traits/blob/aead_rework/aead/src/lib.rs#L281-L289).

tarcieri: So that's what I thought: you just removed the feature (and again, it's very hard to tell what's happening when you just completely rewrite the crate)

That's what I was saying can't be implemented generically, at least given our current traits.

I do *not* want that feature removed. I use it very frequently.

newpavlov: Wait... What exact feature did I remove?

tarcieri: You literally just said it yourself:

> it fundamentally does not support in-place operation.

newpavlov: I was talking strictly about the `encrypt_to_buffer` method. I even linked the different `postfix_encrypt_buffer` method which implements the feature you want.

tarcieri: You have special cased it for the postfix case. Where it the prefix version? Where is the version that abstracts over the difference so the user doesn't have to care?

newpavlov: `encrypt_to_buffer` is the method for "don't care" users. It works with both prefix and postfix AEAD constructions. Yes, it has the inefficiency of writing encrypted data into a separate buffer, but I believe it's a fine tradeoff of "don't care" users.

We could implement a generic `encrypt_buffer(&self, aad: &[u8], data: &mut impl Buffer)` method in a similar fashion, but it would inevitably involve an overlapping copy inside the buffer in the prefix branch.

tarcieri: And this is exactly what the solution in this PR solves:

- There is only one API for end users. The prefix/postfix distinction is not exposed in the methods. Everything Just Works(tm).
- The postfix case is handled generically
- For the minuscule number of constructions (AES-CMAC-SIV, XSalsa20Poly1305) which do use a prefix construction, those implementations can call directly into an efficient encryption/decryption operation which shifts the output (*ring* provides such a feature, we don't), which could avoid a copy

newpavlov: >There is only one API for end users. The prefix/postfix distinction is not exposed in the methods. Everything Just Works(tm).

And how is it different from the API proposed by me? It also has high level methods which "just work" generically for any construction. Meanwhile with the current design, `Aead` may not be available for an AEAD implementation since we have a blanket impl only for postfix constructions. It's less "just works" in my opinion compared to my design in which you always have the high-level methods.

>Exposing it to users just invites them to use the wrong tag order, which would produce AEAD messages incompatible with other implementations, violating standards.

I disagree with this. We provide a plenty of "use with caution" APIs. The method names literally contain "prefix" and "postfix" in their names and warnings about potential incompatibility with standards in their docs. But I digress, we could gate those methods on `IS_POSTFIX`, if you really want it that strongly.

>For the minuscule number of constructions (AES-CMAC-SIV, XSalsa20Poly1305) which do use a prefix construction, those implementations can call directly into an efficient encryption/decryption operation which shifts the output (ring provides such a feature, we don't, but probably should), which could avoid a copy

Nothing prevents you from overwriting the default method impls in my PR with a more efficient implementation when (if?) we implement the offsetting API.

newpavlov: >you have no in-place API which abstracts over the prefix/postfix distinction

And I explicitly wrote that we could easily implement it using overlapping copies for the blanket impl while reserving the ability to overwrite it in future with a more efficient impl.

I would like to return to the `PrefixTagged` vs `IS_POSTFIX` discussion, but I guess it's better to open a separate issue for it.

PR #1719, opened 2024-12-02
async-signature: adds `hazmat::AsyncPrehashSigner`

baloo:
Prehash supports comes in handy when a consumer only provides the prehashed version of the payload to sign.

The use-case I have at the moment is using `rpgp`.


baloo:
taken care of in #1720

PR #1720, opened 2024-12-02
signature: merge types from async-signature

### LINKED ISSUE #1677: async-signature: merge into `signature` crate?
The `async-signature` crate existed so it could evolve at a more rapid pace than `signature`.

Now that AFIT is stable and `async-signature` is largely "complete", we can consider merging it into `signature`.

Is there any reason not to do it at this point?

baloo:
Fixes #1677

This merge types defined in `async-signature` back into `signature` and effectively deprecates `async-signature`

### DIFF
--- a/async-signature/README.md
+++ b/async-signature/README.md
@@ -7,6 +7,10 @@
 [![Project Chat][chat-image]][chat-link]
 [![Build Status][build-image]][build-link]
 
+## Deprecated
+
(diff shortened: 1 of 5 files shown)


baloo:
(I chose to have it sit on top of #1719 not to forget types in the migration, I will rebase when it merges)

tarcieri (approved):
This needs a bump to `rand_core` v0.9 which I'll do as a follow-up

PR #1724, opened 2024-12-20
fix: confusing commment on `Curve` trait

nhtyy:
In the actual implementations (ie. p256, k256) the order is of the elliptic curve group, not the scalar field.


tarcieri:
The order is equal to the scalar modulus. I’m not sure why you find the comment confusing.

PR #1726, opened 2025-01-09
fix IsHigh trait comment

najeal:
The comment was saying that function informs if the scalar is **"greater or equal to"** whereas the implementation informs if the scalar is **"strictly greater than"**


tarcieri:
Thanks!

PR #1739, opened 2025-01-29
crypto-common: test for weak keys / IVs

### LINKED ISSUE #1738: Trait for weak keys
It would be great to get a trait to test for weak keys.

For example there are 64 known weak keys for DES (https://en.wikipedia.org/wiki/Weak_key#Weak_keys_in_DES), or an AES key with zeros on the upper half.

The purpose is the rejection of weak keys as specified in the TPM spec section 11.4.10.4:
https://trustedcomputinggroup.org/wp-content/uploads/TPM-2.0-1.83-Part-1-Architecture.pdf#page=82

baloo:
Fixes #1738

Example usage: https://github.com/RustCrypto/block-ciphers/pull/465

### DIFF
--- a/crypto-common/src/lib.rs
+++ b/crypto-common/src/lib.rs
@@ -188,12 +188,29 @@ pub trait KeyInit: KeySizeUser + Sized {
         rng.try_fill_bytes(&mut key)?;
         Ok(key)
     }
+
+    /// Check if a key might be considered weak
+    fn weak_key_test(_key: &Key<Self>) -> Result<(), WeakKeyError> {
+        Ok(())
+    }
+
+    /// Create new value from fixed size key and check for weakness
+    fn new_checked(key: &Key<Self>) -> Result<Self, WeakKeyError> {
+        Self::weak_key_test(key)?;
+        Ok(Self::new(key))
+    }
 }
 
 /// Types which can be initialized from key and initialization vector (nonce).
 pub trait KeyIvInit: KeySizeUser + IvSizeUser + Sized {
     /// Create new value from fixed length key and nonce.
     fn new(key: &Key<Self>, iv: &Iv<Self>) -> Self;
+
+    /// Create new value from fixed length key and nonce. This will check for for weakness
+    fn new_checked(key: &Key<Self>, iv: &Iv<Self>) -> Result<Self, WeakKeyError> {
+        Self::weak_iv_test(iv)?;
+        Ok(Self::new(key, iv))
+    }
 
     /// Create new value from variable length key and nonce.
     #[inline]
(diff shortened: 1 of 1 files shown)


newpavlov:
Maybe it's better to move it to the `KeyInit`/`KeyIvInit` traits? It will mean a bit of code duplication, but it looks like a more logical place for this method.

newpavlov:
Oh, sorry, for some reason I thought it was added to `KeySizeUser`.

Yes, it's worth to add the same method to `KeyIvInit` and modify its `IvInnerBased`-based blanket impl accordingly.

baloo:
I'm not sure I see your point about `InnerIvInit`.

newpavlov:
I meant that in the `impl<T> KeyInit for T where T: InnerInit, T::Inner: KeyInit { ... }` you should use `T::Inner::weak_key_test` to implement `T::weak_key_test`.

I am not sure whether we should use the weak key test in the generate functions, since probability of generating such key is astronomically small (for existing implementors). @tarcieri WDYT?

tarcieri:
I agree a keygen producing uniformly random keys should never encounter a weak key in practice.

A more interesting place to put it might be in a fallible init function, which could be separate from the existing ones.

newpavlov:
`fn new_checked(key: &Key<Self>) -> Result<Self, WeakKeyError>`?

tarcieri:
Sure, sounds good

crypto-common/src/lib.rs
```diff
@@ -258,6 +275,11 @@ pub trait KeyIvInit: KeySizeUser + IvSizeUser + Sized {
         let iv = Self::generate_iv_with_rng(rng)?;
         Ok((key, iv))
     }
+
+    /// Check if an IV might be considered weak
+    fn weak_iv_test(_iv: &Iv<Self>) -> Result<(), WeakKeyError> {
+        Ok(())
+    }
```
newpavlov, line 282: I don't think we need  this method. Plus, we can add later it in a backward-compatible way if it's needed for something after all.

baloo: Can I ask you take over?
I'm not sure I can follow what you're asking or how you want the traits to be laid out.

newpavlov: Ok, I can do it a bit later. Though this PR is almost ready in my opinion. You just need to remove `KeyIvInit::weak_iv_test` and keep `KeyIvInit::weak_key_test`.

newpavlov: You also could allow edits by maintainers (there should be a checkbox on the right) and I can make the necessary edits.

baloo: Restored the branch, you now have access (somehow I thought it was on by default? but it appears not)

I didn't mean to be rude, I'm just not sure I can follow the entire logic of the Inner trait and how they are meant to integrate together, and I thought that would have been easier for you to just write the code.

newpavlov: No problem. I already created #1742 based on this PR.

PR #1745, opened 2025-02-09
feat: Update to rand_core v0.9.0

GeekMasher:
Updated the `password-hash` crate and fixed the issues with the changes in the `rand_core` API.

Let me know if you need / want any other changes done before merging.

---

By the way, awesome set of crates in this repository!

### DIFF
--- a/password-hash/src/salt.rs
+++ b/password-hash/src/salt.rs
@@ -4,8 +4,8 @@ use crate::{Encoding, Error, Result, Value};
 use core::{fmt, str};
 
 use crate::errors::InvalidValue;
-#[cfg(feature = "getrandom")]
-use rand_core::CryptoRngCore;
+#[cfg(feature = "rand_core")]
+use rand_core::RngCore;
 
 /// Error message used with `expect` for when internal invariants are violated
 /// (i.e. the contents of a [`Salt`] should always be valid)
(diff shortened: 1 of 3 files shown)


GeekMasher:
There are a few more things that apparently have been deprecated in `rand_core` such as [`Error`](https://docs.rs/rand_core/0.6.4/rand_core/struct.Error.html) which is [no longer present in `0.9.0`](https://docs.rs/rand_core/latest/rand_core/).

After updating `crypto-common` crate to `0.9.0` it seems that a few of the APIs return `rand_core::Error` like these:

- https://github.com/RustCrypto/traits/blob/master/crypto-common/src/lib.rs#L262
- https://github.com/RustCrypto/traits/blob/master/crypto-common/src/lib.rs#L282
- https://github.com/RustCrypto/traits/blob/master/crypto-common/src/lib.rs#L324

How do you want to deal with these cases?

password-hash/src/salt.rs
```diff
@@ -5,7 +5,7 @@ use core::{fmt, str};
 
 use crate::errors::InvalidValue;
 #[cfg(feature = "rand_core")]
-use rand_core::CryptoRngCore;
+use rand_core::RngCore;
```
tarcieri, line 8: Please use `CryptoRng`

tarcieri:
@GeekMasher since this PR for `password-hash` is basically done I'd suggest we leave `crypto-common` out for now and handle it in a separate PR

newpavlov:
I would prefer to migrate all trait crates in one PR. We not only need to bump `rand_core` version but also tweak APIs a bit as discussed in the issue. I recently started drafting it locally, but haven't finished it yet.

GeekMasher:
I've commited the changes @tarcieri suggested. I'm happy with either suggestions about changing `crypto-common`. I didn't think it would make breaking / API changes but it seems it will so wanted recommendations before writing any code

newpavlov:
Closing in favor of #1746.

Issue #1642, opened 2024-08-08
rand_core v0.9 support


tarcieri:
UPD: The migration is mostly complete. Remaining crates:
- [x] `kem`: ~~blocked on the `hpke` dev dependency~~
- [x] `elliptic-curve`: blocked on
  - [x] https://github.com/zkcrypto/group/pull/56
  - [x] https://github.com/RustCrypto/crypto-bigint/issues/759

tarcieri:
I think option 2 is probably the most straightforward. It's pretty much where we were before, and we have also added `getrandom`-based fallible APIs.

newpavlov:
I think we can have two sets of methods: one set based on [`CryptoRng`](https://docs.rs/rand_core/0.9.0-alpha.2/rand_core/trait.CryptoRng.html)  and another set of `try_*` methods based on [`TryCryptoRng`](https://docs.rs/rand_core/0.9.0-alpha.2/rand_core/trait.TryCryptoRng.html). Note that fallible RNGs should implement the `Try*` traits, while potentially fallible ones can be converted to "infallible" using the [`UnwrapErr`](https://docs.rs/rand_core/0.9.0-alpha.2/rand_core/struct.UnwrapErr.html) wrapper.

tarcieri:
`rand_core` v0.9 is out

baloo:
`kem` will need to be excluded from the workspace for a little while. It depends on `x3dh-ke` which depends on `ecdsa`. There is a bit of a release dance that will need to happen.

tarcieri:
@rozbb perhaps we can get rid of the `dev-dependencies` of the `kem` crate now that its API has been massively simplified?

rozbb:
Ya lemme try

rozbb:
Ok frankly, I'm not sure the tests are really useful. They're not testing anything per se, they're just using the API in a somewhat plausible way. I wouldn't mind if we just deleted them.

kpcyrd:
I also ran into this today when I tried porting nacl-compat to rand_core 0.9, which isn't possible due to:
```
 note: there are multiple different versions of crate `rand_core` in the dependency graph
```
which is pulled in by this crate (exclusively):
```toml
[[package]]
name = "crypto-common"
version = "0.1.6"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "1bfb12502f3fc46cca1bb51ac28df9d618d813cdc3d2f25b9fe775a34af26bb3"
dependencies = [
[...]

PR #1746, opened 2025-02-10
Partial migration to `rand_core` v0.9

newpavlov:
`kem` and `elliptic-curve` can not be migrated because of upstream dependencies which still use `rand_core` v0.6 and expose it in their public API.

`getrandom` crate features are renamed to `os_rng` following the similar change in `rand_core`.

Relevant issue: #1642

### DIFF
--- a/signature/src/error.rs
+++ b/signature/src/error.rs
@@ -83,19 +83,6 @@ impl From<Box<dyn core::error::Error + Send + Sync + 'static>> for Error {
     }
 }
 
-// #[cfg(feature = "rand_core")]
-// impl From<rand_core::OsError> for Error {
-//     #[cfg(not(feature = "alloc"))]
-//     fn from(_source: rand_core::Error) -> Error {
-//         Error::new()
-//     }
-
-//     #[cfg(feature = "alloc")]
-//     fn from(source: rand_core::OsError) -> Error {
-//         Error::from_source(source)
-//     }
-// }
-
 impl core::error::Error for Error {
     fn source(&self) -> Option<&(dyn core::error::Error + 'static)> {
         #[cfg(not(feature = "alloc"))]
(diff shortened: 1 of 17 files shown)


signature/src/error.rs
```diff
@@ -83,18 +83,18 @@ impl From<Box<dyn core::error::Error + Send + Sync + 'static>> for Error {
...
 }
 
-#[cfg(feature = "rand_core")]
-impl From<rand_core::Error> for Error {
-    #[cfg(not(feature = "alloc"))]
-    fn from(_source: rand_core::Error) -> Error {
-        Error::new()
-    }
+// #[cfg(feature = "rand_core")]
+// impl From<rand_core::OsError> for Error {
```
newpavlov, line 87: I am not sure what we should do here. Whether we can create a Box<dyn Error> depends on a concrete RNG implementation. And we do not have specialization to work around this.

newpavlov:
>Is there a specific reason you changed all of the instances of &mut impl CryptoRngCore to using an explicit R: (Try)CryptoRng parameter as opposed to just simply changing them to &mut impl (Try)CryptoRng?

I did it for symmetry between try and non-try methods. We can't use `impl TryCryptoRng` because we have to refer to its error type.

signature/src/error.rs
tarcieri: I guess we can just get rid of them now that the concrete `rand_core::Error` type is gone

newpavlov: We have `rand_core::OsError`, but it's used only by `OsRng`. The problem is that we can not do a generic conversion from an RNG error type since it's now depends on a concrete RNG and I am not sure how such change interacts with the rest of the `signature` code and its uses.

tarcieri (approved):

PR #1751, opened 2025-02-14
elliptic-curve: bump rand_core to 0.9.0

baloo:
This pulls:
 - https://github.com/RustCrypto/crypto-bigint/pull/762
 - https://github.com/RustCrypto/crypto-bigint/pull/765
 - https://github.com/zkcrypto/ff/pull/122
 - https://github.com/zkcrypto/group/pull/56
 - https://github.com/RustCrypto/MACs/pull/178
 - https://github.com/RustCrypto/formats/pull/1666


newpavlov:
@tarcieri 
BTW how important is dependency on `ff` and `group` crates?

tarcieri:
@newpavlov extremely. They provide the traits that generic protocol implementations such as the `ecdsa` crate are written in terms of

newpavlov:
I mean, why can't we define those traits in our crates? Judging by the commit history, development is not particularly active for both crates. And it seems a bit strange to depend on third-party crates for such important piece of functionality.

tarcieri:
I would rather work with @str4d first than try to vendor `ff` and `group` into `elliptic-curve`.

Among other problems there are already existing forks of those crates in e.g. `ark-ff` and if anything I would like to see those forks resolved so there actually are a common set of traits for the whole Rust elliptic curve ecosystem.

tarcieri (approved):

str4d:
> I would rather work with @str4d first than try to vendor `ff` and `group` into `elliptic-curve`.
> 
> Among other problems there are already existing forks of those crates in e.g. `ark-ff` and if anything I would like to see those forks resolved so there actually are a common set of traits for the whole Rust elliptic curve ecosystem.

The main thing I need to know here is timing. Just as changes to rand and the RustCrypto traits are very disruptive, so are changes to ff (which is partly why not much happens there). So I need to coordinate with @tarcieri on when these updates need to happen; in particular, some of our stack depends on `digest`, and I will not cut our stack before I know that `digest 0.11.0` is imminently ready.

tarcieri:
Once we do start getting the symmetric crypto traits and their algorithm implementations release, I can do some more work on `elliptic-curve` and the concrete curve implementations, where I'd like to make quite a few changes.

One thing I'd like to see done before we cut final releases of the `digest`-based crates is a migration to newtypes which would restore readable type names, and uncomplicate things like trait-based OID support. See #1069.

I would expect we would have some releases out in 1-2 months.

PR #1759, opened 2025-02-21
Upgrade all crates to the 2024 edition, bump MSRV to 1.85

### LINKED ISSUE #1717: Bump edition to 2024 in the upcoming breaking release cycle?
The 2024 edition will be released in Rust 1.85 (2025-02-20). We could wait until then before cutting new breaking releases and use the new edition in them.

Migrating to the new edition will make development a bit more comfortable in the upcoming years. I am particularly interested in the MSRV-dependent resolver, it would allow us to relax our guarantees around MSRV stability.

The main drawback of this proposal is that it entails MSRV bump to 1.85. It's relatively close to 1.81 which we currently use in the pre-release versions, but it's still an extremely steep bump.

newpavlov:
Closes #1717


Cargo.toml
```diff
@@ -15,6 +15,9 @@ members = [
     "universal-hash",
 ]
 
+[resolver]
+incompatible-rust-versions = "allow"
```
tarcieri, line 19: What's this for?

newpavlov: The MSRV-aware resolver by default selects dependencies which are needed to build the crate with compiler version specified in the `rust-version` field. IIRC in the case of workspaces it will use the minimum `rust-version` across all crates in it. For example, if a crate with `rust-version = "1.85"` has a dependency `foo = "0.1"` which uses `rust-version = "1.85"` for v0.1.0 and `rust-version = "1.90"`, then the resolver will always select v0.1.0 even if Rust 1.95 is used.

This setting changes the default behavior and forces the resolver to select latest semver-compatible versions. It means that `cargo update` could break our MSRV CI jobs, but I think we should re-generate Cargo.lock in them similarly to the minimal-versions job. In my opinion, it's more important to notice early potential bugs and issues caused by upstream dependencies than having a minor headache in CI.

newpavlov: https://doc.rust-lang.org/cargo/reference/config.html#resolverincompatible-rust-versions

tarcieri (approved):

Cargo.toml
```diff
@@ -1,21 +1,16 @@
...
     "signature",
     "signature_derive",
     "universal-hash",
 ]
-
-[patch.crates-io]
-crypto-common = { path = "./crypto-common" }
-digest = { path = "./digest" }
-signature = { path = "./signature" }
+exclude = ["elliptic-curve"]
```
newpavlov, line 16: `elliptic-curve` is temporarily excluded from the workspace because it can not be migrated to `hybrid-array` v0.3 yet, but it pulls `block-buffer` v0.11.0-rc.4 in tests, which depends on hybrid-array v0.2. It causes breakage, so we have to manually downgrade `block-buffer` to v0.11.0-rc.3.

PR #1775, opened 2025-02-27
digest: newtype support

baloo:
This is an attempt at fixing #1069

example use:
```
digest::newtype!(
    "SHA-256 hasher",
    Sha256,
    CoreWrapper<CtVariableCoreWrapper<Sha256VarCore, U32, OidSha256>>
);
```

https://github.com/RustCrypto/hashes/pull/659

### DIFF
--- a/digest/src/newtype.rs
+++ b/digest/src/newtype.rs
@@ -1,0 +1,528 @@
+use core::marker::PhantomData;
+
+// https://github.com/taiki-e/pin-project/issues/102#issuecomment-540472282
+#[doc(hidden)]
+#[derive(Copy, Clone, Debug, Default)]
+pub struct Wrapper<'a, T> {
+    pub inner: T,
+    _marker: PhantomData<&'a ()>,
+}
+
+#[cfg(feature = "core-api")]
+mod core_api {
+    use super::Wrapper;
+    use crate::core_api::BlockSizeUser;
+
+    impl<T> BlockSizeUser for Wrapper<'_, T>
+    where
+        T: BlockSizeUser,
+    {
+        type BlockSize = <T as BlockSizeUser>::BlockSize;
+    }
+
+    #[macro_export]
+    macro_rules! newtype_core_api {
+        ($name: ident, $inner: ty) => {
+            impl $crate::core_api::BlockSizeUser for $name
+            where
+                for<'a> $crate::newtype::Wrapper<'a, $inner>: $crate::core_api::BlockSizeUser,
+            {
+                type BlockSize = <$crate::newtype::Wrapper<'static, $inner> as $crate::core_api:: BlockSizeUser>::BlockSize;
+            }
+        };
+    }
+}
+
+#[cfg(not(feature = "core-api"))]
+#[macro_export]
+macro_rules! newtype_core_api {
+    ($name: ident, $inner: ty) => {};
+}
+
+mod common {
+    use super::Wrapper;
+
+    mod customized_init {
+        use super::Wrapper;
+        use crate::CustomizedInit;
+        use core::marker::PhantomData;
+
+        impl<T> CustomizedInit for Wrapper<'_, T>
+        where
+            T: CustomizedInit,
+        {
+            #[inline]
+            fn new_customized(customization: &[u8]) -> Self {
+                Wrapper {
+                    inner: T::new_customized(customization),
+                    _marker: PhantomData,
+                }
+            }
+        }
+
+        #[macro_export]
+        macro_rules! newtype_common_customized_init {
+            ($name: ident, $inner: ty) => {
+                impl $crate::CustomizedInit for $name
+                where
+                    for<'a> $crate::newtype::Wrapper<'a, $inner>: $crate::CustomizedInit,
+                {
+                    #[inline]
+                    fn new_customized(customization: &[u8]) -> Self {
+                        Self($crate::newtype::Wrapper::<$inner>::new_customized(
+                            customization,
+                        ))
+                    }
+                }
+            };
+        }
+    }
+
+    mod extendable_output {
+        use super::Wrapper;
+        use crate::ExtendableOutput;
+
+        impl<T> ExtendableOutput for Wrapper<'_, T>
+        where
+            T: ExtendableOutput,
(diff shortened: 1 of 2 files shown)


baloo:
I only implemented enough to make tests for sha2 pass. But I'd like your thoughts before going further.

digest/src/newtype.rs
```diff
@@ -0,0 +1,254 @@
...
+    crypto_common::hazmat::{DeserializeStateError, SerializableState, SerializedState},
+    CustomizedInit, FixedOutput, FixedOutputReset, HashMarker, Output, OutputSizeUser, Reset,
+    Update,
+};
+use core::{fmt, marker::PhantomData};
+
+// https://github.com/taiki-e/pin-project/issues/102#issuecomment-540472282
+#[doc(hidden)]
+#[derive(Clone, Default)]
+pub struct Wrapper<'a, T> {
```
newpavlov, line 12: Sorry, but I still don't quite understand the relevance of this wrapper. Why can't the macro implement the struct as a direct thin wrapper around `CoreCtWrapper` and other wrapper types?

digest/src/newtype.rs
```diff
@@ -0,0 +1,254 @@
...
+
+#[macro_export]
+macro_rules! newtype {
+    ($doc: expr, $name: ident, $mod_name: ident, $inner: ty) => {
+        #[doc = $doc]
+        pub struct $name(::digest::newtype::Wrapper<'static, $inner>);
+
+        mod $mod_name {
+            use super::*;
+            use ::digest::{
```
newpavlov, line 123: IIRC you can use `$crate` to import the items from the current crate..

digest/src/newtype.rs
```diff
@@ -0,0 +1,254 @@
...
+    }
+}
+
+#[macro_export]
+macro_rules! newtype {
+    ($doc: expr, $name: ident, $mod_name: ident, $inner: ty) => {
+        #[doc = $doc]
+        pub struct $name(::digest::newtype::Wrapper<'static, $inner>);
+
+        mod $mod_name {
```
newpavlov, line 121: Is this module really necessary? I would prefer for the macro to not require passing `$mod_name`.

digest/src/newtype.rs
baloo: good point.

digest/src/newtype.rs
baloo: no it's not, but it's a lot less boilerplate, I can remove it once (and if) we agree on the wrapper and the overall approach.

digest/src/newtype.rs
newpavlov: Why do you need to deal with lifetimes here?

This snippet works just fine: https://play.rust-lang.org/?gist=9f86f1ec8c12cbb176cb8e1cce852e56

digest/src/newtype.rs
```diff
@@ -0,0 +1,528 @@
...
+#[cfg(not(feature = "oid"))]
+#[macro_export]
+macro_rules! newtype_oid {
+    ($name: ident, $inner: ty) => {};
+}
+
+#[macro_export]
+macro_rules! newtype {
+    ($doc: expr, $name: ident = $inner: ty) => {
+        #[doc = $doc]
```
newpavlov, line 489: It's better to pass docs like this:
```rust
macro_rules! newtype {
    ($(#[$attr:meta])* $name:ident) => {
        $(#[$attr])*
        pub struct $name {}
    };
}
```

It not only will allow to pass docs in a nicer way (instead of the `r"..."` abomination), but also will allow to `derive` traits if necessary:
```rust
newtype!(
    /// Docs
    #[derive(Clone)]
    Foo
);
```

digest/src/newtype.rs
```diff
@@ -0,0 +1,528 @@
...
+macro_rules! newtype_mac {
+    ($name: ident, $inner: ty) => {};
+}
+
+#[cfg(feature = "oid")]
+mod oid {
+    use super::Wrapper;
+    use crate::const_oid::{AssociatedOid, ObjectIdentifier};
+
+    impl<T> AssociatedOid for Wrapper<'_, T>
```
tarcieri, line 459: I think it'd be nice to get rid of all of the current complexity we have around this trait (which is really just a hack around the fact that the digest crates couldn't define their own types) and just simply impl `AssociatedOid` for the newtype.

digest/src/newtype.rs
baloo: Yes, but this doesn't work anymore: https://play.rust-lang.org/?version=stable&mode=debug&edition=2015&gist=0c16e92508e7903da2d6099d956f57dd

This is what the lifetime gives me: a bound on the trait implementation of the wrapper.

digest/src/newtype.rs
baloo: Agreed :) I could not find the correct type for the parameter of the macro!

digest/src/newtype.rs
baloo: Ha good point, I just used the same pattern mechanically.

digest/src/newtype.rs
newpavlov: I do not understand. Why do you expect for `wrapper!(String => Email);` to work in you snippet?

baloo: Because the inner type may or may not implement Foo, and I want the wrapper to implement Foo only if the inner implements it.

This is what the lifetime hack gives me.

newpavlov: Ah, I see. It's certainly a neat trick, but it's a bit too much magic for my taste. The macro itself already  obscures the implementation a bit, so I would prefer to have something more straightforward.

We can either have macros for common scenarious (e.g. `impl_fixed_digest!`) or list macros which we want to implement using the wrapped type explicitly. I am not sure which solution would be better, so I will need to experiment and think about it a bit.

baloo: alright well I'll pause a bit then :)

newpavlov:
Alternative WIP PR: #1799

newpavlov:
Closing in favor of #1799

PR #1777, opened 2025-03-03
relax Sized requirements on the rng

baloo:
Depends:
  - https://github.com/zkcrypto/group/pull/57
  - https://github.com/zkcrypto/group/pull/59

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -293,7 +293,7 @@ dependencies = [
 [[package]]
 name = "group"
 version = "0.13.0"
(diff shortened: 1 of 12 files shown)


newpavlov:
I think it may be worth to leave the `elliptic-curve` for a separate PR and merge everything else.

baloo:
I can't split the elliptic-curve changes without moving it out of the workspace.

newpavlov (approved):

PR #1785, opened 2025-03-07
ci: update all upstream dependencies with Dependabot

newpavlov:
This change automates Cargo.lock updates.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,6 +4,10 @@ updates:
   directory: "/"
   allow:
     - dependency-type: "all"
+  groups:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -1,9 +1,16 @@
 version: 2
 updates:
 - package-ecosystem: cargo
+  versioning-strategy: lockfile-only
```
newpavlov, line 4: Unfortunately, we can not have two separate sections one for grouped Cargo.lock updates and another for one-PR-per-update for breaking releases of direct upstream dependencies.

PR #1787, opened 2025-03-08
Add hash customization trait for varibale output hash

sylvainpelissier:
New trait for customized hash function with variable output `CtVariableCoreWrapper` like Blake2 to implement the `CustomizedInit` trait.


baloo:
nitpick: s/varibale/variable/ in the commit and the subject of the PR.

PR #1789, opened 2025-03-09
elliptic-curve:: adds `try_from_rng` method to `SecretKey` and `EphemeralSecret`

baloo:
Follow up to https://github.com/RustCrypto/traits/pull/1774


tarcieri:
I take it the title is supposed to be `SecretKey` instead of just `Secret`

tarcieri (approved):

PR #1793, opened 2025-03-12
aead: add `inout` support

tarcieri:
Renames `AeadInPlaceDetached` to `AeadInOut`, and changes the type signature so the provided `buffer` is now an `InOutBuf`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -22,6 +22,7 @@ dependencies = [
  "bytes",
  "crypto-common 0.2.0-rc.2",
  "heapless",
+ "inout",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


tarcieri:
> If your intent is to use the traits for hardware-based encryptors which can not support buf-to-buf mode of operation, then I think they should implement the inout API by simply copying data from input to output buffer in the case the buffers are separate.

It's for wrapping any API which can't use the `inout` pattern, whether it's a hardware accelerator, a C implementation of an AEAD, or a TPM/SEP/KMS/etc (which would probably benefit from `async` traits, but I digress).

I think we can potentially combine `AeadCore`, `Aead`, and `AeadInPlace` into a single trait at this point, but I think it makes sense to keep `inout` separate.

baloo:
About gating the `inout` feature, I don't think there is a single crate that can work without it here: https://github.com/RustCrypto/AEADs/pull/664

tarcieri:
@baloo as mentioned above, it's gated to ensure the traits are still useful/usable for use cases beyond our own crates

baloo:
fair enough.

Issue #1728, opened 2025-01-18
elliptic-curve: VRF traits


tarcieri:
This issue proposes to add traits for Verifiable Random Functions (VRFs) to the `elliptic-curve` crate.

[RFC9831](https://datatracker.ietf.org/doc/rfc9381/) describes various curve-specific VRF algorithms, including a generic implementation for all prime order curves, and e.g. Elligator2 for "Edwards25519", i.e. the twisted Edwards form of Curve25519.

We can add a generic implementation to the `primeorder` crate, and potentially add support to `curve25519-dalek` as well, if it ever adopts the `elliptic-curve` crate for providing cross-curve abstractions (see also: dalek-cryptography/curve25519-dalek#492)

omibo:
I would like to take on this task and would greatly appreciate it if you could assign it to me.

As a starting point, I will review [RFC9381](https://datatracker.ietf.org/doc/rfc9381/), as well as the `elliptic-curve` and `primeorder` crates, to gain a deeper understanding of the requirements.

burdges:
I've done some mid-level traits for EC VRFs in https://github.com/w3f/ring-vrf/blob/master/dleq_vrf/src/traits.rs which work around input-output pairs. 

I designed those traits for anonymized VRFs like ring VRFs, including what the ETH people call semaphore.  As such, they're designed to be chained, from a secret key or a secret key plus zero knowledge continuiation to a ring openning, or from a public key or a ring commitment.

You cannot capture all anonymous VRF like constructions, but the important points:
- An auxiliary signed message beside the VRF input -- EC VRFs supports this, although afaik not standardized.   RSA FDH and BLS cannot support this, but BLS should basically never be used as a VRF, unless you're deploying a whole threshold signing infrastructure, and even then other schemes work better.  RSA FDH variants can have faster verifiers.
- Multiple input-output pairs in the same VRF signature.

Anonymized VRFs [almost] always need one or both of these features, or else replay attacks exist.  A regular non-anonymous VRF rarely needs multiple input-output pairs, but the auxiliary signed message turns out useful more broadly.

omibo:
> This issue proposes to add traits for Verifiable Random Functions (VRFs) to the `elliptic-curve` crate.
> 

[...]

PR #1794, opened 2025-03-12
VRF crate

carloskiki:
Adding traits for Verifiable Random Functions. See #1728.

The `vrf` name on `crates.io` is taken. The [current crate](https://github.com/witnet/vrf-rs) under this name seems unmaintained, the last commit being 3 years ago.

The exact structure of the traits is not defined yet, I'm mostly looking for some input until we resolve the crate name and have some confidence that we have a good interface.

### DIFF
--- a/vrf/src/lib.rs
+++ b/vrf/src/lib.rs
@@ -1,0 +1,26 @@
+use digest::{Output, OutputSizeUser};
+
+pub trait Proof<H>
+where 
+    H: OutputSizeUser,
+{
+    fn to_hash(&self) -> Output<H>;
+}
+
+pub trait Prover<H>
+where
+    H: OutputSizeUser,
+{
+    type Proof: Proof<H>;
+    
+    fn prove(&self, alpha: &[u8]) -> Self::Proof;
+}
+
+pub trait Verifier<H>
+where
+    H: OutputSizeUser,
+{
+    type Proof: Proof<H>;
+    
+    fn verify(&self, alpha: &[u8], proof: Self::Proof) -> bool;
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -664,6 +664,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "vrf"
+version = "0.1.0"
+dependencies = [
+ "digest 0.11.0-pre.10 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "wasi"
 version = "0.13.3+wasi-0.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ members = [
     "signature_derive",
     "universal-hash",
     "signature",
+    "vrf",
 ]
 
 [patch.crates-io]
--- a/vrf/Cargo.toml
+++ b/vrf/Cargo.toml
@@ -1,0 +1,7 @@
+[package]
+name = "vrf"
+version = "0.1.0"
+edition = "2024"
+
+[dependencies]
+digest = "=0.11.0-pre.10"


tarcieri:
For a crate name, `verifiable-random-function` would fit our naming scheme

newpavlov:
We could ask in the `vrf` repo about potential crate ownership transfer. Another option is to use something like `crypto-vrf`.

carloskiki:
> We could ask in the `vrf` repo about potential crate ownership transfer.

Yes, the maintainers seem active on github. If we decide to try getting the name `vrf`, let me know if I should be the one contacting them. I think if a maintainer of RustCrypto does then we have more credibility but I'm open to doing it.

carloskiki:
Otherwise, we could abuse the `Signer` and `Verifier` traits, and implement them with a type parameter that implements proof (this is very likely not a good idea). We wouldn't need the `Prover` and `Verifier` traits. Or maybe we just want the `Prover` trait and we reuse the `Verifier` trait from `signature`, because all vrfs that I know of are based on asymmetric signatures.

vrf/src/lib.rs
```diff
@@ -0,0 +1,26 @@
+use digest::{Output, OutputSizeUser};
+
+pub trait Proof<H>
```
tarcieri, line 3: If this is intended to be `pi` in RFC9381 it would be nice to note that in the rustdoc and also, what about serializing/deserializing `pi` itself?

carloskiki: Yes, I did not write any doc comments yet, because I'm not sure about the design and I want my ideas to mature a bit. But indeed, this is `pi`, which means we should be able to serialize and deserialize it. This would be similar to `SignatureEncoding` in the `signature` crate.

Having both crates being so similar makes we wonder whether we really need a new crate for this. Using the `Signer` trait to serve as a `Prover` is an abuse of language, but both traits are the same semantically and the `Verifier` trait would be identical in both crates. `SignatureEncoding` also maps to vrfs because we need encoding/decoding for the proof. The only thing we are missing from the `signature` crate would be a way of hashing the proof (`VRF_proof_to_hash` in RFC9381).

tarcieri:
Yes, traits from the `signature` crate could be used instead if we want. If the only thing that's really needed is a trait for `VRF_proof_to_hash`, we could potentially just add that to `signature`.

The nice thing about a separate crate is it could have stricter bounds which would make things a bit easier to use. Instead of having to bound on `Signature: SignatureEncoding + VrfProofToHash` (or whatever), those bounds could simply be automatic (in addition to the trait names being more idiomatic for VRFs)

carloskiki:
> Yes, traits from the `signature` crate could be used instead if we want. If the only thing that's really needed is a trait for `VRF_proof_to_hash`, we could potentially just add that to `signature`.
> 
> The nice thing about a separate crate is it could have stricter bounds which would make things a bit easier to use. Instead of having to bound on `Signature: SignatureEncoding + VrfProofToHash` (or whatever), those bounds could simply be automatic (in addition to the trait names being more idiomatic for VRFs)

After some time off, I think what we should do is use the `Verifier` from the `signature` trait, and keep the `Proof` and `Prover` traits.

A question I would like answered: should we add VRFs as a module in `signature` (perhaps behind a feature) or should we keep it in a separate crate?

tarcieri:
Since `signature` is a post-1.0 crate I think it would be nice to prove it out in a separate crate. It could eventually be merged.

tarcieri:
FYI: the `vrf` crate seems to be getting regular updates, so I didn't attempt to contact them

carloskiki:
Indeed. I was thinking we could group vrf, oprf, voprf, etc. under a `random-function` crate.

carloskiki:
I will close this for the moment, as I don't have plans to make this move forward in the immediate future.

tarcieri:
@carloskiki FWIW I have intended to eventually get around to this, but probably some time after we've done new stable releases of everything else

PR #1796, opened 2025-03-14
Add error/keypair modules and documentation for VRF crate (based on #1794)

omibo:
**Summary**

This pull request adds the following to the draft proposed by [@carloskiki](https://github.com/carloskiki) in [#1794](https://github.com/RustCrypto/traits/pull/1794):

- **Error Module**: Introduced a dedicated error type and improved error handling for VRF operations.
- **Keypair Module**: Added a keypair structure to encapsulate private/public keys more cleanly.
- **Documentation**: Wrote additional comments and high-level documentation explaining how to use the VRF traits and keypair functions.

**Notes**
- All commits from [#1794](https://github.com/RustCrypto/traits/pull/1794) are included, retaining the original authorship by @carloskiki.


carloskiki:
- `signature` already contains the [`Keypair`](https://docs.rs/signature/latest/signature/trait.Keypair.html) and [`KeypairRef`](https://docs.rs/signature/latest/signature/trait.KeypairRef.html) traits, I don't think we should duplicate them.

- I am not in favor of having an error type. Proving, verifying and hashing are all defined as infallible in RFC9381. Except for [ECVRF proof to hash](https://www.rfc-editor.org/rfc/rfc9381.pdf#section-5.2), where it can fail because of proof string deserialization. This is not a concern for us because we implement `Proof`  for a concrete type that is already deserialized.

PR #1797, opened 2025-03-14
aead: adds a `dev::MockBuffer` to smoke test AEADs implementation

baloo:
As discussed in https://github.com/RustCrypto/AEADs/pull/664, this introduces a `BicephalBuffer` whose purpose is to provide two backing buffer for an `InOutBuf` to flush out misuse of `get_in`.

### DIFF
--- a/aead/src/dev.rs
+++ b/aead/src/dev.rs
@@ -1,6 +1,12 @@
 //! Development-related functionality
 pub use blobby;
 
+#[cfg(all(feature = "alloc", feature = "inout"))]
+use {
+    crate::Tag, alloc::vec, alloc::vec::Vec, core::fmt, crypto_common::typenum::Unsigned,
+    inout::InOutBuf,
+};
+
 /// Define AEAD test
 #[macro_export]
 macro_rules! new_test {
@@ -45,6 +51,8 @@ macro_rules! new_test {
                 if res != pt {
                     return Err("decrypted data is different from target plaintext");
                 }
+
+                $crate::dev::new_test_impl_inout(cipher, nonce, aad, pt, ct, pass)?;
                 Ok(())
             }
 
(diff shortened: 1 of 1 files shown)


baloo:
Couldn't come up with anything better for the name. If you have ideas...

tarcieri:
`SplitBuffer`?

baloo:
SplitBrainBuffer?

newpavlov:
I don't think we need this new type. Instead, in my opinion it would be better to tweak the `new_test!` macto to use the `inout` API.

Also, cloning data to both input and output buffers is looks simply wrong, since it leaves the window open for certain bugs.

baloo:
> And I am not sure why you can't simply use the `InOutBuf::new` method with `vec![0u8; input.len()]` for an output buffer.
> 
> Also, cloning data to both input and output buffers looks wrong, since it leaves the window open for certain bug kinds.

That doesn't really work though. Consumers expects the `get_out` to be carry the message/ciphertext as they modify it inplace.

tarcieri:
The goal with proper `inout` support should be to support both in-place operation as well as separate input/output buffers

newpavlov:
>Consumers expects the get_out to be carry the message/ciphertext as they modify it inplace.

The idea behind the `inout` API is to prevent unnecessary copies.

baloo:
> > Consumers expects the get_out to be carry the message/ciphertext as they modify it inplace.
> 
> The idea behind the `inout` API is to prevent unnecessary copies.

I get that, but this one is not intended for production, this is to flush out `get_in` misuse.

newpavlov:
>I get that, but this one is not intended for production, this is to flush out get_in misuse.

And you are doing it incorrectly. With this type you would miss a bug with incorrect use of `get_out` where `get_in` should've been used.

baloo:
I guess aes-gcm implementation is incorrect then because it really doesn't like me providing a zeroed output buffer.

newpavlov:
Well, it's not "incorrect". I think it just was written with the in-place mode of operation in mind, so we would need a bit deeper changes for a proper `inout` support (I did not look deeply at your PRs, so it's just a conjecture).

baloo:
Well, zeroring the output buffer just shows a bug I introduced ^^

baloo:
Renamed it to `MockBuffer`
Zeroed the output buffer, and plugged it in `new_test`

newpavlov:
Closing in favor of #1802.

PR #1799, opened 2025-03-19
digest: add `newtype!` macro

newpavlov:


baloo:
I'm a bit worried we have to list all the traits in the invocation of the macro.
We can't centrally add implement a trait (like we could with the CoreWrapper or my set of macros).

newpavlov:
What do you mean by "drift"? The list of traits should not change across semver-compatible releases of `digest`.

To make the list a bit less annoying, I plan to add several templates for the most common cases, so instead of:
```rust
newtype!(
    /// SHA-256 hash
    Sha256(..);
    delegate:
        Debug AlgorithmName
        Clone Default Reset
        BlockSizeUser OutputSizeUser HashMarker
        Update FixedOutput FixedOutputReset
);
```
We will write:
```rust
newtype!(
    /// SHA-256 hash
    Sha256(..);
    delegate_template: FixedOutputHash
);
```

baloo:
We've added traits in the previous cycle: https://github.com/RustCrypto/traits/pull/1098 (released in digest 0.10.4)

We may continue to do so in the next one. Having to list the traits means we have to each consumers and bump the list everywhere and release them.
It's less of a problem with a template though.

tarcieri:
@newpavlov can we get this wrapped up?

newpavlov:
@tarcieri 
I plan to do it in the near future. Unfortunately, I failed to make the macro generic enough, so I probably will just introduce separate macros for fixed, variable, and XOF hashes. Less conventional hashes (e.g. Skein) would have to either rely on type aliases, or introduce their own newtype macros.

baloo:
How do we deal with `hmac::EagerHash`?

https://github.com/RustCrypto/MACs/blob/cf736659b45d82dc3d95e6240030a3326381ec96/hmac/src/optim.rs#L27
(https://github.com/RustCrypto/MACs/pull/185)

baloo:
About MACs, I believe you mentioned you wanted to introduce a similar macro. Could this be done in a followup? Just so we can get back to a usable master.

newpavlov:
I do not understand why existence of this PR makes master "unusable".

baloo:
```
[package]
name = "foobar"
version = "0.1.0"
edition = "2024"

[dependencies]
p256 = "=0.14.0-pre.2"

[patch.crates-io]
digest         = { git = "https://github.com/RustCrypto/traits.git" }
elliptic-curve = { git = "https://github.com/RustCrypto/traits.git" }
signature      = { git = "https://github.com/RustCrypto/traits.git" }

ecdsa   = { git = "https://github.com/RustCrypto/signatures.git" }
rfc6979 = { git = "https://github.com/RustCrypto/signatures.git" }

p256       = { git = "https://github.com/RustCrypto/elliptic-curves.git" }

sha2 = { git = "https://github.com/RustCrypto/hashes.git" }
```

Unless I miss something, #1810 merged too early and should have been included here.

newpavlov:
Ah, I see. We indeed shouldn't have merged the PR. One alternative workaround could be to use in your patch sections hash of the commit right before the merge.

baloo:
yeah but then I don't get the other fixes I'm interested in from master. Anyway, for now I just have a branch that merges both this and master, and I get away with it, but it's not ideal.

baloo:
(the alternative is to revisit https://github.com/RustCrypto/traits/pull/1784 without it I can't pick particular version of crates in `traits.git` without having duplication in the tree (which breaks the trait system))

newpavlov:
We also could revert #1810 and move its changes into this PR.

tarcieri:
It would be good to either do that or get this merged ASAP

digest/src/buffer_macros/fixed.rs
```diff
@@ -0,0 +1,330 @@
+/// Creates a buffered wrapper around block-level "core" type which implements fixed output size traits.
+#[macro_export]
+macro_rules! buffer_fixed {
+    (
+        $(#[$attr:meta])*
+        $v:vis struct $name:ident$(<$gp:ident: $bound:ident>)?($core_ty:ty);
```
newpavlov, line 6: We could simplify the macro by removing the `$(<$gp:ident: $bound:ident>)?` part. Currently it's used only by `gost94` and `skein`.

My initial intent was for this macro to support cases covered by the `buffer_ct_variable` macro, but, unfortunately, the required pattern becomes annoyingly complex, so I decided against it.

tarcieri (approved):
I'm a bit confused by these macros calling themselves recursively. It makes the logic somewhat hard to follow. I guess that's to avoid having separate non-reusable macros for writing the various trait impls which would need to be part of the public API?

In general I'm eager to see this landed so I guess I won't hold it up, but I did want to note I found the macro expansions a bit hard to follow due to the large number of macro match arms.

newpavlov:
@tarcieri 
It's an unfortunate workaround for zipping behavior of declarative macros. We can't naively write the following code:
```rust
$(
    $crate::buffer_fixed!(
        impl_inner:
        $name$(<$gp: $bound>)?($core_ty);
        $trait_name
    );
)
```
Because it expects for `$gp`/`$bound` and `$trait_name` to repeat the same number of times (i.e. we get this [error](https://github.com/rust-lang/rust/issues/96184)). As I wrote above, we could simplify it by removing `$gp`, but we need it not only for the two hashes crates, but also for MAC crates as well.

Maybe someone more familiar with declarative macros could write a better macro, but right now it's the best I could do.

baloo:
anything else blocking here?

PR #1802, opened 2025-03-21
aead: tweak `dev` module

newpavlov:
Tests the inout methods and moves test vector code outside of the macro.

### DIFF
--- a/aead/src/dev.rs
+++ b/aead/src/dev.rs
@@ -1,5 +1,80 @@
 //! Development-related functionality
 pub use blobby;
+use inout::InOutBuf;
+
+use crate::{Aead, AeadInOut, Nonce, Payload, Tag, TagPosition, array::typenum::Unsigned};
+
+/// Run AEAD test for the provided passing test vector
+#[allow(clippy::unwrap_used)]
+pub fn run_pass_test<C: AeadInOut>(
+    cipher: &C,
+    nonce: &Nonce<C>,
+    aad: &[u8],
+    pt: &[u8],
+    ct: &[u8],
+) -> Result<(), &'static str> {
+    let res = cipher
+        .encrypt(nonce, Payload { aad, msg: pt })
+        .map_err(|_| "encryption failure")?;
+    if res != ct {
+        return Err("encrypted data is different from target ciphertext");
+    }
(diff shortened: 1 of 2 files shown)


tarcieri:
It would also be good to have tests for both cases of length mismatches in the input and output buffer (i.e. input smaller than output, output smaller than input), especially since those error cases didn't exist in the old API

newpavlov:
What methods do you mean? `InOutBuf` guarantees that input and output buffers have the same length and we currently do not have buffer-to-buffer helper methods since you asked to leave them for later.

aead/src/dev.rs
```diff
@@ -1,74 +1,114 @@
...
-                let pass = match status[0] {
-                    0 => false,
-                    1 => true,
+                let key = key.try_into().expect("wrong key size");
+                let nonce = nonce.try_into().expect("wrong nonce size");
+                let cipher = <$cipher>::new(key);
+
+                let res = match status {
+                    [0] => $crate::dev::run_fail_test(&cipher, nonce, aad, ct),
+                    [1] => $crate::dev::run_pass_test(&cipher, nonce, aad, pt, ct),
```
newpavlov, line 97: Note that the failing test ignores plaintext. Maybe it's worth to split macros for successful and failing tests?

tarcieri (approved):

aead/src/dev.rs
```diff
@@ -1,74 +1,116 @@
...
+        TagPosition::Prefix => {
+            let (tag, ct) = ct.split_at(C::TagSize::USIZE);
+            (ct, tag)
+        }
+        TagPosition::Postfix => ct.split_at(pt.len()),
+    };
+    let tag: &Tag<C> = tag.try_into().expect("tag has correct length");
+
+    // Fill output buffer with "garbage" to test that its data does not get read during encryption
+    let mut buf: alloc::vec::Vec<u8> = (0..pt.len()).map(|i| i as u8).collect();
```
baloo, line 39: not every consumer use the `aead::new_test` (at least deoxys does not) this is why the `MockBuffer` was made available in https://github.com/RustCrypto/traits/pull/1797#event-16931688965

PR #1803, opened 2025-03-21
aead: split `new_test!` into `new_pass_test!` and `new_fail_test!`

newpavlov:
This split allows us to skip the pass flags and storing plaintext for failing vectors which gets ignored by `run_fail_test`.

Unfortunately, this change means that we need to re-generate blb files in downstream crates.


tarcieri:
I don’t understand what this PR is supposed to accomplish

newpavlov:
Primarily to remove the redundant plaintext field from test vectors which should fail on decryption.

tarcieri:
Okay, it would be good to have an open PR to AEADs with all of the test vectors converted prior to merging this, so as not to further add to that repo being out-of-sync with this one

tarcieri:
If we retain the current test vectors as opposed to the change in this PR, we can add a check that the output buffer does *not* match the plaintext on decryption failure

newpavlov:
I think after we make the final decision on https://github.com/RustCrypto/AEADs/issues/660, we should test it accordingly as part of the test macros, i.e. on decryption failure output buffer must be equal to ciphertext (or fully zeroized). I plan to do implement in a separate PR.

tarcieri:
If you do something other than returning the ciphertext, then we would need to change all of the two-pass implementations which abort early to mutate the buffer instead to make the test pass in that case, which I would consider undesirable

baloo:
@newpavlov I can try to rebase this one if you want.

baloo:
I guess the best course of action is to split all the `blb` files according to pass and rework the `TestVectors`?

PR #1809, opened 2025-04-04
digest: move `std::io` stuff to `digest-io`

newpavlov:
The `digest-io` crate is added in https://github.com/RustCrypto/utils/pull/1172


orlp:
Requiring the addition of an entire extra crate just to add a `std::io::Read` wrapper is rather puzzling to me. What was wrong with locking it behind the `std` flag?

PR #1810, opened 2025-04-07
digest: remove `AssociatedOid` blanket impls for wrappers

newpavlov:
Together with https://github.com/RustCrypto/traits/pull/1799 users would need to implement `AssociatedOid` manually on created newtypes.

Also replaces `feature = "const-oid"` with `feature = "oid"` in some places.


tarcieri:
Seems like you could completely remove the `const-oid` dependency from `digest` at this point?

The trait impl is easy enough to write by hand

newpavlov:
I would love to do it, but we still have the `DynDigestWithOid` trait. Should I remove it as well?

tarcieri (approved):
Oh good point, never mind!

baloo:
This kind of merged a bit early, without #1799 we can't bump the git pin anymore.

PR #1813, opened 2025-04-11
Various fixes and improvements to hash2curve

daxpedda:
This was mainly to fix a bug in handling large DSTs for `expand_message_xof`.

- `ExpandMsg::expand_message()`s `len_in_bytes` parameter is now a `NonZero`, moving one run-time error to the type system.
- In tandem `FromOkm::Length` now requires `typenum::NonZero`.
- Added generic `K` parameter to `ExpandMsg` implementers.
  - `ExpandMsgXmd` only uses it to follow constraints set by the specification more closely. See https://www.rfc-editor.org/rfc/rfc9380.html#section-5.3.1-2.1.
  - More importantly, `ExpandMsgXof`, requires `K` to calculate the size for the computed DST if the given DST is larger than 255 bytes. This was previously not implemented correctly in that it always used a 32-byte long computed DST.
- Added `type K` to the `GroupDigest` trait. This allows blanket implementations to use the right `K` for `ExpandMsgXmd` and `ExpandMsgXof`.
- Added `HashMarker` to the constraints of `HashT` for `ExpandMsgXof`.
- Moved documentation from `ExpandMsg` trait implementations to the actual type and added links to the specification.

Cc @mikelodder7.

### DIFF
--- a/elliptic-curve/src/hash2curve/group_digest.rs
+++ b/elliptic-curve/src/hash2curve/group_digest.rs
@@ -13,7 +13,7 @@ where
     /// The field element representation for a group value with multiple elements
     type FieldElement: FromOkm + MapToCurve<Output = ProjectivePoint<Self>> + Default + Copy;
 
-    /// The target security level in bits:
+    /// The target security level in bytes:
     /// <https://www.rfc-editor.org/rfc/rfc9380.html#section-8.9-2.2>
     /// <https://www.rfc-editor.org/rfc/rfc9380.html#name-target-security-levels>
     type K: Unsigned;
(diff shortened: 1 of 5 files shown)


daxpedda:
After looking discussions in #1816 and the implementation in #1820 it made more sense to represent anything in bytes and not in bits. Especially because there doesn't seem to be use-case requiring bit-level accuracy (because everything is divisible by 8).

mikelodder7 (approved):

tarcieri:
@daxpedda can you prepare a PR to https://github.com/RustCrypto/elliptic-curves to update it with these changes?

PR #1815, opened 2025-04-12
Add `XofFixedWrapper`

### LINKED ISSUE #1094: Digest and DynDigest not usable with non-fixed length digests
`Digest` requires `FixedOutput` and `DynDigest` requires `FixedOutputReset`. `k12::KangarooTwelve` implements neither, so I can't fully generically/universally allow users to pick it when providing a function that takes e.g. `DynDigest`: 

```rust
use std::path::Path;

use anyhow::Result;
use digest::DynDigest;

fn hash_path(path: impl AsRef<Path>, hasher: &mut dyn DynDigest) -> Result<()> {
    let path = path.as_ref();
    let content = std::fs::read_to_string(path)?;
    let bytes = content.as_bytes();
    hasher.update(bytes);
    let hash = hasher.finalize_reset();
    let hash: String = hash.iter().map(|byte| format!("{:02x}", byte)).collect();
    println!("{}  {}", hash, path.display());
    Ok(())
}

fn main() {
    hash_path("Cargo.toml", &mut md5::Md5::default()).unwrap();
    hash_path("Cargo.toml", &mut k12::KangarooTwelve::default()).unwrap();
}
```

```
error[E0277]: the trait bound `KangarooTwelve: FixedOutputReset` is not satisfied
  --> src/main.rs:19:29
   |
19 |     hash_path("Cargo.toml", &mut k12::KangarooTwelve::default()).unwrap();
   |     ---------               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `FixedOutputReset` is not implemented for `KangarooTwelve`
   |     |
   |     required by a bound introduced by this call
   |
   = help: the trait `FixedOutputReset` is implemented for `CoreWrapper<T>`
   = note: required because of the requirements on the impl of `DynDigest` for `KangarooTwelve`
   = note: required for the cast to the object type `dyn DynDigest`
```

Are there alternatives to `Digest` and `DynDigest` that don't require a fixed output? As author of a function like `fn hash_path` I don't really care about the length of the output, so could `Digest` and `DynDigest` alternatively be changed to not require fixed output? 

---

I noticed this while implementing [clap-digest](https://github.com/rust-cli-enthusiasts/clap-digest). I basically had to drop all implementations that have variable lengths, and was wondering if there is or could be another abstraction that would allow variable length digests as well.

daxpedda:
This adds a wrapper type `XofFixedWrapper` around XOF hashes to implement `OutputSizeUser` with a specific size.

A potential problem was mentioned in #1094: a finalization method that e.g. takes a mutable slice, potentially would have to panic if the size doesn't match. However, AFAICS this isn't an issue anymore, because all finalization methods now only take and return fixed-length-arrays.

Resolves #1094.


newpavlov:
I don't think this wrapper should be in the `core_api` module since it works with the non-core traits.

daxpedda:
Moved to root.

newpavlov (approved):

PR #1827, opened 2025-04-20
signature: replace `signature_derive` with blanket impls

tarcieri:
Adds blanket impls of:
- `Signer` for types which impl `PrehashSignature + DigestSignature`
- `Verifier` for types which impl `PrehashSignature + DigestVerifier`

These blanket impls eliminate the need to have a `signature_derive` crate, which only existed to write the `Signer` and `Verifier` impls for this particular case.

Since this is a breaking change, bumps the version to `3.0.0-pre`


tarcieri:
I think we're at the point where we can do a major version bump, and getting rid of `signature_derive` would be a great reason to do so, IMO.

IIRC this change was originally suggested by @baloo but I can't find where.

We'll need to verify it works against https://github.com/RustCrypto/signatures

baloo:
I think that was for the AsyncSigner? (https://github.com/RustCrypto/traits/pull/1419#discussion_r1430714539)

tarcieri:
Aah, right!

newpavlov (approved):

tarcieri:
@baloo hmm, unfortunately it isn't possible to add that blanket impl because it conflicts with the blanket impl of `AsyncSigner` for `Signer`

baloo:
I think that works (see https://github.com/RustCrypto/signatures/pull/945), there are a couple feature flags to tweak but otherwise that looks good.

tarcieri:
Okay, let’s go with it despite the inconsistency. I guess we could add macro_rules to do the other delegations which is simpler than custom derive

PR #1832, opened 2025-04-24
Implement `Zeroize` for `NonIdentity`

daxpedda:
This PR implements `Zeroize` for `NonIdentity` setting it to G to prevent breaking any invariants.

I just realized that `elliptic-curve` accepts G as a valid `PublicKey`, which seems a bit dangerous to me. I guess 1 is also a valid `SecretKey`. I have no idea if other libraries take these cases into account.


tarcieri:
> I just realized that elliptic-curve accepts G as a valid PublicKey, which seems a bit dangerous to me.

The only disallowed value currently is the additive identity. You could potentially propose restricting some other values which clearly shouldn't be valid public keys.

tarcieri (approved):

PR #1833, opened 2025-04-24
Add `NonIdentity::mul_by_generator()`

daxpedda:
This PR adds `NonIdentity::mul_by_generator()`, which is similar to the `MulByGenerator` trait, but returns a `NonIdentity` instead of a `ProjectivePoint`. This is quite useful for getting the public key from a `NonZeroScalar` without having to go through the whole conversion dance.

### DIFF
--- a/elliptic-curve/src/point/non_identity.rs
+++ b/elliptic-curve/src/point/non_identity.rs
@@ -214,7 +214,7 @@ impl<P: group::Group> Zeroize for NonIdentity<P> {
 #[cfg(all(test, feature = "dev"))]
 mod tests {
     use super::NonIdentity;
-    use crate::dev::{AffinePoint, ProjectivePoint};
+    use crate::dev::{AffinePoint, NonZeroScalar, ProjectivePoint, SecretKey};
     use group::GroupEncoding;
     use hex_literal::hex;
     use zeroize::Zeroize;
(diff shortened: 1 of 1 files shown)


tarcieri:
FYI: `MulByGenerator` is gone as the functionality has been moved to `Group`

elliptic-curve/src/point/non_identity.rs
```diff
@@ -72,6 +72,16 @@ where
...
+
+    /// Multiply by the generator of the prime-order subgroup.
+    pub fn mul_by_generator<C: CurveArithmetic>(scalar: NonZeroScalar<C>) -> Self
+    where
+        P: Copy + Mul<Scalar<C>, Output = P>,
+    {
+        Self {
+            point: P::generator() * *scalar.as_ref(),
+        }
+    }
```
tarcieri, line 84: This should probably delegate to the `Group`-specific implementation (which can use precomputed lookup tables), e.g.

```suggestion
        P: Group,
    {
        Self {
            point: P::mul_by_generator(scalar),
        }
    }
```

daxpedda: Nice! Done!

elliptic-curve/src/point/non_identity.rs
```diff
@@ -72,6 +72,16 @@ where
             point: self.point.to_affine(),
         }
     }
+
+    /// Multiply by the generator of the prime-order subgroup.
+    pub fn mul_by_generator<C: CurveArithmetic>(scalar: NonZeroScalar<C>) -> Self
```
tarcieri, line 77: Perhaps borrow here since the function you're invoking also borrows?

daxpedda: Done!

tarcieri (approved):

PR #1840, opened 2025-05-01
Revert "signature: replace `signature_derive` with blanket impls (#1827)"

### LINKED ISSUE #1831: signature v3: type inference regression
It seems after #1827 type inference is not working as well:

https://github.com/RustCrypto/RSA/pull/505#issuecomment-2825571013

```
error[E0282]: type annotations needed
  --> tests/pkcs1v15.rs:16:9
   |
16 |     let signature = signing_key.sign(bytes);
   |         ^^^^^^^^^
...
22 |     assert_eq!(signature.to_string(), expected);
   |                --------- type must be known at this point
   |
help: consider giving `signature` an explicit type
   |
16 |     let signature: /* Type */ = signing_key.sign(bytes);
   |                  ++++++++++++
```

It can no longer automatically infer the signature type in the case there is only a single non-overlapping impl, which seems pretty bad from an ergonomic perspective.

tarcieri:
This reverts commit bf4774878c7c2e36451119a0ad24f1db6dc93cd8.

Per #1831 this change breaks inference when there is a single explicit impl of the `Signer` trait.

It also wasn't possible to add corresponding blanket impls to the `Async*` traits, e.g. `AsyncSigner` for `AsyncDigestSigner`, because of the existing blanket impl of `AsyncSigner` for `Signer` which we definitely want to preserve.

As a general rule of thumb, blanket impls only make sense if they work 100% of the time, which doesn't seem to be happening here.

Closes #1831


baloo:
Wait, which crates actually consumes the signature_derive again?

tarcieri:
@baloo I think most of the crates avoid it to avoid pulling in the proc macro stack (in which case maybe it's not a great solution)

baloo:
well where I'm trying to go is: maybe we can just pull the `signature_derive` altogether? I can't find a single consumer of it.

tarcieri:
Sure

baloo:
This is actually used in yubihsm.rs!

PR #1844, opened 2025-05-10
digest: impl `BufferKindUser` for newtype

baloo:
Used for supporting hmac (see https://github.com/RustCrypto/MACs/pull/185)


baloo:
cc @newpavlov

newpavlov:
As I wrote [here](https://github.com/RustCrypto/MACs/pull/185#discussion_r2083350524), I don't think we need it. Instead, you should rely on the `CoreProxy` trait.

PR #1850, opened 2025-05-17
Use `Output = True` instead of `NonZero`

daxpedda:
Bounds like `IsLess` aren't actually enforced unless the `IsLess::Output` is checked. This was done by adding `Le<...>: NonZero`. This PR simplifies this check by adding `IsLess<..., Output = True>`.

These checks were entirely missing in the hash2curve implementation, so I added the improved check here as well.


newpavlov (approved):
Interesting. This indeed results in nicer bounds.

But could you remove the `digest` changes? I plan to remove most of the wrappers as part of #1799, so changing them here would result only in unnecessary merging conflicts.

daxpedda:
Done.

PR #1854, opened 2025-05-22
elliptic-curve: scalar `Mul` bounds

tarcieri:
Adds bounds on `CurveArithmetic::Scalar` that it has a `Mul<P>` impl for both `AffinePoint` and `ProjectivePoint`


tarcieri:
Note: will do a corresponding PR to https://github.com/RustCrypto/elliptic-curves before merging this

PR #1857, opened 2025-05-25
Revert "digest: remove `AssociatedOid` blanket impls for wrappers

baloo:
This reverts commit c48ea2a4f76127bd3febce8240095b893e6edb1c (#1810)

This bring back `master` in a state that can be consumed with a `[patch.crates-io]` (see https://github.com/RustCrypto/traits/pull/1799#issuecomment-2885545411)


baloo:
Can’t we revert it back in #1799 ? The downstream changes are already in the downstream PR for #1799

PR #1859, opened 2025-05-27
elliptic-curve: apply digest API breaking changes

baloo:


tarcieri:
Shouldn't this pull in `digest` from the local repo?

tarcieri:
So you're saying you want this merged even though it breaks the build?

baloo:
Or split elliptic-curve integration tests in a crate outside the workspace in an unpublished crate.

tarcieri:
Well we wouldn't bump it until a release, when the `patch.crates-io` directive can be removed

PR #1864, opened 2025-05-28
Improve `BatchInvert`

daxpedda:
This PR does a couple of things:
- Changes the implementation to require two write stores instead of 3 write stores and one read.
- Add a `BatchInvert::batch_invert_mut()` methods that allows users to pass one of the two stores themselves. With this a single store can be allocated instead of 3.
- Reduce the bounds from `Invert + Mul + Copy + Default + ConditionallySelectable` to `Field`.

This was largely inspired by [`curve25519-dalek`s implementation](https://github.com/dalek-cryptography/curve25519-dalek/blob/dd5bd108d6985491acb3de25497fb082a29c9fb7/curve25519-dalek/src/scalar.rs#L788-L837). I adjusted it a bit to save two unnecessary multiplications by one.

Companion PR: https://github.com/RustCrypto/elliptic-curves/pull/1205.

### DIFF
--- a/elliptic-curve/src/ops.rs
+++ b/elliptic-curve/src/ops.rs
@@ -21,6 +21,16 @@ pub trait BatchInvert<FieldElements: ?Sized>: Field + Sized {
     fn batch_invert(
         field_elements: &FieldElements,
     ) -> CtOption<<Self as BatchInvert<FieldElements>>::Output>;
+
+    /// Invert a batch of field elements in-place.
+    ///
+    /// # ⚠️ Warning
+    ///
+    /// Even though `field_elements` is modified regardless of success, on failure it does not
+    /// contain correctly inverted scalars and should be discarded instead.
+    ///
+    /// Consider using [`Self::batch_invert()`] instead.
+    fn batch_invert_mut(field_elements: &mut FieldElements) -> Choice;
 }
 
 impl<const N: usize, T> BatchInvert<[T; N]> for T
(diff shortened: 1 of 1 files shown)


elliptic-curve/src/ops.rs
```diff
@@ -1,78 +1,77 @@
...
 /// at an amortized cost that should be practically as efficient as a single inversion.
-pub trait BatchInvert<FieldElements: ?Sized>: Invert + Sized {
+pub trait BatchInvert<FieldElements: ?Sized>: Field + Sized {
     /// The output of batch inversion. A container of field elements.
     type Output: AsRef<[Self]>;
 
     /// Invert a batch of field elements.
     fn batch_invert(
         field_elements: &FieldElements,
     ) -> CtOption<<Self as BatchInvert<FieldElements>>::Output>;
```
tarcieri, line 23: It seems like you could just have a default impl here:

```suggestion
    ) -> CtOption<<Self as BatchInvert<FieldElements>>::Output> {
        let mut field_elements_inverses = *field_elements;
        let inversion_succeeded = Self::batch_invert_mut(&mut field_elements_inverses);
        CtOption::new(field_elements_inverses, inversion_succeeded)
    }
```

Six of one, half dozen of another though

tarcieri:
Funny, I was just looking at dalek's `batch_invert` after implementing https://github.com/dalek-cryptography/curve25519-dalek/pull/759 and thinking something similar to this was possible, thanks

elliptic-curve/src/ops.rs
daxpedda: Hm, but it would require `ToOwned`, which is behind `alloc` unfortunately. Not sure how else to pull this of.

We could change `batch_invert` to take ownership instead of a reference and remove `batch_invert_mut()` entirely?

tarcieri (approved):

PR #1865, opened 2025-05-28
Remove insecure `BatchInvert` API

daxpedda:
Implementing the idea I had in https://github.com/RustCrypto/traits/pull/1864#discussion_r2112346173, which in hindsight I actually think is better.

The new implementation actually *should* take ownership instead of just cloning *for* the user. See [C-CALLER-CONTROL API guidelines](https://rust-lang.github.io/api-guidelines/flexibility.html?highlight=clone#caller-decides-where-to-copy-and-place-data-c-caller-control).

This has the advantage of getting read of the insecure `batch_invert_mut` API where users have to make sure to actually check the returned `Choice` and make sure they don't use the input slice.

I could also quickly add the following for even more control if desirable:
```rust
#[cfg(feature = "alloc")]
impl<'this, T> BatchInvert<&'this mut [Self]> for T
where
    T: Field,
{
    type Output = &'this mut [Self];

    fn batch_invert(field_elements: &'this mut [Self]) -> CtOption<&'this mut [Self]> {
        let mut field_elements_pad: Vec<Self> = vec![Self::default(); field_elements.len()];
        let inversion_succeeded = invert_batch_internal(field_elements, &mut field_elements_pad);

        CtOption::new(field_elements, inversion_succeeded)
    }
}
```

### DIFF
--- a/elliptic-curve/src/ops.rs
+++ b/elliptic-curve/src/ops.rs
@@ -9,7 +9,7 @@ use ff::Field;
 use subtle::{Choice, CtOption};
 
 #[cfg(feature = "alloc")]
-use alloc::{borrow::ToOwned, vec::Vec};
+use alloc::vec::Vec;
 
 /// Perform a batched inversion on a sequence of field elements (i.e. base field elements or scalars)
 /// at an amortized cost that should be practically as efficient as a single inversion.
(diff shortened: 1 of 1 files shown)


tarcieri:
I like the blanket impl idea, and it seems like you could have a similar one for immutable slices which returns a `Vec`

tarcieri (approved):

PR #1868, opened 2025-05-28
digest v0.11.0-rc.0

tarcieri:
NOTE: this temporarily disables the `ecdh`, `hash2curve`, and `oprf` features of `elliptic-curve` to make the release possible.

Going forward we can remove the `=` from the `digest` version requirement now that we're done making major breaking changes, which should allow for more flexible upgrades.

They can be re-enabled when the `hmac`, `hkdf`, `sha2`, and `sha3` crates have been updated.

### DIFF
--- a/signature/Cargo.toml
+++ b/signature/Cargo.toml
@@ -13,12 +13,8 @@ categories = ["cryptography", "no-std"]
 description = "Traits for cryptographic signature algorithms (e.g. ECDSA, Ed25519)"
 
 [dependencies]
-digest = { version = "=0.11.0-pre.10", optional = true, default-features = false }
+digest = { version = "0.11.0-rc.0", optional = true, default-features = false }
 rand_core = { version = "0.9", optional = true, default-features = false }
-
-[dev-dependencies]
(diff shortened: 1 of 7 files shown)


signature/Cargo.toml
```diff
@@ -13,13 +13,9 @@ categories = ["cryptography", "no-std"]
 description = "Traits for cryptographic signature algorithms (e.g. ECDSA, Ed25519)"
 
 [dependencies]
-digest = { version = "=0.11.0-pre.10", optional = true, default-features = false }
+digest = { version = "0.11.0-rc.0", optional = true, default-features = false }
 rand_core = { version = "0.9", optional = true, default-features = false }
 
-[dev-dependencies]
-hex-literal = "1"
-sha2 = { version = "=0.11.0-pre.5", default-features = false }
```
tarcieri, line 21: Turns out these were no-longer used

newpavlov (approved):

newpavlov:
Note that I have one additional breaking change in mind: I want to make serialization of eager hashes slightly more efficient.

Issue #1808, opened 2025-04-02
`chacha20`: Seeking near the end of the keystream causes `try_current_pos` to return an error


jpdoyle:
The following test fails on the last line:
```
    #[test]
    fn test_current_pos() {
        let offset: u64 = 256u64 << 30;

        let mut cipher = chacha20::ChaCha20::new(&Default::default(), &Default::default());
        cipher
            .try_seek(offset - 64)
            .expect("Couldn't seek to nearly 256GB");
        assert_eq!(cipher.try_current_pos::<u64>().unwrap(),
            offset - ((1<<6))
        );
        cipher
            .try_seek(offset - 63)
            .expect("Couldn't seek to nearly 256GB");
        cipher.try_current_pos::<u64>().unwrap();
    }
```
The root cause appears to be a combination of the incorrect looping counter behavior highlighted in https://github.com/RustCrypto/stream-ciphers/issues/391 and from the behavior of `try_seek` in `StreamCipherCoreWrapper`. `try_seek` sets the block position to `(((256u64 << 30)-63) / 64) as u32`, which is `u32::MAX`, and then calls `write_keystream_block`, which increments the counter, causing it to loop back to 0. Then the `try_current_pos` calls `get_block_pos()`, which returns 0 -- see https://github.com/RustCrypto/traits/blob/67131afeab8f235bb1698f94585bb84c8afd9716/cipher/src/stream/wrapper.rs#L176-L202

`SeekNum::get_block_byte` is then called with a value of 0 for the `block` parameter, and the checked subtraction on [this line](https://github.com/RustCrypto/traits/blob/67131afeab8f235bb1698f94585bb84c8afd9716/cipher/src/stream.rs#L225) returns `None`.

newpavlov:
The reason behind this issue is the following.

For efficiency sake we want to be lazy with keystream generation. It means that during stream cipher initialization we do not generate any keystream blocks and the buffer gets initialized into the "exhausted" state (we do it by setting cursor position to block size, which results in number of unread bytes equal to zero). This also allows us to store cursor position in the first byte of the buffer instead of introducing a separate field, since it's guaranteed that after block generation we will consume at least one byte from it (otherwise we would leave the buffer in the "exhausted" state).

[...]

PR #1874, opened 2025-05-29
cipher v0.5.0-rc.0

tarcieri:


newpavlov (approved):
I am fine with cutting a pre-release, but before the full release we should fix #1808 which may involve some breaking changes.

PR #1882, opened 2025-06-02
aead: deprecated `AeadInPlace` trait

tarcieri:
`AeadInPlace` was removed in #1798, however in migrating some older code I thought it would be really helpful to have backwards compatibility and deprecations in place to help people migrate to the new names, especially as the "in place" -> "inout" migrations aren't entirely intuitive.

### DIFF
--- a/aead/src/lib.rs
+++ b/aead/src/lib.rs
@@ -329,6 +329,98 @@ pub trait AeadInOut: AeadCore {
     }
 }
 
+/// Legacy in-place stateless AEAD trait.
+///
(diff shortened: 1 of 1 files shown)


newpavlov (approved):
Personally, I am not a fan of having deprecated items in a new breaking release, but I guess it could help users with gradual migration.

tarcieri:
These migrations are just so much easier when your code still compiles but you get helpful warnings about what needs to be changed and where, instead of compile errors and having to look through crate-by-crate changelogs

PR #1890, opened 2025-06-03
elliptic-curve: impl `BatchInvert` for `NonZeroScalar`

daxpedda:
This implements `BatchInvert` for `NonZeroScalar`.

To accomplish this, I did the following notable things:
- Remove all trait bounds on `BatchInvert` itself.
- Remove `CtOption` from `BatchInvert::batch_invert`s return type.
- Expose `invert_batch_internal()` internally, which now takes an `invert` function instead of requiring `trait Invert`.
- Implement `MulAssign` for `NonZeroScalar`.

Things that could still be added:
- Mirror all `BatchInvert` implementations on `Scalar`, I left the ones taking a reference to a slice out.
- Implement `MulAssign` for `NonZeroScalar` on more combinations.

It might be a tiny bit simpler if I implemented `Default` on `NonZeroScalar`, but that seemed like a footgun to me.

### DIFF
--- a/elliptic-curve/src/ops.rs
+++ b/elliptic-curve/src/ops.rs
@@ -1,7 +1,7 @@
 //! Traits for arithmetic operations on elliptic curve field elements.
 
 use core::iter;
-pub use core::ops::{Add, AddAssign, Mul, Neg, Shr, ShrAssign, Sub, SubAssign};
+pub use core::ops::{Add, AddAssign, Mul, MulAssign, Neg, Shr, ShrAssign, Sub, SubAssign};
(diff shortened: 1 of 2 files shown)


tarcieri (approved):

tarcieri:
@daxpedda hmm, probably should've asked for a companion PR for this to https://github.com/RustCrypto/elliptic-curves for this, having a little trouble updating `primeorder`

tarcieri:
Never mind, got it: https://github.com/RustCrypto/elliptic-curves/pull/1234

PR #1891, opened 2025-06-05
elliptic-curve: expose `AffineCoordinates::y`

### LINKED ISSUE #1019: Add AffineYCoordinate trait?
Understanding the arguments pro/con coordinates from https://github.com/RustCrypto/traits/pull/817...

For a generic-ish Frost implementation, I have need to determine if the output of distributed keygen has even y-coordinate, to perform the Taproot adjustment. I can get away with a local `AffineYCoordinateParity` trait (`fn is_y_even(&self) -> bool`) that I implement on `k256` and bound on, but maybe "for symmetry" it wouldn't be that horrible to add either `AffineYCoordinate` or at least `AffineYCoordinateParity` to the `AffineArithmetic::AffinePoint` bounds?

tarcieri:
In the past we've deliberately avoided exposing the y-coordinate to prevent the possibility of things like invalid curve attacks, although with time we have exposed more and more to support things like alternative point compression formats. See #1237 for some history.

We're now trying to use these traits with Edwards curves like Curve25519 (in `curve25519-dalek`) and Ed448-Goldilocks, which use compressed Edwards y-coordinates as their compressed point format. That requires y-coordinate access.

As such, this changes the previous `y_is_odd` method, which was used to implement SEC1-like compressed points, to a full `fn y` which returns a serialized field element for the y-coordinate.

Closes #1019

### DIFF
--- a/elliptic-curve/src/dev.rs
+++ b/elliptic-curve/src/dev.rs
@@ -476,6 +476,10 @@ impl AffineCoordinates for AffinePoint {
     fn y(&self) -> FieldBytes {
         unimplemented!();
     }
+
+    fn y_is_odd(&self) -> Choice {
+        unimplemented!();
+    }
 }
 
 impl ConstantTimeEq for AffinePoint {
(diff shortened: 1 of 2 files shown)


tarcieri:
With one minor hiccup, there's now a companion PR for this here: https://github.com/RustCrypto/elliptic-curves/pull/1230

PR #1896, opened 2025-06-08
Implement `BatchNormalize` for `NonIdentity`

### LINKED ISSUE #1889: Implement `BatchNormalize` for `NonIdentity`
Implementing `BatchNormalize` for `NonIdentity` is currently not possible because the implementation needs access to various internals that are currently not abstracted over in `elliptic-curve`. So the implementation currently has to live where `ProjectivePoint` is defined, e.g. `primeorder`.

Because `NonIdentity` is a non-local type for e.g. `primeorder`, it can't implement `BatchNormalize` for it. However, we can always implement it in `elliptic-curve` by allocating more memory to basically make a copy of the given list converted to `ProjectivePoint`s and call its `BatchNormalize` implementation.

One more thing that comes to mind: make `NonIdentity` `#[transparent]` and transmute the given slice to `ProjectivePoint`s without having to allocate additional memory. Which is `unsafe` ofc.

daxpedda:
As discussed in #1889. I will add some tests in `elliptic-curves` as well.

Resolves #1889.
Companion PR: https://github.com/RustCrypto/elliptic-curves/pull/1248.

### DIFF
--- a/elliptic-curve/src/point/non_identity.rs
+++ b/elliptic-curve/src/point/non_identity.rs
@@ -308,6 +308,7 @@ impl<P: Group> Zeroize for NonIdentity<P> {
 #[cfg(all(test, feature = "dev"))]
 mod tests {
     use super::NonIdentity;
+    use crate::BatchNormalize;
     use crate::dev::{AffinePoint, NonZeroScalar, ProjectivePoint, SecretKey};
     use group::GroupEncoding;
     use hex_literal::hex;
@@ -372,5 +373,39 @@ mod tests {
         let pk = sk.public_key();
 
         assert_eq!(point.to_point(), pk.to_projective());
+    }
+
+    #[test]
+    fn batch_normalize() {
+        let point = ProjectivePoint::from_bytes(
+            &hex!("02c9afa9d845ba75166b5c215767b1d6934e50c3db36e89b127b8a622b120f6721").into(),
+        )
+        .unwrap();
+        let point = NonIdentity::new(point).unwrap();
+        let points = [point, point];
+
+        for (point, affine_point) in points
+            .into_iter()
+            .zip(NonIdentity::batch_normalize(&points))
+        {
+            assert_eq!(point.to_affine(), affine_point);
+        }
+    }
+
+    #[test]
+    #[cfg(feature = "alloc")]
+    fn batch_normalize_alloc() {
+        let point = ProjectivePoint::from_bytes(
+            &hex!("02c9afa9d845ba75166b5c215767b1d6934e50c3db36e89b127b8a622b120f6721").into(),
+        )
+        .unwrap();
+        let point = NonIdentity::new(point).unwrap();
+        let points = vec![point, point];
+
+        let affine_points = NonIdentity::batch_normalize(points.as_slice());
+
+        for (point, affine_point) in points.into_iter().zip(affine_points) {
+            assert_eq!(point.to_affine(), affine_point);
+        }
     }
 }
--- a/.github/workflows/elliptic-curve.yml
+++ b/.github/workflows/elliptic-curve.yml
@@ -88,3 +88,21 @@ jobs:
       - run: cargo test --no-default-features
       - run: cargo test
       - run: cargo test --all-features
+
+  test-careful:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@nightly
+      - run: cargo install cargo-careful
+      - run: cargo careful test --all-features
+
+  test-miri:
+    runs-on: ubuntu-latest
+    env:
+      MIRIFLAGS: "-Zmiri-symbolic-alignment-check -Zmiri-strict-provenance"
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@nightly
+      - run: rustup component add miri && cargo miri setup
+      - run: cargo miri test --all-features
--- a/elliptic-curve/src/dev.rs
+++ b/elliptic-curve/src/dev.rs
@@ -4,7 +4,7 @@
 //! the traits in this crate.
 
 use crate::{
-    Curve, CurveArithmetic, FieldBytesEncoding, PrimeCurve,
+    BatchNormalize, Curve, CurveArithmetic, FieldBytesEncoding, PrimeCurve,
     array::typenum::U32,
     bigint::{Limb, U256},
     error::{Error, Result},
@@ -17,12 +17,16 @@ use crate::{
     zeroize::DefaultIsZeroes,
 };
 use core::{
+    array,
     iter::{Product, Sum},
     ops::{Add, AddAssign, Mul, MulAssign, Neg, Sub, SubAssign},
 };
 use ff::{Field, PrimeField};
 use hex_literal::hex;
 use pkcs8::AssociatedOid;
+
+#[cfg(feature = "alloc")]
+use alloc::vec::Vec;
 
 #[cfg(feature = "bits")]
 use ff::PrimeFieldBits;
@@ -582,6 +586,23 @@ pub enum ProjectivePoint {
 
     /// Is this point a different point corresponding to a given [`AffinePoint`]
     Other(AffinePoint),
+}
+
+impl<const N: usize> BatchNormalize<[ProjectivePoint; N]> for ProjectivePoint {
+    type Output = [AffinePoint; N];
+
+    fn batch_normalize(points: &[ProjectivePoint; N]) -> [AffinePoint; N] {
+        array::from_fn(|index| points[index].into())
+    }
+}
+
+#[cfg(feature = "alloc")]
+impl BatchNormalize<[ProjectivePoint]> for ProjectivePoint {
+    type Output = Vec<AffinePoint>;
+
+    fn batch_normalize(points: &[ProjectivePoint]) -> Vec<AffinePoint> {
+        points.iter().copied().map(AffinePoint::from).collect()
+    }
 }
 
 impl ConstantTimeEq for ProjectivePoint {


elliptic-curve/src/point/non_identity.rs
```diff
@@ -103,6 +108,66 @@ impl<P> AsRef<P> for NonIdentity<P> {
...
+            size_of::<P::AffineRepr>(),
+            size_of::<NonIdentity<P::AffineRepr>>()
+        );
+        debug_assert_eq!(
+            align_of::<P::AffineRepr>(),
+            align_of::<NonIdentity<P::AffineRepr>>()
+        );
+
+        #[expect(unsafe_code, reason = "`NonIdentity` is `repr(transparent)`")]
+        let points: &[P; N] = unsafe { mem::transmute(points) };
```
tarcieri: I think a pointer cast provides more compile-time checks, possibly including some of your `debug_assert`s above:

```suggestion
        let points = unsafe { &*(points as *const [Self; N] as *const [P; N]) };
```

Unfortunately I don't think you can use the [`cast`](https://doc.rust-lang.org/std/primitive.pointer.html#method.cast) function here.

elliptic-curve/src/point/non_identity.rs
```diff
@@ -103,6 +108,66 @@ impl<P> AsRef<P> for NonIdentity<P> {
...
+        debug_assert_eq!(
+            size_of::<P::AffineRepr>(),
+            size_of::<NonIdentity<P::AffineRepr>>()
+        );
+        debug_assert_eq!(
+            align_of::<P::AffineRepr>(),
+            align_of::<NonIdentity<P::AffineRepr>>()
+        );
+
+        #[expect(unsafe_code, reason = "`NonIdentity` is `repr(transparent)`")]
```
tarcieri: I guess this takes the place of a `SAFETY` comment? I haven't seen this particular style before

daxpedda: This seems to be a bug in [`clippy::missing_safety_doc`](https://rust-lang.github.io/rust-clippy/master/index.html#missing_safety_doc): even without `reason` it doesn't trigger.

I replaced this with `// SAFETY` instead.

elliptic-curve/src/point/non_identity.rs
daxpedda: I completely removed any usage of `transmute()`.

elliptic-curve/src/point/non_identity.rs
```diff
@@ -103,6 +108,81 @@ impl<P> AsRef<P> for NonIdentity<P> {
...
+    fn batch_normalize(points: &[Self; N]) -> [NonIdentity<P::AffineRepr>; N] {
+        // Ensure casting is safe.
+        // This always succeeds because `NonIdentity` is `repr(transparent)`.
+        debug_assert_eq!(size_of::<P>(), size_of::<NonIdentity<P>>());
+        debug_assert_eq!(align_of::<P>(), align_of::<NonIdentity<P>>());
+
+        #[allow(unsafe_code)]
+        // SAFETY: `NonIdentity` is `repr(transparent)`.
+        let points: &[P] = unsafe { slice::from_raw_parts(points.as_ptr().cast(), N) };
+        let points = points.try_into().expect("slice should be size `N`");
```
daxpedda: I requested `array::from_raw_parts()` here: https://github.com/rust-lang/libs-team/issues/602.

daxpedda:
I completely removed any usage of `mem::transmute()` and `mem::transmute_copy()` and opted for safer alternatives.

One notably change: instead of `mem::transmute_copy()` for fixed arrays, I used `array::map()`. This seems to be optimized to a `memcpy` by compilers ([Godbolt](https://rust.godbolt.org/z/1Pdn5KbTr)). Additionally this gets rid of one `unsafe` call.

elliptic-curve/src/point/non_identity.rs
daxpedda: The best alternative is probably `ptr::read()`.

elliptic-curve/src/lib.rs
```diff
@@ -5,7 +5,7 @@
     html_logo_url = "https://raw.githubusercontent.com/RustCrypto/media/8f1a9894/logo.svg",
     html_favicon_url = "https://raw.githubusercontent.com/RustCrypto/media/8f1a9894/logo.svg"
 )]
-#![forbid(unsafe_code)]
+#![deny(unsafe_code)]
```
tarcieri, line 9: It'd be good to leave a comment here about what unsafety is allowed

```suggestion
#![deny(unsafe_code)] // only for newtype casting
```

...or something to that effect

elliptic-curve/src/point/non_identity.rs
tarcieri: A pointer cast should work and avoid the intermediate slice?

elliptic-curve/src/point/non_identity.rs
```diff
@@ -103,6 +108,81 @@ impl<P> AsRef<P> for NonIdentity<P> {
...
+        debug_assert_eq!(
+            size_of::<P::AffineRepr>(),
+            size_of::<NonIdentity<P::AffineRepr>>()
+        );
+        debug_assert_eq!(
+            align_of::<P::AffineRepr>(),
+            align_of::<NonIdentity<P::AffineRepr>>()
+        );
+
+        affine_points.map(|point| NonIdentity { point })
```
daxpedda, line 137: My comment https://github.com/RustCrypto/traits/pull/1896/files#r2142491767 was supposed to be about this line, apologies.

Here, the best alternative to `mem::transmute_copy()`, is `ptr::read()`.

However, as noted before, `array::map()` seems to optimize just fine.

elliptic-curve/src/lib.rs
daxpedda: Done.

elliptic-curve/src/point/non_identity.rs
daxpedda: I have to say I prefer the `slice::from_raw_parts()` + `try_into()` approach to pointer casting.
```suggestion
        let points: &[P; N] = unsafe { &*(points as *const [Self; N]).cast() };
```

The fallible conversion can also be optimized away by the compiler: https://rust.godbolt.org/z/eoaWG4sYd.

I pushed the pointer cast into a separate commit, let me know if you have a change of heart and I can revert it.

tarcieri: It seems a lot more straightforward to me, rather than going from a `Sized` type to a DST and then back, even if the compiler can optimize the bounds check away. I believe the cast even ensures the compiler will statically assert a size check?

That said, I don't suppose `points.as_ptr().cast()` works there?

daxpedda: > I believe the cast even ensures the compiler will statically assert a size check?

What kind of size check? I [checked a couple of things in the playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=48d7b87c0877ab4dcfb93cbce29188ba) and they all seem to pass.

AFAIK casts don't check a whole lot, which is why I prefer going with anything else really.

> That said, I don't suppose `points.as_ptr().cast()` works there?

Done.

tarcieri: I think a cast still checks more than transmute, although I’d have to ask the dark arts channel what. Transmute checks nothing AFAIK.

daxpedda: > I think a cast still checks more than transmute, although I’d have to ask the dark arts channel what. Transmute checks nothing AFAIK.

Indeed, this is my understanding as well. I was comparing it to `slice::from_raw_parts()` + `try_into()`.

elliptic-curve/src/point.rs
```diff
@@ -40,9 +40,9 @@ pub trait AffineCoordinates {
 
 /// Normalize point(s) in projective representation by converting them to their affine ones.
 #[cfg(feature = "arithmetic")]
-pub trait BatchNormalize<Points: ?Sized>: group::Curve {
+pub trait BatchNormalize<Points: ?Sized> {
     /// The output of the batch normalization; a container of affine points.
-    type Output: AsRef<[Self::AffineRepr]>;
+    type Output;
```
tarcieri, line 45: Seems like these are breaking changes that might need a corresponding PR to https://github.com/RustCrypto/elliptic-curves

daxpedda: Done in https://github.com/RustCrypto/elliptic-curves/pull/1248 already.

elliptic-curve/src/point/non_identity.rs
```diff
@@ -103,6 +107,81 @@ impl<P> AsRef<P> for NonIdentity<P> {
...
+        let ptr = affine_points.as_mut_ptr();
+        let length = affine_points.len();
+        let capacity = affine_points.capacity();
+        core::mem::forget(affine_points);
+
+        #[allow(unsafe_code)]
+        // SAFETY: `NonIdentity` is `repr(transparent)`.
+        unsafe {
+            Vec::from_raw_parts(ptr.cast(), length, capacity)
+        }
```
tarcieri: Similarly here it seems like all of this could be replaced by casting `Vec<P::AffineRepr>` to `Vec<NonIdentity<P::AffineRepr>>`

elliptic-curve/src/point.rs
tarcieri: Oh sorry, missed that

elliptic-curve/src/point/non_identity.rs
daxpedda: I have to say that I find pointer casting much more fickle and harder to verify then using `*::from_raw_parts()`.

In any case, it is done.

daxpedda: Case in point: I just quickly tried `Vec::as_ptr()`, which fails, because its not a pointer to the `Vec`, but to the inner slice.

On a side note: I can't use `slice::as_ptr()` because it returns a thin pointer. But slices require a fat pointer for the length.

tarcieri: Aah, yeah it’d need a transmute.

I find one relatively declarative line of unsafe code preferable to several which tear the Vec apart and put it back together again

daxpedda: > Aah, yeah it’d need a transmute.

I already did it with `let points: &[P] = unsafe { &*(points as *const [NonIdentity<P>] as *const [P]) };`, which preserves the fat pointer. Or do you want me to change that to a `mem::transmute()`?

elliptic-curve/src/point/non_identity.rs
```diff
@@ -103,6 +107,79 @@ impl<P> AsRef<P> for NonIdentity<P> {
...
+        #[allow(unsafe_code)]
+        // SAFETY: `NonIdentity` is `repr(transparent)`.
+        let result: Vec<NonIdentity<P::AffineRepr>> = unsafe {
+            (&affine_points as *const Vec<P::AffineRepr>)
+                .cast::<Vec<NonIdentity<P::AffineRepr>>>()
+                .read()
+        };
+        core::mem::forget(affine_points);
+
+        result
```
tarcieri: We could just go the safe route here:

```suggestion
    affine_points.into_iter().map { |point| NonIdentity { point } ).collect() 
```

The `unsafe` usage here seems like a big dance to ensure there isn't a second heap allocation, but is there actually one or is the compiler able to reuse the existing one with safe code?

elliptic-curve/src/point/non_identity.rs
tarcieri: Alternative suggestion: https://github.com/RustCrypto/traits/pull/1896#discussion_r2144993168

tarcieri:
@daxpedda can you also add `cargo careful` and `cargo miri` to CI, similar to these?

https://github.com/RustCrypto/crypto-bigint/blob/dff9eb8/.github/workflows/crypto-bigint.yml#L98-L105

https://github.com/RustCrypto/crypto-bigint/blob/dff9eb8/.github/workflows/crypto-bigint.yml#L133-L147

It would also be good to find a way to exercise this code with tests somewhere in this repo, possibly leveraging `MockCurve`.

elliptic-curve/src/point/non_identity.rs
daxpedda: Nope, the compiler seems to be able to figure it out: https://rust.godbolt.org/z/z6M75dnsE.

daxpedda: Done, let me know if you want me remove the `debug_assert!`s.

tarcieri (approved):
Looks good now. I can handle the CI changes as a followup.

daxpedda:
> Looks good now. I can handle the CI changes as a followup.

Already on it. Trying to add a test right now.

elliptic-curve/src/point/non_identity.rs
```diff
@@ -103,6 +107,72 @@ impl<P> AsRef<P> for NonIdentity<P> {
...
+
+    fn batch_normalize(points: &[Self; N]) -> [NonIdentity<P::AffineRepr>; N] {
+        // Ensure casting is safe.
+        // This always succeeds because `NonIdentity` is `repr(transparent)`.
+        debug_assert_eq!(size_of::<P>(), size_of::<NonIdentity<P>>());
+        debug_assert_eq!(align_of::<P>(), align_of::<NonIdentity<P>>());
+
+        #[allow(unsafe_code)]
+        // SAFETY: `NonIdentity` is `repr(transparent)`.
+        let points: &[P; N] = unsafe { &*points.as_ptr().cast() };
```
tarcieri, line 124: Curious if you could extract an `AsRef` impl here, but perhaps I can experiment with that myself

daxpedda: Apparently not, because `array` is a foreign type. Same applies to slices.

I guess we could add it as a method?

PR #1898, opened 2025-06-09
digest: add `CoreProxy::compose/decompose` methods

newpavlov:


digest/src/buffer_macros/xof.rs
```diff
@@ -87,7 +87,7 @@ macro_rules! buffer_xof {
     ) => {
         $crate::buffer_xof!(
             impl_inner: $name($core_ty);
-            Debug Clone BlockSizeUser CoreProxy
+            Debug Clone BlockSizeUser
```
newpavlov, line 90: Note that XOF readers no longer implement `CoreProxy` since `CoreProxy::Core` is bounded by `BufferKindUser`. We could make the buffer another associated type, but I don't think it's worth the trouble.

If necessary, in future we may add inherent methods to the XOF readers.

PR #1903, opened 2025-06-14
elliptic-curve: extract `NonIdentity`/`NonZeroScalar` casting methods

tarcieri:
These methods encapsulate safely casting references for arrays and slices of `NonIdentity<P>` to array/slice references to the inner `P` type, and the same for `NonZeroScalar` and its inner `Scalar<C>` type.

Note the choice of method names follows from similar ones in `hybrid-array`: https://docs.rs/hybrid-array/latest/hybrid_array/struct.Array.html#method.cast_slice_to_core

cc @daxpedda

### DIFF
--- a/elliptic-curve/src/point/non_identity.rs
+++ b/elliptic-curve/src/point/non_identity.rs
@@ -22,7 +22,7 @@ use crate::{BatchNormalize, CurveArithmetic, NonZeroScalar, Scalar};
 /// In the context of ECC, it's useful for ensuring that certain arithmetic
 /// cannot result in the identity point.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
-#[repr(transparent)]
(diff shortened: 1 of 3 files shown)


elliptic-curve/src/point/non_identity.rs
```diff
@@ -114,26 +147,8 @@ where
...
-
-        // Ensure `array::map()` can be optimized to a `memcpy`.
-        debug_assert_eq!(
-            size_of::<P::AffineRepr>(),
-            size_of::<NonIdentity<P::AffineRepr>>()
-        );
-        debug_assert_eq!(
-            align_of::<P::AffineRepr>(),
-            align_of::<NonIdentity<P::AffineRepr>>()
-        );
```
tarcieri, line 135: I removed these because these properties should always hold true for a `repr(transparent)` newtype

PR #1907, opened 2025-06-16
cipher: add methods for writing keystream

newpavlov:
Closes #424

### DIFF
--- a/cipher/src/stream.rs
+++ b/cipher/src/stream.rs
@@ -104,6 +104,16 @@ pub trait StreamCipher {
         self.try_apply_keystream_inout(buf.into())
     }
 
+    /// Write keystream to `buf`.
+    ///
+    /// If end of the keystream will be achieved with the given data length,
+    /// method will return [`StreamCipherError`] without modifying provided `data`.
+    #[inline]
+    fn try_write_keystream(&mut self, buf: &mut [u8]) -> Result<(), StreamCipherError> {
+        buf.fill(0);
+        self.try_apply_keystream(buf)
+    }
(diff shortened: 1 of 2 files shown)


cipher/src/stream.rs
```diff
@@ -130,6 +140,16 @@ pub trait StreamCipher {
         self.try_apply_keystream(buf).unwrap();
     }
 
+    /// Write keystream to `buf`.
+    ///
+    /// # Panics
+    /// If end of the keystream will be reached with the given data length,
+    /// method will panic without modifying the provided `data`.
+    #[inline]
+    fn write_keystream(&mut self, buf: &mut [u8]) {
```
nazar-pc, line 149: Any strong reason to prefer `&mut [u8]` over `&mut [MaybeUninit<u8>]`? That is exactly the type that `Vec::spare_capacity_mut()` returns and I'd rather not rely on compiler guessing and removing extra zeroing.

In fact the method could me something like this:
```rust
fn write_keystream(&mut self, buf: &mut [MaybeUninit<u8>]) -> &mut [u8] {
```

Where return type is `buf` with bytes filled in.

nazar-pc (commented):
Thanks for this!

cipher/src/stream.rs
newpavlov: We try to minimize amount of unsafe code in our crates and usually in practice bothering with `MaybeUninit<u8>` is not worth the trouble, especially with code like this which is very easy for the compiler to remove unnecessary buffer zeroization.

Issue #1911, opened 2025-06-21
elliptic-curve: Move away from the `Curve` and `CurveArithmetic` traits


carloskiki:
I have been thinking of ways to improve the `elliptic-curve` crate so that it can be properly implemented for all types of curves, and I think the best way we could do this is by deprecating the family of `Curve` traits. I know this would be a big change, and I am not sure I am right, but here's how I see it:

I like to use zero sized types as proofs. _Proof_ that you have acquired lock, _proof_ that a method was called, _proof_ that something was instantiated, etc. (See this [amazing blog](https://www.hardmo.de/article/2021-03-14-zst-proof-types.md), that's where I get my opinion from.) Zero sized types that can simply be constructed do not do much except serve as labels (which is what `Curve` instances do). I claim that we don't need these "labels", because we already have crates. For example, instead of `elliptic_curve::AffinePoint<P256>`, I would rather see `p256::AffinePoint`.

I went through `elliptic-curve` to see how this would transform the crate. Here is an example with the `ecdh` module. Currently the `diffie_hellman` function is defined as such:

```rust
pub fn diffie_hellman<C>(
    secret_key: impl Borrow<NonZeroScalar<C>>,
    public_key: impl Borrow<AffinePoint<C>>,
) -> SharedSecret<C>
where
    C: CurveArithmetic,
{
    let public_point = ProjectivePoint::<C>::from(*public_key.borrow());
    let secret_point = (public_point * secret_key.borrow().as_ref()).to_affine();
    SharedSecret::new(secret_point)
}
```

If we remove the `CurveArithmetic` trait, and instead consider what we want `secret_key` and `public_key` to implement, it could look something like this:

```rust
pub fn diffie_hellman<P>(
    secret_scalar: impl Borrow<NonZeroScalar<P::Scalar>>,
    public_point: impl Borrow<P>,
) -> SharedSecret<P>
where
    P: group::Group + AffineCoordinates,
{
    let secret_point = *public_point.borrow() * secret_scalar.borrow().as_ref();
    SharedSecret::new(secret_point)
}
```

Instead of requiring that some `Curve` trait is implemented, we require that a point implements `group::Group` and `AffineCoordinates`. We are removing a layer of indirection by directly specifying our requirements on the data we are using.

Removing this indirection would work for most other types in the library, for example it would be `NonZeroScalar<S>` instead of `NonZeroScalar<C: CurveArithmetic>` which would match `NonIdentity<P>`.

[...]

Issue #1177, opened 2022-12-24
Use `Group` instead of `Curve` when appropriate


daxpedda:
With https://github.com/dalek-cryptography/curve25519-dalek/pull/473 hopefully getting done, it would be nice to use more types with Ristretto255.

Unfortunately types like `PublicKey`, `SecretKey`, `NonZeroScalar`, `hash2curve::GroupDigest`, and the `ecdh` module all work over `Curve`, which can not be implemented over Ristretto255 (as far as I understand).

So transitioning these types to use `Group` instead could allow us to use these types with Ristretto255 too.

tarcieri:
I think you're confusing [`elliptic_curve::Curve`](https://docs.rs/elliptic-curve/0.12.3/elliptic_curve/trait.Curve.html) with [`group::Curve`](https://docs.rs/group/0.13.0/group/trait.Curve.html). The names are indeed confusing, however they serve two different functions.

`elliptic_curve::Curve` is pretty much the core trait of that respective crate, and provides a baseline level of functionality even for curves which may lack a full arithmetic implementation, such as ensuring scalars are in range of the curve's order. Notably the existence of that trait predates any integration with `ff`/`group` or perhaps less ambiguous names could have be selected.

As for getting everything to work with Ristretto255, that will probably take some refactoring and maybe some combinations of features just don't make sense. But it's definitely not just as simple as `s/Curve/Group/`.

daxpedda:
> I think you're confusing [`elliptic_curve::Curve`](https://docs.rs/elliptic-curve/0.12.3/elliptic_curve/trait.Curve.html) with [`group::Curve`](https://docs.rs/group/0.13.0/group/trait.Curve.html).

How so? Do you mean because I was proposing to use `group::Group`, which isn't a good replacement for `elliptic_curve::Curve` (now that I look at it)?

> As for getting everything to work with Ristretto255, that will probably take some refactoring and maybe some combinations of features just don't make sense. But it's definitely not just as simple as `s/Curve/Group/`.

As far as I could tell, it would take some major restructuring and I'm clueless enough not to really be able to say what direction this should go.

[...]

PR #1925, opened 2025-07-06
elliptic-curve: make hash2curve traits depend on `Group` rather than `CurveArithmetic`

carloskiki:
This was previously requested as part of #1177, and some related discussion took place in #1911.
 
This seems relevant as the main trait name is `GroupDigest`.

I will also open a PR to all implementing crates.

### DIFF
--- a/elliptic-curve/src/hash2curve/group_digest.rs
+++ b/elliptic-curve/src/hash2curve/group_digest.rs
@@ -1,7 +1,7 @@
 //! Traits for handling hash to curve.
 
 use super::{ExpandMsg, MapToCurve, hash_to_field};
-use crate::{ProjectivePoint, Result};
+use crate::Result;
 use hybrid_array::typenum::Unsigned;
 
 /// Hash arbitrary byte sequences to a valid group element.
@@ -32,7 +32,7 @@ pub trait GroupDigest: MapToCurve {
     ///
     /// [`ExpandMsgXmd`]: crate::hash2curve::ExpandMsgXmd
     /// [`ExpandMsgXof`]: crate::hash2curve::ExpandMsgXof
-    fn hash_from_bytes<X>(msg: &[&[u8]], dst: &[&[u8]]) -> Result<ProjectivePoint<Self>>
+    fn hash_from_bytes<X>(msg: &[&[u8]], dst: &[&[u8]]) -> Result<Self>
     where
         X: ExpandMsg<Self::K>,
     {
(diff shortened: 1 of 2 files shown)


tarcieri:
cc @daxpedda

daxpedda:
In #1177 the point was to make it possible to use the types provided in `elliptic-curve` (`PublicKey`, `SecretKey` and so on) with `curve25519-dalek` as well. I don't know the details, but apparently `ed448-goldilocks` is able to implement `Curve` and `CurveArithmetic` for both Ed448 and Decaf448 (stalled on https://github.com/RustCrypto/elliptic-curves/issues/1229), so I assume the same can eventually be done in `curve25519-dalek`.

Regarding the change: I do think that implementing `MapToCurve` directly on `ProjectivePoint` is indeed more appropriate because it only deals with `ProjectivePoint` types. However the same can't be said for `GroupDigest` because it also acts on `Scalar`. So I do think `GroupDigest` should remain on the curve label itself. I don't know a trait that can facilitate this apart from `CurveArithmetic`.

In the end however this is just a preference and while I don't want to approve the change I'm not against it either. Either way works in practice.

carloskiki:
Indeed, we can implement the `Curve` traits for `curve25519-dalek`, I have a draft implementation of it already, but I don't know if we should.

> However the same can't be said for GroupDigest because it also acts on Scalar.

The `Group` trait has an associated `Scalar` type, which fulfills everything we need for `GroupDigest`. If you look at `CurveArithmetic`:

```rust
pub trait CurveArithmetic: Curve {
    // type AffinePoint: ...;

    type ProjectivePoint: group::Group<Scalar = Self::Scalar> /* + ... */;

    // type Scalar: ...;
```
We require that `CurveArithmetic::Scalar` _is_ `Group::Scalar` of the projective point. So the implementation in this PR is correct in using `Group::Scalar` in place of `CurveArithmetic::Scalar`.

daxpedda:
> > However the same can't be said for GroupDigest because it also acts on Scalar.
> 
> The `Group` trait has an associated `Scalar` type, which fulfills everything we need for `GroupDigest`.

I understand that its possible. What I was trying to say is that its quite strange to go through `ProjectivePoint` to call `hash_to_scalar()`.

daxpedda:
How about we just split `GroupDigest` into two traits: one to implement on `ProjectivePoint` and the other on `Scalar`? I honestly think this is the most appropriate given how the end-user actually ends up using the API.

I guess this PR can be closed and moved to `elliptic-curves` now that `hash2curve` is extracted in its own crate.

tarcieri:
Closing because `hash2curve` is now in https://github.com/RustCrypto/elliptic-curves but we can still discuss this as a possible change if you want to reopen a new PR

PR #1940, opened 2025-07-24
digest: implement `Zeroize` for the buffering newtypes

baloo:
This is a fixup to #1799


baloo:
cc @newpavlov 
I believe that fell through https://rustcrypto.zulipchat.com/#narrow/channel/260036-general/topic/sha2.3A.3ASha512.20no.20longer.20implements.20ZeroizeOnDrop/near/527569795

digest/src/buffer_macros/fixed.rs
```diff
@@ -39,6 +39,21 @@ macro_rules! buffer_fixed {
...
         }
+
+        impl$(< $( $lt $( : $clt $(+ $dlt )* )? ),+ >)? Drop for $name$(< $( $lt ),+ >)? {
+            #[inline]
+            fn drop(&mut self) {
+                #[cfg(feature = "zeroize")]
+                {
+                    use $crate::zeroize::Zeroize;
+                    self.core.zeroize();
+                    self.buffer.zeroize();
```
newpavlov: I don't think we need this explicit `Drop` impl. You just need to verify that both `core` and `buffer` implement `ZeroizeOnDrop`. It would also make the `decompose` changes unnecessary.

In other words, I think it's better to move the `ZeroizeOnDrop` implementation into a dedicated macro block (with checks on `core` and `buffer`) and add it to the `FixedHashTraits` list.

newpavlov (changes requested):
Thank you! I indeed forgot about it...

digest/src/buffer_macros/fixed.rs
baloo: I've pushed a fix for that, but now I'll need a review on https://github.com/RustCrypto/utils/pull/1192 ^^

digest/src/buffer_macros/fixed.rs
```diff
@@ -476,5 +476,30 @@ macro_rules! buffer_fixed {
...
+    };
+
+    // Implements `ZeroizeOnDrop`
+    (
+        impl_inner: $name:ident
+        $(< $( $lt:tt $( : $clt:tt $(+ $dlt:tt )* )? ),+ >)?
+        ($core_ty:ty);
+        ZeroizeOnDrop $($trait_name:ident)*;
+    ) => {
+        // Verify that `$core_ty` implements `ZeroizeOnDrop`
```
newpavlov: Verify that `$core_ty` and `Buffer<$core_ty>` implement `ZeroizeOnDrop`

PR #1941, opened 2025-08-01
Update ed25519-dalek repo link

leif-hesperides:
old link: https://github.com/dalek-cryptography/ed25519-dalek
new link: https://github.com/dalek-cryptography/curve25519-dalek


tarcieri:
Thanks for noticing this. I went with #1943 instead

PR #1948, opened 2025-08-05
Only run CI on latest commit of PR branches.

carloskiki:
This is a little hack I make use of in my projects and found a bit annoying when opening PRs here. When successive commits are made to a PR, cancels CI workflows that are not ran on the latest commit - saving CI time and making things quicker.

I can also make this PR to other RustCrpyto repos if this is something we want.

### DIFF
--- a/.github/workflows/aead.yml
+++ b/.github/workflows/aead.yml
@@ -16,8 +16,7 @@ env:
   CARGO_INCREMENTAL: 0
   RUSTFLAGS: "-Dwarnings"
   
-# If new code is pushed to a PR branch, then cancel in progress workflows for that PR. Ensures that
-# we don't waste CI time, and returns results quicker https://github.com/jonhoo/rust-ci-conf/pull/5
+# Cancels CI jobs when new commits are pushed to a PR branch
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
--- a/.github/workflows/cipher.yml
+++ b/.github/workflows/cipher.yml
@@ -17,8 +17,7 @@ env:
   CARGO_INCREMENTAL: 0
   RUSTFLAGS: "-Dwarnings"
 
-# If new code is pushed to a PR branch, then cancel in progress workflows for that PR. Ensures that
-# we don't waste CI time, and returns results quicker https://github.com/jonhoo/rust-ci-conf/pull/5
+# Cancels CI jobs when new commits are pushed to a PR branch
 concurrency:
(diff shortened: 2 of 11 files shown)


carloskiki:
Works!

newpavlov:
Are you sure that it does not influence the master branch? In the case if we merge two PRs in short succession, we don't want to cancel CI for the first merge.

carloskiki:
This won't affect push and merge to any branches on the repo.

> Use `concurrency` to ensure that only a single job or workflow using the same concurrency group will run at a time.

from: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#concurrency.

The concurrency group is defined by
```
group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
```

When pushing/merging to a branch on this repo, `github.head_ref = ""`:

> The head_ref or source branch of the pull request in a workflow run. This property is only available when the event that triggers a workflow run is either pull_request or pull_request_target.

So, the group will be `${{ github.workflow }}-${{ github.run_id }}`, where `run_id` is unique for every workflow run. Since all jobs are in unique groups (because of `run_id`), they will run in parallel (won't be interrupted). Only for PRs, the group will be the same for two commits to the same PR, meaning older jobs will be `cancel(ed)-in-progress`.

.github/workflows/aead.yml
```diff
@@ -15,6 +15,12 @@ defaults:
 env:
   CARGO_INCREMENTAL: 0
   RUSTFLAGS: "-Dwarnings"
+  
+# If new code is pushed to a PR branch, then cancel in progress workflows for that PR. Ensures that
+# we don't waste CI time, and returns results quicker https://github.com/jonhoo/rust-ci-conf/pull/5
```
newpavlov: Replace these comment with something like this:
>Cancels CI jobs when new commits are pushed to a PR branch

newpavlov (approved):
Sounds good to me then. Assuming @tarcieri does not have objections, I think we can merge it. For consistency we probably should also adopt this approach in other repos as well.

tarcieri:
Yeah this is a big problem and something I do manually sometimes, so doing it in an automated manner sounds great

baloo:
`signatures` and `formats` are two places where this could be beneficial too.

carloskiki:
@newpavlov will make a PR to the other crates later today.

.github/workflows/aead.yml
carloskiki: Replaced in https://github.com/RustCrypto/traits/pull/1948/commits/cecda9f66aeae62b47d2f476e207ba5409cd51cd.

PR #1972, opened 2025-08-24
Allow `serde` without `alloc`

### LINKED ISSUE #1960: Does `elliptic-curve` with only the `serde` feature enabled still require `alloc`?
Hi,

We're trying to use some primitives from the `p256` crate on an embedded target and would like to rely on the `serde` impls it provides, but sadly this transitively pulls in an `alloc` requirement that we cannot and do not want to satisfy.

Preliminary checks suggest that the problem is that `p256/serde` [activates `elliptic-curve/serde`](https://github.com/RustCrypto/elliptic-curves/blob/master/p256/Cargo.toml#L58), which in turn unconditionally [activates its `alloc` feature](https://github.com/RustCrypto/traits/blob/master/elliptic-curve/Cargo.toml#L70).

Is there a way to get rid of this requirement so that we can use the generated `serde` impls for `Signature` without being forced to draw in `alloc`?

Thanks,

Jona

datdenkikniet:
Fixes #1960 

I've tried my best to dig into the associated parts here, and it seems that the `der::Document` encoding used as intermediate step is the culprit.

Sadly I don't have enough insight to validate my TODO's. Will leave separate comments on those that I hope RustCrypto maintainers have better answers to.


elliptic-curve/src/public_key.rs
```diff
@@ -436,6 +458,7 @@ where
         let public_key_bytes = self.to_encoded_point(false);
         let subject_public_key = der::asn1::BitStringRef::new(0, public_key_bytes.as_bytes())?;
 
+        // TODO: use `encode_as_der` here?
```
datdenkikniet, line 461: Using `encode_as_der` seemed like a hassle here. Is there a better approach than `encode_as_der` followed directly by `Document::decode().unwrap()`?

Also suffers from the encoding length question, and this specific conversion will require `alloc` anyways (because `der::Document` contains a `Vec<u8>`)

elliptic-curve/src/public_key.rs
```diff
@@ -483,7 +506,11 @@ where
     where
         S: ser::Serializer,
     {
-        let der = self.to_public_key_der().map_err(ser::Error::custom)?;
+        // TODO: can we determine DER encoding length up-front? Using `MockCurve` gives
```
datdenkikniet, line 509: Is there a way to collect this information from some associated const? Was hard to figure out

PR #1974, opened 2025-09-01
aead: Update to heapless 0.9.1

sosthene-nitrokey:
This patch updates the heapless dependency of `aead` to version 0.9.1 It updates the buffer implementation to support for the new `View` types and the `LenType` optimization

More info here: https://blog.rust-embedded.org/heapless-091/

### DIFF
--- a/.github/workflows/aead.yml
+++ b/.github/workflows/aead.yml
@@ -39,13 +39,30 @@ jobs:
         with:
           toolchain: ${{ matrix.rust }}
           targets: ${{ matrix.target }}
-      - run: cargo check --all-features
+      - run: cargo check --features arrayvec,blobby,bytes,alloc,dev,os_rng
       - run: cargo build --target ${{ matrix.target }} --release --no-default-features
       - run: cargo build --target ${{ matrix.target }} --release --no-default-features --features alloc
       - run: cargo build --target ${{ matrix.target }} --release --no-default-features --features bytes
-      - run: cargo build --target ${{ matrix.target }} --release --no-default-features --features heapless
       - run: cargo build --target ${{ matrix.target }} --release --no-default-features --features rand_core
 
+  build-heapless:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        rust:
+          - 1.87.0 # MSRV
+          - stable
+        target:
+          - thumbv7em-none-eabi
+          - wasm32-unknown-unknown
+    steps:
+      - uses: actions/checkout@v5
+      - uses: RustCrypto/actions/cargo-cache@master
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: ${{ matrix.rust }}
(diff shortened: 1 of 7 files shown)


sosthene-nitrokey:
Current MSRV of `heapless` 0.9.1 is 1.87.

newpavlov:
It may be worth to remove `heapless` support from `aead` v0.6.0 to preserve MSRV 1.85 and add it back in v0.6.1.

tarcieri:
Alternatively we can just document that the `heapless` feature requires an MSRV of 1.87

sosthene-nitrokey:
> It may be worth to remove `heapless` support from `aead` v0.6.0 to preserve MSRV 1.85 and add it back in v0.6.1.

We could also rename the current `heapless` feature to `heapless-v0.8` and add a `heapless-v0.9` feature now or in the future.

I also think that keeping the `MSRV` at 1.85 for default features should be fine. Cargo should show a nice error message if heapless is brought into the dependency tree (for that we need a new 0.9.2 release of heapless that documents the MSRV in the `Cargo.toml` but that can be done).

newpavlov:
>Alternatively we can just document that the heapless feature requires an MSRV of 1.87

This would go against the new MSRV policy: https://github.com/RustCrypto/meta/issues/26

I don't think we should bother with supporting both heapless v0.8 and v0.9.

tarcieri:
It seems a little weird/confusing to me to remove and add back a feature just for the sake of that policy.

Perhaps we could document an exception for `heapless` upon initial release instead, which we could remove and bump MSRV afterwards?

newpavlov:
Introducing ad hoc exceptions to the decided policy is weirder in my opinion. I would prefer to release `aead` v0.6.0 with MSRV 1.87 (despite 1.85 being supplied in Debian) than to make such exception.

I think it will be easier for users to see the tradeoff between MSRV and `heapless` support in the `aead` changelog than to observe weird errors of not being able to find a compatible `heapless` version.

tarcieri:
> despite 1.85 being supplied in Debian

Oh, is that what the next Debian release is targeting? I guess that's a pretty good reason to release everything as MSRV 1.85 initially

newpavlov:
>Oh, is that what the next Debian release is targeting? 

Yes, though it's not next, but already the latest stable release. :) https://packages.debian.org/trixie/rust/rustc

sosthene-nitrokey:
Releasing with MSRV of 1.87 in the Cargo.toml will break users with 1.85 trying to use `aead` without `heapless`.

newpavlov:
No, it will not. MSRV-aware resolver will not select v0.6.1 for such users.

sosthene-nitrokey:
Ok that's why you want 0.6.0 without heapless support. I understand that.

tarcieri:
Yeah I'm fine with that approach too

newpavlov:
Closing in favor of tracking issue #2000.

PR #1975, opened 2025-09-01
aead: Implement Buffer for &mut impl Buffer

sosthene-nitrokey:


tarcieri:
What's the use case?

sosthene-nitrokey:
With `heapless` 0.9.1, there is the `&mut VecView<u8>` which one might want to use as a `dyn Buffer`. However, `VecView` is `!Sized`, so `&mut VecView<u8>` can't coerce to a `&mut dyn Buffer`. With this `&mut &mut VecView<u8>` can be converted to a `&mut dyn Buffer`. I encountered this when making a wrapper type in our own library: https://github.com/trussed-dev/trussed-staging/blob/e26f6ae5846f92fe1023c44fa2e9116a05d1aebd/src/chunked/mod.rs#L102

We could also implement `Buffer` directly on `&mut VecView<u8>` instead in https://github.com/RustCrypto/traits/pull/1974

tarcieri:
@sosthene-nitrokey aah, the trait object hack. Yeah I've kind of soured on that sort of thing specifically because it's counterintuitive / clever-in-a-bad-way and we can probably change the relevant APIs to accept `?Sized` since I don't think there's any reason they can't be. We went through something similar with `CryptoRng`(Core) where the mut-ref-impls-trait hack was supposed to be the solution for trait objects, but it just confused people, so we added (back) `?Sized` everywhere so it's clear the APIs accept trait objects

> We could also implement Buffer directly on &mut VecView<u8> instead in https://github.com/RustCrypto/traits/pull/1974

Sounds good to me

tarcieri:
@sosthene-nitrokey if you'd like to submit a PR to add `?Sized` bounds where `Buffer` so trait objects can be used, that'd be great

sosthene-nitrokey:
The issue is not missing `?Sized` bounds. It's that a `!Sized` type cannot coerce to `dyn Buffer` because then the pointer would need to be a "fat fat" pointer (i.e. the address, the vtable pointer, the size of the slice).

PR #1983, opened 2025-09-02
cipher: restore `StreamCipherCoreWrapper::from_core`

tarcieri:
Removed in #1959 but there are definitely still quite a few usages and I don't see ways to completely replace all of them with `KeyIvInit` (though that seems ideal).

See RustCrypto/AEADs#710


tarcieri:
This seems sufficient to unblock https://github.com/RustCrypto/AEADs/pull/710 which is now green in CI with a patch to use this branch so I'm going to go ahead and merge.

@newpavlov I'm fine with removing `from_core`, but `belt-dwp` and `ccm` in AEADs need to be updated accordingly. It seems like they can probably be updated to use the `KeyIvInit` impl, but offhand I'm not sure why it's not working and would rather complete the `hybrid-array` upgrade than investigate that for now

PR #1987, opened 2025-09-02
crypto-common v0.2.0-rc.4

tarcieri:
Notably includes `hybrid-array` v0.4 support


tarcieri:
Note: I'll do a batch release for the rest of the crates (with a possible exception of `elliptic-curve`) but I wanted to make sure this one got bumped appropriately as it's the common dependency

PR #1996, opened 2025-09-04
Add `AffineCoordinates::decode()`

### LINKED ISSUE #1994: Uncompressed Point Deserialization
The main motivation here is to allow `primeorder` dependents, e.g. `p256`, to construct `AffinePoint`s from `FieldElement`s. Currently this is a problem with hash2curve, where we have to [go through `AffinePoint::decompress()`](https://github.com/RustCrypto/elliptic-curves/blob/c9f8b2274fa90b755c5708185cf844f3a6bcb658/p256/src/arithmetic/hash2curve.rs#L68-L71).

However, this also overlaps with https://github.com/dalek-cryptography/curve25519-dalek/issues/817. So the proposal should be quite straightforward:
```rust
pub trait DecodePoint<C: Curve>: Sized {
    fn decode(x: &FieldBytes<C>, y: &FieldBytes<C>) -> CtOption<Self>;
}
```
Then expose a const method on `AffinePoint`s themselves.

However, specifically for hash2curve, we will need an unchecked method to avoid the additional overhead. So I propose adding another trait:
```rust
pub trait UncheckedDecodePoint<C: Curve>: Sized {
    fn decode(x: C::FieldElement, y: C::FieldElement) -> Self;
}
```
I believe this would not be publicly accessible anyway, because users don't have access to `FieldElement` directly.

Happy to make a PR upon approval.

daxpedda:
As discussed this PR adds the new API `AffineCoordinates::decode()`.

Resolves #1994.


elliptic-curve/src/point.rs
```diff
@@ -21,10 +21,13 @@ pub type ProjectivePoint<C> = <C as CurveArithmetic>::ProjectivePoint;
 
 /// Access to the affine coordinates of an elliptic curve point.
 // TODO: use zkcrypto/group#30 coordinate API when available
-pub trait AffineCoordinates {
+pub trait AffineCoordinates: Sized {
     /// Field element representation with curve-specific serialization/endianness.
     type FieldRepr: AsRef<[u8]>;
 
+    /// Creates an affine point from its coordinates.
+    fn decode(x: &Self::FieldRepr, y: &Self::FieldRepr) -> CtOption<Self>;
```
tarcieri: It's not really "decoding" so much as validating that the field elements are a valid solution to the curve equation.

Maybe `from_coordinates`?

daxpedda: True enough.

> Maybe `from_coordinates`?

I don't have a better suggestion either. Will go with just that.

tarcieri:
Can merge when the https://github.com/RustCrypto/elliptic-curves PR is ready

PR #1999, opened 2025-09-05
aead: remove `heapless` support

newpavlov:
`heapless` v0.9 has MSRV 1.87 while we would like to keep MSRV 1.85 for the initial aead v0.6.0 release. We plan to add support in v0.6.1 immediately after v0.6.0 release.

### DIFF
--- a/aead/CHANGELOG.md
+++ b/aead/CHANGELOG.md
@@ -7,15 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 ### Fixed
-- minor documentation error in `AeadCore::TagSize` ([#1351])
-- fixup `hybrid-array` migration ([#1531])
+- Minor documentation error in `AeadCore::TagSize` ([#1351])
+- Fixup `hybrid-array` migration ([#1531])
 
 ### Changed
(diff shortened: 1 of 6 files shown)


aead/CHANGELOG.md
```diff
@@ -24,14 +23,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Bump `crypto-common` to v0.2.0-pre.5 ([#1496])
 
 ### Added
-- enable `missing_debug_implementations` lint and add `Debug` impls ([#1411])
+- Enable `missing_debug_implementations` lint and add `Debug` impls ([#1411])
+
+### Removed
+- `heapless` support ([#1999])
```
tarcieri: Perhaps note it will be added back in a future release?

newpavlov:
Note that we would have to remove the `heapless` features from implementation crates as well. Maybe we should not forward `arrayvec` and `heapless` features in implementation crates?

tarcieri:
If nothing else, the docs depend on the feature to be doctestable

newpavlov:
Ah, true. I missed that we have such doctests.

newpavlov:
The dev macros will be fixed later as part of https://github.com/RustCrypto/AEADs/issues/716

PR #2004, opened 2025-09-12
signature: use `Fn(&mut D)` for `*DigestSigner`/`*DigestVerifier`

### LINKED ISSUE #2003: signature: change `DigestSigner`/`DigestVerifier` to be ML-DSA "external mu" compatible?
Originally `signature` had `DigestSigner`/`DigestVerifier` traits then added `hazmat::{PrehashSigner, PrehashVerifier}` trait. The original goal of the `DigestSigner`/`DigestVerifier` traits was to leverage type safety to ensure that the input is always a hash. We later added the prehash traits out of necessity, especially observing people utilizing hacks like impl'ing pseudo-`Digest`s that emit the hash to work around the constraints of the API in environments where prehashes were being used but computed significantly earlier than the signature function.

The two traits end up being quite similar to the point I've thought about removing `DigestSigner`/`DigestVerifier` due to the overlap, which would decouple `signature` from the `digest` crate entirely. But instead of that, I think we can do a slight tweak to the API which would make it possible to impl these traits (but not `hazmat::{PrehashSigner, PrehashVerifier}`), in an ML-DSA external mu-compatible way. Here's a simplified version:

```rust
pub trait DigestSigner<D: Digest, S> {
    fn sign_digest<F: Fn(&mut D)>(&self, f: F) -> S;
}
```

...so instead of `DigestSigner` and `DigestVerifier` taking a user-supplied digest instance as they do today, the signer/verifier type which impls this trait instead initializes the provided `Digest` type themselves, then supply it to a callback function which is expected to perform a series of `update`s. This means the initialization can include hashing a leading message prefix, which is what's needed to implement an IUF-like API for external mu.

Such an API should work for all existing users of the `signature::Digest*` traits as well as ML-DSA.

cc @daxpedda

daxpedda:
I honestly tried to find the right words for the documentation, but it all sounds rather off to me.

My suspicion is that we will actually need `FnMut` here if users intend to use some `Reader` API inside.

Companion PR: https://github.com/RustCrypto/signatures/pull/1064.
Resolves #2003.

### DIFF
--- a/signature/src/signer.rs
+++ b/signature/src/signer.rs
@@ -3,7 +3,7 @@
 use crate::error::Error;
 
 #[cfg(feature = "digest")]
-use crate::digest::Digest;
+use crate::digest::Update;
 
 #[cfg(feature = "rand_core")]
 use crate::rand_core::{CryptoRng, TryCryptoRng};
@@ -55,7 +55,7 @@ pub trait SignerMut<S> {
     fn try_sign(&mut self, msg: &[u8]) -> Result<S, Error>;
 }
 
-/// Sign the given prehashed message [`Digest`] using `Self`.
+/// Sign the given prehashed message `Digest` using `Self`.
 ///
 /// ## Notes
 ///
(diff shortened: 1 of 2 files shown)


signature/src/signer.rs
```diff
@@ -234,9 +242,10 @@ pub trait AsyncDigestSigner<D, S>
 where
     D: Digest,
 {
-    /// Attempt to sign the given prehashed message [`Digest`], returning a
-    /// digital signature on success, or an error if something went wrong.
-    async fn sign_digest_async(&self, digest: D) -> Result<S, Error>;
+    /// Attempt to sign a message by updating the received [`Digest`] with it,
+    /// returning a digital signature on success, or an error if something went wrong.
+    async fn sign_digest_async<F: Fn(&mut D) -> Result<(), Error>>(&self, f: F)
```
tarcieri:
```suggestion
    async fn sign_digest_async<F: AsyncFn(&mut D) -> Result<(), Error>>(&self, f: F)
```

daxpedda: Ooops! Fixed.

signature/src/verifier.rs
```diff
@@ -38,11 +38,17 @@ pub trait MultipartVerifier<S> {
...
+/// than a raw digest value.
 ///
 /// [Fiat-Shamir heuristic]: https://en.wikipedia.org/wiki/Fiat%E2%80%93Shamir_heuristic
 #[cfg(feature = "digest")]
 pub trait DigestVerifier<D: Digest, S> {
-    /// Verify the signature against the given [`Digest`] output.
-    fn verify_digest(&self, digest: D, signature: &S) -> Result<(), Error>;
+    /// Verify the signature against the received [`Digest`] output,
+    /// by updating it with the message.
+    fn verify_digest<F: Fn(&mut D) -> Result<(), Error>>(
```
daxpedda, line 52: Using `Result` here because the function is already fallible and we don't offer a `try_*` equivalent.

tarcieri:
I think it would be good to add some comments about the expected contract of this function, namely that it should expect to perform all of its updates in a single invocation, and if called repeatedly it should perform the same sequence of updates

daxpedda:
Update for the signatures repo is up: https://github.com/RustCrypto/signatures/pull/1064.

signature/src/signer.rs
```diff
@@ -55,7 +55,7 @@ pub trait SignerMut<S> {
     fn try_sign(&mut self, msg: &[u8]) -> Result<S, Error>;
 }
 
-/// Sign the given prehashed message [`Digest`] using `Self`.
+/// Sign the given prehashed message `Digest` using `Self`.
```
daxpedda, line 58: Now that we switched to `Update`, does using `Digest` in the documentation still make sense or should I use some other term?

tarcieri: I think it's fine to keep using `Digest` in the documentation. The difference between a digest and a XOF is not something end users need to really concern themselves with. And most of the generic impls can probably stick to using a `D: Digest` bound.

daxpedda:
FYI because `Digest` isn't a supertrait of `Update` I kinda did the implementation with `FixedOutput + Update `. But I could also write `Digest + Update`.

Also just noticing that if we actually do `FixedOutput + Update` we should also add `HashMarker`? Should I add `HashMarker` in the trait here?

EDIT: Went with `Digest + Update` for now.

tarcieri:
@daxpedda this is kind of annoying re: `Digest + Update`:

```
error[E0034]: multiple applicable items in scope
   --> k256/src/schnorr/verifying.rs:119:53
    |
119 |                 msg.iter().for_each(|&slice| digest.update(slice));
    |                                                     ^^^^^^ multiple `update` found
    |
    = note: candidate #1 is defined in an impl of the trait `Digest` for the type `D`
    = note: candidate #2 is defined in an impl of the trait `Update` for the type `Sha256`
```

Maybe it should just be `Default + Update + FixedOutput`? (which AFAICT would also work for Shake256`)

daxpedda:
Yeah, we can just use `Default + FixedOutput + Update`, but in implementations. I don't believe Shake supports `FixedOutput`, where we can just skip it.

I mentioned this before, but we might want to add `HashMarker` as well?

tarcieri:
Aah, I see, it looks like it did in `sha3` v0.10 but not in v0.11. Sure, `HashMarker` sounds good.

PR #2006, opened 2025-09-12
elliptic-curve: change `Curve::ORDER` to be `Odd`

tarcieri:
Previously it was `NonZero`.

This associated constant is for the curve's prime order subgroup, which will always be odd.

There are many modular algorithms which require the modulus is odd, and bounding the `ORDER` on `Odd` ensures those algorithms will be compatible with the scalar field.

The `primefield` crate and the params we are computing for every field of every curve are also `Odd` (i.e. the `modulus` of `MontyParams`).


tarcieri:
Companion PR: https://github.com/RustCrypto/elliptic-curves/pull/1416

PR #2014, opened 2025-09-15
digest: move `EagerHash` from `hmac`

tarcieri:
Originally added in https://github.com/RustCrypto/MACs/pull/151

The situation is `ecdsa` has a trait for the "default" hash function to use for a particular curve, which we want to work with RFC6979 and ergo HMAC, so we would like to use an `EagerHash` bound.

But `rfc6979` is an optional dependency, so we can't depend on its transitive accessibility to `rfc6979::hmac`. This PR added `hmac` as a dependency gated on the `digest` feature of `ecdsa`:

https://github.com/RustCrypto/signatures/pull/1076

But as I was looking at `EagerHash` I noticed there's nothing HMAC-specific about it at all and it can easily be moved to `digest`, so we're able to use it in bounds without any other dependencies besides `digest`, which would be nice.

cc @daxpedda

### DIFF
--- a/digest/src/block_api.rs
+++ b/digest/src/block_api.rs
@@ -3,7 +3,7 @@
 //! Usage of traits in this module in user code is discouraged. Instead use
 //! core algorithm wrapped by the wrapper types, which implement the
 //! higher-level traits.
-use crate::InvalidOutputSize;
+use crate::{Digest, HashMarker, InvalidOutputSize};
 
 pub use block_buffer::{Eager, Lazy};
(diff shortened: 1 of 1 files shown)


daxpedda:
LGTM!

Can do follow-ups tomorrow if they aren't already done by then.

daxpedda:
Let me know if you want me to update any other repo.

tarcieri:
@daxpedda let me know if there are any remaining PRs I haven't reviewed

PR #2020, opened 2025-09-20
chore(deps): bump blobby to 0.4.0-pre.1, remove patch.crates-io

baloo:


newpavlov:
You need to also update Cargo.lock

PR #2023, opened 2025-09-28
elliptic-curve: add `point::LookupTable` and `BasepointTable`

tarcieri:
Implementation is originally from the `k256` crate, but written to be generic over a `Point` type which is bound by the `Group` trait.

Adds a `LookupTable` type which supports precomputed lookup tables for a given curve point which can be selected from in constant time.

Also adds a feature-gated `BasepointTable` type with a `const fn` constructor which builds a table for precomputed fixed-base scalar multiplication using the `Group::generator` as the base point.

The `const fn` support for `BasepointTable` is achieved via a `LazyLock`-style pattern which computes the table lazily upon first use in a synchronized fashion. It supports using either `std::sync::LazyLock`, or for `no_std` targets it also supports using the `critical-section` feature of the `once_cell` crate.

### DIFF
--- a/.github/workflows/elliptic-curve.yml
+++ b/.github/workflows/elliptic-curve.yml
@@ -45,6 +45,7 @@ jobs:
       - run: cargo build --target ${{ matrix.target }} --release --no-default-features --features alloc
       - run: cargo build --target ${{ matrix.target }} --release --no-default-features --features arithmetic
       - run: cargo build --target ${{ matrix.target }} --release --no-default-features --features bits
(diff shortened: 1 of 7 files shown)


tarcieri:
I originally tried extracting this into `primeorder` in https://github.com/RustCrypto/elliptic-curves/pull/1475 but IMO that's not a good fit for it, as the implementation is generic such that it can support any elliptic curve, not just prime order curves, and is otherwise completely independent from everything `primeorder` provides. So `elliptic-curve` seems like a better place.

cc @fjarri

PR #2049, opened 2025-10-25
kem: use `TryCryptoRng` for `encapsulate`

tarcieri:
The method is fallible anyway, so errors should be propagated rather than panicking

### DIFF
--- a/kem/src/lib.rs
+++ b/kem/src/lib.rs
@@ -10,7 +10,7 @@
 #![warn(missing_docs, unused_qualifications, missing_debug_implementations)]
 
 use core::fmt::Debug;
-use rand_core::CryptoRng;
+use rand_core::TryCryptoRng;
 
 /// A value that can be encapsulated to. Often, this will just be a public key. However, it can
 /// also be a bundle of public keys, or it can include a sender's private key for authenticated
@@ -20,7 +20,7 @@ pub trait Encapsulate<EK, SS> {
     type Error: Debug;
 
     /// Encapsulates a fresh shared secret
-    fn encapsulate<R: CryptoRng + ?Sized>(&self, rng: &mut R) -> Result<(EK, SS), Self::Error>;
+    fn encapsulate<R: TryCryptoRng + ?Sized>(&self, rng: &mut R) -> Result<(EK, SS), Self::Error>;
 }
 
 /// A value that can be used to decapsulate an encapsulated key.


tarcieri:
Ugh, tried to do a downstream PR for this and it ended up being rather annoying. I'll leave this open for now but I'm a little confused about what to do about cases where we still need to pass a `TryCryptoRng` as a `CryptoRng` (`unwrap_err` was complaining about `?Sized` bounds)

newpavlov:
>I'm a little confused about what to do about cases where we still need to pass a TryCryptoRng as a CryptoRng

Why do you need to do it in implementation crates? With this change you should be able to pass `CryptoRng` into `encapsulate`, but I am not sure why you need to do it in the other direction. Users may wrap RNG in `UnwrapErr` or `UnwrapMut`, but it should not be our concern.

kem/src/lib.rs
```diff
@@ -20,7 +20,7 @@ pub trait Encapsulate<EK, SS> {
     type Error: Debug;
 
     /// Encapsulates a fresh shared secret
-    fn encapsulate<R: CryptoRng + ?Sized>(&self, rng: &mut R) -> Result<(EK, SS), Self::Error>;
+    fn encapsulate<R: TryCryptoRng + ?Sized>(&self, rng: &mut R) -> Result<(EK, SS), Self::Error>;
```
newpavlov, line 23: It may be worth to add a note to the method docs that users can pass types which implement `CryptoRng` into this method.

tarcieri:
Because all of the downstream APIs use `CryptoRng`, including ones in `x25519-dalek`

newpavlov:
>Because all of the downstream APIs use CryptoRng

Maybe you meant "upstream" here? So you want to use `x25519-dalek` which operates over `CryptoRng` to implement the KEM trait which we would like to migrate to `TryCryptoRng`, correct?

Have you tried to use [`UnwrapMut`](https://docs.rs/rand_core/latest/rand_core/struct.UnwrapMut.html)? Although, ideally, it would be better to update the upstream dependencies to support `TryCryptoRng`.

tarcieri:
It’s for `x-wing` (a hybrid) as well as `dhkem` which wraps several ECC crates, but that’s more or less the idea.

I tried `unwrap_err`, I can try `unwrap_mut` I guess

tarcieri:
Okay, `unwrap_mut` worked and I have a PR to update everything to use the new API, so I'll go ahead and merge this

PR #2052, opened 2025-10-29
cipher: fix seeking implementation in the stream cipher wrapper

### LINKED ISSUE #1808: `chacha20`: Seeking near the end of the keystream causes `try_current_pos` to return an error

newpavlov:
The change forbids seeking to the last keystream block and its application. Additionally, unchecked methods added as an escape hatch.

Closes #1808

### DIFF
--- a/cipher/src/stream.rs
+++ b/cipher/src/stream.rs
@@ -97,8 +97,8 @@ pub trait AsyncStreamCipher: Sized {
 /// making the cipher vulnerable to chosen plaintext attack. Typically, the repetition period is
 /// astronomically large, rendering keystream repetition impossible to encounter in practice.
 ///
-/// However, some counter-based stream ciphers use small counters (e.g. 32 bits) and allow seeking
-/// across the keystream. This can result in practical encounters with keystream repetition.
+/// However, counter-based stream ciphers allow seeking across the keystream, and some also use
+/// small counters (e.g. 32 bits). This can result in triggering keystream repetition in practice.
 ///
 /// To guard against this, methods either panic (e.g. [`StreamCipher::apply_keystream`]) or
 /// return [`StreamCipherError`] (e.g. [`StreamCipher::try_apply_keystream`]) when


tarcieri:
Is this really a fix for #1808? I would think a real fix would make the last keystream block usable

newpavlov:
See my [comment](https://github.com/RustCrypto/traits/issues/1808#issuecomment-3462868903) in the issue. I spent a significant time trying to implement the last byte approach, but I couldn't get a satisfying solution.

newpavlov:
Another alternative could be to add a special "exhausted" state to the wrapper (either as a separate field, or as part of `ReadBuffer`), but I don't quite like it either. Adding the field would make all buffered stream cipher implementations bigger (often by more than one byte because of alignment), while modifying `ReadBuffer` would mean that we need to check the special case every time. Also, I would like to keep the `block-buffer` code as simple as possible.

Considering that users arguably should not rely on the last keystream block in practice, I am not sure it's worth the trouble.

newpavlov:
@tarcieri 
Could you take a look at the updated docs?

tarcieri (approved):
Docs look good to me

PR #2055, opened 2025-10-30
kem: add associated `Encapsulator` to `Decapsulate`

tarcieri:
Otherwise there's no way to generically access one from the other

### DIFF
--- a/kem/src/lib.rs
+++ b/kem/src/lib.rs
@@ -38,5 +38,5 @@ pub trait Decapsulate<EK, SS> {
     fn decapsulate(&self, encapsulated_key: &EK) -> Result<SS, Self::Error>;
 
     /// Retrieve the encapsulator associated with this decapsulator.
-    fn to_encapsulator(&self) -> Self::Encapsulator;
+    fn encapsulator(&self) -> Self::Encapsulator;
 }


tarcieri:
cc @rozbb 

I feel like this is the biggest missing piece in the current `kem` API, and with it we could get rid of e.g. `ml_dsa::KemCore`.

Before merging this, I'll make a companion PR to https://github.com/rustcrypto/kems to see if there are any showstoppers.

Edit: opened https://github.com/RustCrypto/KEMs/pull/155 which impls these changes, it wasn't a problem

kem/src/lib.rs
```diff
@@ -28,9 +28,15 @@ pub trait Encapsulate<EK, SS> {
...
+    type Encapsulator: Encapsulate<EK, SS>;
+
     /// Decapsulation error
     type Error: Debug;
 
     /// Decapsulates the given encapsulated key
     fn decapsulate(&self, encapsulated_key: &EK) -> Result<SS, Self::Error>;
+
+    /// Retrieve the encapsulator associated with this decapsulator.
+    fn to_encapsulator(&self) -> Self::Encapsulator;
```
tarcieri: Bikeshed: this could be `encapsulator`?

In the concrete KEM impls there's `encapsulation_key`, but I worry that might get confused with `EK` (i.e. *encapsulated* key)

tarcieri:
Going to go ahead and merge this, but I'd appreciate any retroactive feedback

PR #2056, opened 2025-10-30
kem: add `KeyInit` support to `Decapsulate`

tarcieri:
This notably provides a common way to initialize KEM decapsulation keys along with a common way to do key generation

### DIFF
--- a/kem/src/lib.rs
+++ b/kem/src/lib.rs
@@ -8,6 +8,8 @@
 #![forbid(unsafe_code)]
 #![warn(missing_docs, unused_qualifications, missing_debug_implementations)]
 
+pub use crypto_common::{KeyInit, KeySizeUser};
+
 use core::fmt::Debug;
 use rand_core::TryCryptoRng;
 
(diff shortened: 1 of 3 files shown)


tarcieri:
cc @rozbb

kem/src/lib.rs
```diff
@@ -26,7 +28,7 @@ pub trait Encapsulate<EK, SS> {
 ///
 /// Often, this will just be a secret key. But, as with [`Encapsulate`], it can be a bundle
 /// of secret keys, or it can include a sender's private key for authenticated encapsulation.
-pub trait Decapsulate<EK, SS> {
+pub trait Decapsulate<EK, SS>: KeyInit {
```
newpavlov, line 31: Are you sure that `KeyInit` should be a super trait of `Decapsulate`? Couldn't `Decapsulate` in theory be implemented on top of a HSM?

tarcieri: Good point. I guess I'll close this, but perhaps we can use `KeyInit` by convention in the KEM crates.

newpavlov: It may be worth to re-export `KeyInit` from `kem` to promote its use.

PR #2057, opened 2025-10-30
kem: re-export `KeyInit` and `KeySizeUser` from `crypto-common`

tarcieri:
Encourages the use of these traits for initializing decapsulators that impl the `Decapsulate` type.

(This is effectively #2056 without the supertrait bound)


tarcieri:
cc @rozbb

PR #2063, opened 2025-11-04
Bump `rand_core` to v0.10.0-rc-2

tarcieri:
Since `OsRng` has been moved to `rand`, and `rand` depends on `chacha20` which depends on `cipher` -> `crypto-common` which needs access to `OsRng`, this temporarily vendors the relevant code (which is a thin wrapper around `getrandom`) until we can sort out a better solution for accessing `OsRng` without creating a circular dependency on `rand`.

cc @dhardy

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -138,8 +138,7 @@ dependencies = [
 [[package]]
 name = "crypto-bigint"
 version = "0.7.0-rc.9"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


tarcieri:
I guess another way to solve this would be to remove `OsRng` entirely and use `getrandom` directly

tarcieri:
#2064 switches (back) to using `getrandom` directly, which eliminates the `OsRng` dependency

PR #2064, opened 2025-11-04
aead+cipher+crypto-common: depend on `getrandom` directly

tarcieri:
Replaces usages of `OsRng` with `getrandom`, and `OsError` with `getrandom::Error`.

In `rand_core` v0.10 prereleases, `OsRng` has been moved to the `rand` crate (at least for now). The `rand` crate also depends on `chacha20`, so if we were to depend on `rand` it would create a circular dependency. We can avoid this circular dependency by using `getrandom` directly.

By changing the feature name back to `getrandom`, it also matches the existing feature name in stable releases, which is one less thing people need to change in updates.

Since `aead` was consuming `OsRng` via `crypto-common`, the latter has been changed to re-export `getrandom`.

See also: #2063

cc @dhardy

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -152,6 +152,7 @@ dependencies = [
 name = "crypto-common"
 version = "0.2.0-rc.4"
 dependencies = [
+ "getrandom",
  "hybrid-array",
  "rand_core",
 ]
--- a/aead/Cargo.toml
+++ b/aead/Cargo.toml
@@ -28,7 +28,7 @@ bytes = { version = "1", optional = true, default-features = false }
 default = ["rand_core"]
 alloc = []
 dev = ["blobby", "alloc"]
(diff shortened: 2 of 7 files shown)


tarcieri:
This is more or less where we were at after #1379. I'm not sure why we switched back to using `OsRng` instead? (I guess that was in #1746)

dhardy:
It doesn't look like anything here benefits from an RNG interface.

#1746 has `generate_*_with_rng` methods. I guess those were dropped because there is no reason not to use `getrandom`?

tarcieri:
The `(try_)generate_*_with_rng` APIs are retained in this PR, it's just the ones that use the ambient system RNG have been changed to use `getrandom` instead, as they were in #1379

tarcieri:
Going to merge this in order to make progress on #2063

PR #2079, opened 2025-11-06
password-hash: remove unnecessary expect

baloo:


tarcieri (approved):
Aah nice, I was wondering how to do that

baloo:
There is an even better syntax, but I could not remember it.

tarcieri:
The only thing I see better is `Result::into_ok` which is unstable

Issue #1897, opened 2025-06-08
crypto-common: fallible alternative to `KeyInit`


tarcieri:
I was looking at using `KeyInit` with `elliptic-curve`, but for that application we need to return an error in the event the deserialized integer for the key overflows the curve's order.

Perhaps there could be a fallible equivalent to `KeyInit` like `TryKeyInit`?

newpavlov:
Sounds good to me. I wonder whether we should add a blanket impl of `TrykeyInit` for `T: KeyInit`.

newpavlov:
@tarcieri 
Since you closed #2092, what do you plan to do with this issue? Should it be closed as "not planned"?

tarcieri:
Can make another attempt after #2096

newpavlov:
The trait was added in #2097.

PR #2085, opened 2025-11-09
elliptic-curve: `getrandom` feature

tarcieri:
Adds support for generating `SecretKey` and `NonZeroScalar` using the system's cryptographically secure random number generator. Notably this renames the former `SecretKey::random` and `NonZeroScalar::random` methods to `SecretKey::generate` and `NonZeroScalar::generate`, which take no parameters and are infallible.

This avoids the need for the user to import an `OsRng` type, or worry about the generation failing (which it won't on most notable modern OSes).

If a user still wants to handle RNG errors, the `try_from_rng` method still exists, and they can pass `OsRng` if they'd like.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -188,6 +188,7 @@ dependencies = [
  "crypto-bigint",
  "digest",
  "ff",
+ "getrandom",
  "group",
  "hex-literal",
  "hkdf",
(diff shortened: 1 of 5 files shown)


tarcieri:
I think #1897 could potentially obviate the need for this, but I wanted to have a stopgap in place first before exploring that as an alternative.

Issue #2036, opened 2025-10-13
generic-array 0.x deprecation warnings


Turbo87:
The `crypto-common` crate depends on `generic-array@0.14` which was deprecated a couple of hours ago (see https://github.com/fizyk20/generic-array/commit/d859e59b29412c8434563716e418884c6a9266b2).

Should the dependency be updated to `generic-array@1`?

I noticed in [some other places](https://github.com/RustCrypto/traits/pull/1462) we migrated from `generic-array` to `hybrid-array` instead. Would that be a better direction?

In any case, I supposed this will require a breaking change release for the `crypto-common` crate and probably all crates depending on it too, since `generic-array` is used in the public API of the `crypto-common` crate.

/cc @tarcieri

newpavlov:
We can not do anything on our side apart from releasing new crate versions which use `hybrid-array`. Also, see the `generic-array` issue: https://github.com/fizyk20/generic-array/issues/158

Turbo87:
> We can not do anything on our side apart from releasing new crate versions which use `hybrid-array`.

yeah, that was my suggestion. is there a reason why we shouldn't do that?

newpavlov:
There are some minor blockers which I would like to resolve before cutting new releases. Hopefully, it will be done before the year's end.

I believe that the best near-term solution would be to remove the deprecation warning from `generic-array` v0.14.

Turbo87:
I see, thanks

tarcieri:
@newpavlov seems like we're going to get a lot of these. It would probably be good to wrap things up sooner than later

tarcieri:
Reopened as suggested on #2041

Pr0methean:
When can we expect the version that fixes this, and what will be its MSRV? This is currently breaking https://github.com/zip-rs/zip2, and I'm not prepared to make all other dependencies harder to update by removing Cargo.lock from .gitignore (which isn't recommended for libraries anyway).

newpavlov:
We plan to release new versions before the year's end. MSRV will be equal to 1.85.

The easiest solution in the near term is to pin `generic-array` to v0.14.7 either by committing Cargo.lock (it works fine for libraries as you can see in our repos) or by adding `generic-array = "=0.14.7"` to your Cargo.toml. I know that the latter can be brittle and generally is not recommended, but it should work fine in this particular case.

oconnor663:
[...]

PR #2088, opened 2025-11-12
Pin generic-array to v0.14.7 and release crypto-common v0.1.7

newpavlov:
Workaround for #2036.


newpavlov:
The CI breakage is quite annoying and as far as I can see is not relevant to the pinning. Maybe we should ignore it?

PR #2089, opened 2025-11-12
crypto-common: add changelog entry for v0.1.7

newpavlov:


newpavlov:
@tarcieri 
It seems we forgot to add a changelog entry for migration to `hybrid-array`.

tarcieri:
@newpavlov I typically go through the commit history and write a curated changelog shortly before release

newpavlov:
I sometimes do the same, but I think ideally we should strive to add changelog entries in the same PR which introduces the changes.

tarcieri:
Probably a better topic for a meta post.

We could do that but also have a lot of changes which are reverted or later obviated by other changes, which need to be edited out of the final changelog

Issue #2060, opened 2025-10-31
crypto-common: `KeyInit` for public keys / separate keygen trait?


tarcieri:
I was looking at replacing the [`ml_kem::EncodedSizeUser`](https://docs.rs/ml-kem/0.2.1/ml_kem/trait.EncodedSizeUser.html) trait with `KeySizeUser` and `KeyInit`. This works fine for [`DecapsulationKey`](https://docs.rs/ml-kem/0.2.1/ml_kem/kem/struct.DecapsulationKey.html) (i.e. private keys), but it's a little bit weird for [`EncapsulationKey`](https://docs.rs/ml-kem/0.2.1/ml_kem/kem/struct.EncapsulationKey.html) since the [`KeyInit::generate_key*`](https://docs.rs/crypto-common/0.2.0-rc.4/crypto_common/trait.KeyInit.html#method.generate_key) methods don't make sense for public keys.

I was wondering if perhaps they should be moved to their own separate keygen trait.

newpavlov:
Yeah, `KeyInit` should not be used for public keys since it assumes that any randomly sampled key is a valid one. I am not sure that `KeyInit` is a great fit for private keys either. It may work fine for currently implemented KEMs since they use key as a seed, but in future we may have algorithms which are more similar to elliptic curves and reject some keys.

Maybe we should introduce separate traits for key pairs which would work with both ECC and KEMs? #1897 looks relevant as well.

tarcieri:
Yeah, that sounds good. See e.g. https://docs.rs/ml-kem/latest/ml_kem/trait.KemCore.html#tymethod.generate

I can impl `KeySizeUser` on public keys still I guess, but it's missing a trait to do the encoding to an `Array`. I guess I can define one in `kem` for that for now.

tarcieri:
> It may work fine for currently implemented KEMs since they use key as a seed, but in future we may have algorithms which are more similar to elliptic curves and reject some keys.

I think there's pretty widespread agreement to use seeds going forward, and if the underlying algorithm needs to reject certain keys, it can use deterministic rejection sampling ala RFC6979

tarcieri:
#2096 moves RNG to specific traits for interfacing with `getrandom` and `rand_core`, which would address this issue

tarcieri:
We now have `TryKeyInit` and `Generate`, so I'm calling this done

PR #2092, opened 2025-11-21
crypto-common: add `TryKeyInit` trait

### LINKED ISSUE #1897: crypto-common: fallible alternative to `KeyInit`

tarcieri:
Adds a trait providing fallible key initialization, similar to the existing `KeyInit` trait, but designed to handle the case that not all bytestrings of a given length represent valid keys.

This is primarily useful in the context of public-key cryptography, e.g. scalars representing elliptic curve private keys.

The API and method names are duplicated from `KeyInit`. It is assumed that `KeyInit` and `TryKeyInit` have an either-or relationship, i.e. types will not impl both `KeyInit` and `TryKeyInit`, and consumers of code which is generic over these traits will not be attempting to abstract over the `KeyInit`/`TryKeyInit` distinction, but one or the other will make sense in a given context (e.g. symmetric cryptography uses `KeyInit`, ECC uses `TryKeyInit`)

Key generation methods are provided which generically implement rejection sampling as a reasonabe baseline for unbiased random sampling of potential keys. As a provided method, implementers of `TryKeyInit` can override it if they can provide more efficient means of generating keys.

Closes #1897

### DIFF
--- a/crypto-common/src/lib.rs
+++ b/crypto-common/src/lib.rs
@@ -153,7 +153,7 @@ pub trait AlgorithmName {
     fn write_alg_name(f: &mut fmt::Formatter<'_>) -> fmt::Result;
 }
 
-/// Types which can be initialized from key.
+/// Types which can be initialized from a key.
 pub trait KeyInit: KeySizeUser + Sized {
     /// Create new value from fixed size key.
     fn new(key: &Key<Self>) -> Self;
@@ -209,7 +209,7 @@ pub trait KeyInit: KeySizeUser + Sized {
     }
 }
 
-/// Types which can be initialized from key and initialization vector (nonce).
+/// Types which can be initialized from a key and initialization vector (nonce).
 pub trait KeyIvInit: KeySizeUser + IvSizeUser + Sized {
     /// Create new value from fixed length key and nonce.
     fn new(key: &Key<Self>, iv: &Iv<Self>) -> Self;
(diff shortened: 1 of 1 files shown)


tarcieri:
> `KeyInit` and `TryKeyInit` have an either-or relationship, i.e. types will not impl both `KeyInit` and `TryKeyInit` [...] ECC uses TryKeyInit

There is a counterexample to this: the Ed25519 signature system uses a seed to derive the private scalar as the result of SHA-512 acting as a KDF, along with "clamping" to ensure the scalar is in range. So Ed25519 keys could use `KeyInit`.

But for ECDSA keys, or secp256k1 Schnorr, we'd want `TryKeyInit`. So perhaps having a way to abstract over the distinction should be considered.

(There is also a similar concern for X25519 static keys)

crypto-common/src/lib.rs
```diff
@@ -323,6 +323,73 @@ pub trait KeyIvInit: KeySizeUser + IvSizeUser + Sized {
...
+        // Use rejection sampling to find a key which initializes successfully in an unbiased manner
+        loop {
+            let mut key = Key::<Self>::default();
+            getrandom::fill(&mut key)?;
+
+            if Self::new(&key).is_ok() {
+                return Ok(key);
+            }
+        }
+    }
```
tarcieri, line 362: re: #2060 implementing random bytestring generation in `hybrid-array` (which I still think is an independently good idea) won't suffice for these use cases, since they need algorithm-specific logic to select valid keys.

The approach in this PR is able to provide a generic unbiased key generation implementation which can have algorithm-specific overrides to improve efficiency.

I think we could potentially drop the keygen for `KeyInit` and point to `hybrid-array` for such cases, but having generic keygen for `TryKeyInit` is still independently useful, IMO.

tarcieri:
With #2096 I think the RNG code in this PR can be substantially improved.

It would be nice to retain a generic implementation of rejection sampling though. Perhaps it could work with a blanket impl and a marker trait, e.g. `pub trait RejectionSamplingRng: TryKeyInit`

crypto-common/src/lib.rs
newpavlov: On the first glance, it certainly could be useful to have unbiased generation as part of the trait. Especially for algorithms which do not use all most significant bits in the key. Using naive rejection sampling can be really inefficient in such cases.

But looking at the current code, I can not say I am happy with it... Not only it forces us to keep the `rand_core`/`getrandom` features in `crypto-common`, but it also could be inefficient. For example, `let key = Alg::generate_key()?; let alg = Alg::new(&key).unwrap();` results in an unnecessary key validation as part of the `new` method (which is not guaranteed to be optimized out by the compiler), but also forces users to use `unwrap` or handle unreachable error.

How about introducing an associated type for "validated" key? Something like this:
```rust
pub trait TryKeyInit: KeySizeUser + Sized {
    type ValidatedKey: Into<Key<Self>> + TryFrom<Key<Self>, Error = InvalidKey>;
    fn new_from_validated(key: &Self::ValidatedKey) -> Self;
    fn new(key: &Key<Self>) -> Result<Self, InvalidKey> {
        key.try_into().map(Self::new_from_validated)
    }
}
```

This way we will be able to move key generation code to the validated key type without burdening the `TryKeyInit` trait. I haven't thought much about method names, so feel free to rename them.

tarcieri: >  let key = Alg::generate_key()?; let alg = Alg::new(&key).unwrap(); results in an unnecessary key validation as part of the new method 

This problem does not exist with the `Generate`/`FromCryptoRng` traits, it's inhereted by following the existing pattern of `KeyInit`, which is why I suggested separating the key generation traits out of `KeyInit` (#2060) and implemented an alternative which can generate random keys of any kind which you receive fully initialized (#2096), which fully addresses this concern.

tarcieri:
I'm just going to close this as it's very messy and I'm unhappy with it

crypto-common/src/lib.rs
newpavlov: Are you sure that we do not need the `TryKeyInit` with "validated" type? It would be orthogonal to #2096 and we even could add `Generate` bound on the associated type.

tarcieri: Except for micro-optimizing the case where keygen gives you back a "validated" bytestring rather than `Self`, I don't see a use case for it.

How else would you initialize a "validated" type other than fully initializing the `Self` and then throwing it away? And at that point, why not keep `Self` around, rather than the validated bytestring?

newpavlov: Validated byte string may be smaller than `Self`. You also could have alternative ways for generating it (e.g. using deterministic rejection sampling).

PR #2097, opened 2025-11-23
crypto-common: add `TryKeyInit` trait

tarcieri:
Take two for #2092 

Adds a trait providing fallible key initialization, similar to the existing `KeyInit` trait, but designed to handle the case that not all bytestrings of a given length represent valid keys.

This is primarily useful in the context of public-key cryptography, e.g. scalars representing elliptic curve private keys.

The API and method names are duplicated from `KeyInit`. It is assumed that `KeyInit` and `TryKeyInit` have an either-or relationship, i.e. types will not impl both `KeyInit` and `TryKeyInit`, and consumers of code which is generic over these traits will not be attempting to abstract over the `KeyInit`/`TryKeyInit` distinction, but one or the other will make sense in a given context (e.g. symmetric cryptography uses `KeyInit`, ECC uses `TryKeyInit`)

### DIFF
--- a/crypto-common/src/lib.rs
+++ b/crypto-common/src/lib.rs
@@ -153,7 +153,7 @@ pub trait AlgorithmName {
     fn write_alg_name(f: &mut fmt::Formatter<'_>) -> fmt::Result;
 }
 
-/// Types which can be initialized from key.
+/// Types which can be initialized from a key.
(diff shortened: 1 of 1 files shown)


tarcieri:
The `KeyInit` and `TryKeyInit` APIs are now both completely free of key generation functionality. I think such comments belong on the relevant key generation APIs instead.

PR #2098, opened 2025-11-24
aead: remove nonce generation APIs

tarcieri:
~~Depends on: #2096~~

Replaces them with the `Generate` trait (#2096)

Moves documentation about generating random nonces to the `Nonce` type.

### DIFF
--- a/crypto-common/src/lib.rs
+++ b/crypto-common/src/lib.rs
@@ -19,10 +19,10 @@ mod generate;
 pub use hybrid_array as array;
 pub use hybrid_array::typenum;
 
+#[cfg(feature = "getrandom")]
+pub use getrandom::Error as RngError;
 #[cfg(feature = "rand_core")]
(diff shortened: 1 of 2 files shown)


crypto-common/src/lib.rs
```diff
@@ -19,10 +19,10 @@ mod generate;
 pub use hybrid_array as array;
 pub use hybrid_array::typenum;
 
+#[cfg(feature = "getrandom")]
+pub use getrandom::Error as RngError;
```
newpavlov, line 23: Irrelevant change?

tarcieri: `aead` was the last thing that depended on the re-export of `getrandom` from `crypto-common`. This line is removing it, although `rustfmt` oddly makes that a little less obvious.

Edit: also note the re-export was recently introduced as part of the `rand_core` v0.10-rc* upgrade, so it's not worth noting in the changelog

PR #2106, opened 2025-12-05
password-hash: simplify `CustomizedPasswordHash::Params`

tarcieri:
Uses core traits to handle string serialization, rather than relying on types in `password_hash::phc`


tarcieri:
Need to actually run this change by https://github.com/RustCrypto/password-hashes but this otherwise decouples `password-hash` from `phc`, aside from its use as the return type, which can be converted to a generic parameter

tarcieri:
I have a local PR to RustCrypto/password-hashes working with this change, so I'm going to merge then push that up

PR #2110, opened 2025-12-07
password-hash: add generic `H` param to traits

tarcieri:
Adds a generic `H` parameter to the `(Customized)PasswordHasher` and `PasswordVerifier` traits which specifies the return type and the hash type respectively.

A generic parameter is used instead of an associated constant in order to support overlapping impls. Notably multiple commonly used password hashing algorithms (e.g. PBKDF2, scrypt) support encoding in either the Modular Crypt Format (MCF) or Password Hashing Competition (PHC) string format, which are implemented as the `mcf::PasswordHash` and `phc::PasswordHash` types respectively.

We already have open issues to add MCF support where we currently don't support it (RustCrypto/password-hashes#727,
RustCrypto/password-hashes#747).

The blanket impl of `PasswordVerifier` for `CustomizedPasswordHasher` is retained, but only for `phc::PasswordHash`, as supporting it for MCF requires algorithmic-specific interpretation of the hash. However, algorithms can provide their own impls of `PasswordVerifier<mcf::PasswordHash>`.

### DIFF
--- a/password-hash/src/lib.rs
+++ b/password-hash/src/lib.rs
@@ -49,16 +49,20 @@ use core::{
 pub type Version = u32;
 
 /// Trait for password hashing functions.
-pub trait PasswordHasher {
+///
+/// Generic around a password hash to be returned (typically [`PasswordHash`])
+pub trait PasswordHasher<H> {
     /// Simple API for computing a [`PasswordHash`] from a password and
     /// salt value.
     ///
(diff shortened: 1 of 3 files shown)


tarcieri:
Note: this almost completely decouples `password-hash` from `phc::PasswordHash`, to the point `phc` can be made an optional feature and potentially extracted into its own crate.

PR #2116, opened 2025-12-09
password-hash: use `phc` crate

tarcieri:
Replaces the `phc` submodule with a re-export of the `phc` crate, which was just extracted from the `password-hash` crate in RustCrypto/formats#2111.

A dependency on the crate is needed to make the blanket impl of `PasswordVerifier` for `PasswordHasher` work (notably the former is object safe, and a similar blanket impl can't be supported for MCF since algorithm-specific rules are required)

This also significantly simplifies the `Error` type now that it's only responsible for errors involving the traits, rather than errors from parsing password hashes.


password-hash/src/error.rs
```diff
@@ -0,0 +1,63 @@
...
+    },
+
+    /// Invalid parameters.
+    ParamsInvalid,
+
+    /// Invalid password.
+    PasswordInvalid,
+
+    /// Invalid salt.
+    SaltInvalid,
```
tarcieri, line 40: Note: several of these are significantly less fine-grained, notably you can't tell whether salts and outputs are too small or too big. Hopefully people can figure it out from context, though we may consider adding it back.

The `ParamInvalid` error should be a significant improvement in that it informs you exactly which parameter was invalid (although perhaps does a poor job of telling you why)

PR #2190, opened 2026-01-14
[WIP] `pake` crate

tarcieri:
Traits which describe the functionality of Password-Authenticated Key Exchange algorithms, which can be used to establish a shared secret over an insecure channel using only the client's knowledge of a password and the server's knowledge of a password hash for authentication.

### DIFF
--- a/.github/workflows/pake.yml
+++ b/.github/workflows/pake.yml
@@ -1,0 +1,55 @@
+name: pake
+
+on:
+  pull_request:
+    paths:
+      - "kdf/**"
+      - "pake/**"
+      - "Cargo.*"
+  push:
+    branches: master
+
+defaults:
+  run:
+    working-directory: pake
+
+env:
+  CARGO_INCREMENTAL: 0
+  RUSTFLAGS: "-Dwarnings"
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        rust:
+          - 1.85.0 # MSRV
+          - stable
+        target:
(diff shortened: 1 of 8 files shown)


rozbb:
Why is "proof" used here? Usually we just refer to message 1 and message 2 for a 2-message PAKE. Here's an example of 3 popular PAKEs
Screenshot 2026-01-23 at 11 55 39 AM
I wrote them as one round, but it doesn't matter implementation-wise, since usually one side is the designated initiator anyway.

tarcieri:
@rozbb happy to make the language more generic. The terminology there was coming from SRP which is what I was working on at the time (mostly I wrote this up trying to think of how to refactor the existing `srp` crate API into something more manageable)

rozbb:
Ok nice. Also, I assume you only want to include 2-message PAKEs for now? Those are the only standardized one atm.

I intend on implementing [CPace](https://www.ietf.org/archive/id/draft-irtf-cfrg-cpace-17.html) in Rust soon, so I will be able to have informed feedback on this.

tarcieri:
Yes, that sounds good. FWIW we have [`AuCPace`](https://github.com/RustCrypto/PAKEs/tree/master/aucpace)

makavity:
Hey, @tarcieri. Since we've released SWU function, my next plan is to do `bpace` protocol. Should I wait for the traits, or I can PR it right now?

tarcieri:
@makavity I'd say work on this is stalled for now, so don't let it be a blocker.

I intend to pick this up again after finishing stable releases of everything else.

PR #2191, opened 2026-01-14
CI: test docs build with `--cfg docsrs`

tarcieri:
Ensures our docs will actually build on docs.rs.

After publishing `kdf` I noticed since it's an ancient PR it still had `doc_auto_cfg`. This should reproduce the failure.

### DIFF
--- a/kdf/src/lib.rs
+++ b/kdf/src/lib.rs
@@ -1,5 +1,5 @@
 #![no_std]
-#![cfg_attr(docsrs, feature(doc_auto_cfg))]
+#![cfg_attr(docsrs, feature(doc_cfg))]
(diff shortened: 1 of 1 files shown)


tarcieri:
Alright, failure reproduced:

```
error[E0557]: feature has been removed
 --> kdf/src/lib.rs:2:29
  |
2 | #![cfg_attr(docsrs, feature(doc_auto_cfg))]
  |                             ^^^^^^^^^^^^ feature has been removed
  |
  = note: removed in 1.92.0; see <https://github.com/rust-lang/rust/pull/138907> for more information
  = note: merged into `doc_cfg`
```

Now let me push the fix.

PR #2195, opened 2026-01-19
Bump `rand_core` to v0.10.0-rc-4

tarcieri:
It seems like the changes are backwards compatible with every crate except for `elliptic-curve`, where some reborrowing was required

### DIFF
--- a/elliptic-curve/src/scalar/blinded.rs
+++ b/elliptic-curve/src/scalar/blinded.rs
@@ -38,10 +38,10 @@ where
     C: CurveArithmetic,
 {
     /// Create a new [`BlindedScalar`] from a scalar and a [`CryptoRng`].
-    pub fn new<R: CryptoRng + ?Sized>(scalar: Scalar<C>, rng: &mut R) -> Self {
+    pub fn new<R: CryptoRng + ?Sized>(scalar: Scalar<C>, mut rng: &mut R) -> Self {
         Self {
             scalar,
-            mask: Scalar::<C>::random(rng),
+            mask: Scalar::<C>::random(&mut rng),
         }
     }
 }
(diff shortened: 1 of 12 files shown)


tarcieri:
FYI @baloo @dhardy

baloo:
huum, no there is something fishy here, you shouldn't have to do that.

baloo:
I think this is because `CryptoRng` no longer implies `RngCore` and that trait now has to be explicit.

I've opened https://github.com/RustCrypto/traits/pull/2197 to remove the extraneous references

dhardy:
> I think this is because `CryptoRng` no longer implies `RngCore` and that trait now has to be explicit.

It very much should do. If it doesn't that's a bug. (I know the traits are written differently, but as best I can tell this is still implied.)

elliptic-curve/src/scalar/blinded.rs
```diff
@@ -38,10 +38,10 @@ where
     C: CurveArithmetic,
 {
     /// Create a new [`BlindedScalar`] from a scalar and a [`CryptoRng`].
-    pub fn new<R: CryptoRng + ?Sized>(scalar: Scalar<C>, rng: &mut R) -> Self {
+    pub fn new<R: CryptoRng + ?Sized>(scalar: Scalar<C>, mut rng: &mut R) -> Self {
         Self {
             scalar,
-            mask: Scalar::<C>::random(rng),
+            mask: Scalar::<C>::random(&mut rng),
```
newpavlov, line 44: Why not keep `Scalar::<C>::random(rng)`?

baloo: Got taken care of in https://github.com/RustCrypto/traits/pull/2197

PR #2197, opened 2026-01-19
elliptic-curve: fixup rand_core API use

baloo:
This fixes changes made in https://github.com/RustCrypto/traits/pull/2195

This is due to the refactor made in rand_core in https://github.com/rust-random/rand_core/pull/45 which dropped the "trait dependency" between CryptoRng and RngCore

### DIFF
--- a/elliptic-curve/src/point/non_identity.rs
+++ b/elliptic-curve/src/point/non_identity.rs
@@ -5,7 +5,7 @@
 use core::ops::{Deref, Mul};
 
 use group::{Group, GroupEncoding, prime::PrimeCurveAffine};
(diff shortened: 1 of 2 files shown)


newpavlov:
Hm... I expected that `CryptoRng` should work just fine since it implements `TryCryptoRng<Error=!>` and thus `TryRngCore<Error=!>` and it in turn should mean that it implements `RngCore`, but I guess the compiler is not able to see through the double indirection and it's worth to change the definition to `CryptoRng: TryRngCore<Error=!> + RngCore`.

tarcieri:
Yeah that is weird, I don't get why the blanket impl doesn't suffice

PR #2198, opened 2026-01-19
workspace: avoid duplicate crypto-common definition when using git dependencies

baloo:
When consuming an master release with `[patch.crates-io]` crypto-common would get duplicated because of the `path = "../` in each crate.

This moves the override to the workspace `Cargo.toml` and remove that duplicate crypto-common crate instance.

### DIFF
--- a/aead/Cargo.toml
+++ b/aead/Cargo.toml
@@ -16,7 +16,7 @@ such as AES-GCM as ChaCha20Poly1305, which provide a high-level API
 """
 
 [dependencies]
-crypto-common = { version = "0.2.0-rc.10", path = "../crypto-common" }
+crypto-common = "0.2.0-rc.10"
 inout = "0.2.2"
 
 # optional dependencies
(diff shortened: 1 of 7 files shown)


aead/Cargo.toml
```diff
@@ -16,7 +16,7 @@ such as AES-GCM as ChaCha20Poly1305, which provide a high-level API
 """
 
 [dependencies]
-crypto-common = { version = "0.2.0-rc.10", path = "../crypto-common" }
+crypto-common = { version = "0.2.0-rc.10" }
```
newpavlov: It's worth to change it to just `crypto-common = "0.2.0-rc.10"` here and in the other crates.

tarcieri:
Agree with @newpavlov's suggestion

baloo:
That's fixed right?

tarcieri (approved):

tarcieri:
Yep, looks good

PR #2205, opened 2026-01-20
Bump `rand_core` to v0.10.0-rc-5

tarcieri:


elliptic-curve/src/point/non_identity.rs
```diff
@@ -88,7 +88,7 @@ where
     P: ConditionallySelectable + ConstantTimeEq + CurveGroup + Default,
 {
     /// Generate a random `NonIdentity<ProjectivePoint>`.
-    pub fn random<R: CryptoRng + RngCore + ?Sized>(rng: &mut R) -> Self {
+    pub fn random<R: CryptoRng + ?Sized>(rng: &mut R) -> Self {
```
tarcieri, line 91: Confirm the new bounds work /cc @baloo

PR #2213, opened 2026-01-21
crypto-common: add `KeyExport` trait

tarcieri:
Adds a companion trait to `KeyInit*` for exporting a key to a byte array sized according to `KeySizeUser`.

This is needed when generating e.g. asymmetric secret keys types using the `Generate` trait and it returns an actual key type (e.g. `SigningKey`, `DecryptionKey`, `DecapsulationKey`) as opposed to its serialized byte representation.

This PR also shows another use case for it: there's currently a gap in the `kem` crate around serialization where you can generically go from `Decapsulate::encapsulator` to the associated encapsulation key for a KEM, but there is currently no generic way to serialize it.

It adds `TryKeyInit` and `KeyExport` bounds to `kem::Encapsulate`, because as the public key component we should always be able to serialize/deserialize it.

(NOTE: happy to split the `kem` changes out, that would probably be for the best, but I thought it would help illustrate the use case)

### DIFF
--- a/crypto-common/CHANGELOG.md
+++ b/crypto-common/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Sealed `BlockSizes` trait implemented for types from `U1` to `U255`
 - `Generate` trait ([#2096])
+- `KeyExport` trait ([#2213])
 
 ### Changed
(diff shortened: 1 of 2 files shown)


tarcieri:
Removed the `kem` changes and added a CHANGELOG entry in [419d83c](https://github.com/RustCrypto/traits/pull/2213/commits/419d83c1ad14e0ac0016a290546dd9b5bd87431c)

Issue #2214, opened 2026-01-22
Annoying to impl `kem::Encapsulate`


rozbb:
The current definition of `kem::Encapsulate` is

```rust
pub trait Encapsulate<EK, SS> {
    /// Encapsulation error
    type Error: core::error::Error;

    /// Encapsulates a fresh shared secret
    fn encapsulate_with_rng<R: TryCryptoRng + ?Sized>(
        &self,
        rng: &mut R,
    ) -> Result<(EK, SS), Self::Error>;

    // ...
}
```
Note that `Error` is an associated type, but `encapsulate_with_rng` takes a fallible RNG, **whose error type can be anything** (as long as it's `Debug + Display`). So an implementer who wants to faithfully return an RNG error needs to either 1) define `Encapsulate::Error` to store a `Box<dyn Debug + Display>` in case the RNG fails, or 2) not store the RNG error at all, and just return a unit type `RngError`. Both of these options aren't great. (1) is particularly odd because usually you only care about fallible RNGs in embedded settings, where you might not have an allocator anyway.

One solution I like is to remove `TryCryptoRng` from the `kem` API entirely, and just accept `CryptoRng`.

I'll strengthen it further, actually: I am not convinced that ANY low-level cryptography crate should accept a `TryCryptoRng` in its API. imo there is no reason that an IO error should show up in a public key encryption, or a polynomial secret sharing operation. I think it only serves to clutter the API, and frustrate error handling (like we see above). If a user has a fallible RNG and wants to use a low-level cryptography library, I think the correct path is to use that RNG to make a 32 byte seed, and then instantiate an infallible seedable RNG with it.

tarcieri:
This might help: #2216

Part of what makes handling the fallible RNG annoying currently is you have to return the associated `Error` type, but the `encapsulate_with_rng` method has its own RNG, so you need to define an error type just to map to your own custom error type you can associate and return just to handle RNG failures.

#2216 makes `encapsulate_with_rng` return `R::Error`, so you should just be able to use `try_fill_bytes` or what have you along with `?` to propagate the error.

tarcieri:
I think I'd be okay changing `encapsulate_with_rng` to take a `CryptoRng`.

If people want fallibility, perhaps later we could add a `TryEncapsulate` trait which takes a `TryCryptoRng` (though we run into the problem of what to do for converting `R::Error` into `Self::Error` again)

tarcieri:
[...]

PR #2216, opened 2026-01-22
kem: remove associated `Error` types

tarcieri:
All of our `kem` implementations either use `type Error = Infallible` or use the error type exclusively for handling RNG errors.

That's good, because having an error case for decapsulation introduces a potential sidechannel, which can be eliminated by instead using implicit rejection that returns a pseudorandom rejection symbol as its output.

This removes the `Error` types and makes `Decapsulate::decapsulate` infallible in order to close the potential sidechannel having fallible decapsulation provides.

`Encapsulate::encapsulate_with_rng` now only uses the `Result` for handling RNG errors and returns `R::Error`, which should hopefully help mitigate the concerns in #2214.

For end users, `Encapsulate::encapsulate` now provides infallible encapsulation using the system RNG.

### DIFF
--- a/kem/src/lib.rs
+++ b/kem/src/lib.rs
@@ -12,24 +12,24 @@ pub use crypto_common::{Generate, KeyExport, KeySizeUser, TryKeyInit, typenum::c
 
 use rand_core::TryCryptoRng;
 
+#[cfg(feature = "getrandom")]
+use {crypto_common::getrandom::SysRng, rand_core::TryRngCore};
+
 /// Encapsulator for shared secrets.
 ///
 /// Often, this will just be a public key. However, it can also be a bundle of public keys, or it
 /// can include a sender's private key for authenticated encapsulation.
 pub trait Encapsulate<EK, SS>: TryKeyInit + KeyExport {
-    /// Encapsulation error
-    type Error: core::error::Error;
-
     /// Encapsulates a fresh shared secret
-    fn encapsulate_with_rng<R: TryCryptoRng + ?Sized>(
-        &self,
-        rng: &mut R,
-    ) -> Result<(EK, SS), Self::Error>;
+    fn encapsulate_with_rng<R>(&self, rng: &mut R) -> Result<(EK, SS), R::Error>
+    where
+        R: TryCryptoRng + ?Sized;
 
     /// Encapsulate a fresh shared secret generated using the system's secure RNG.
     #[cfg(feature = "getrandom")]
-    fn encapsulate(&self) -> Result<(EK, SS), Self::Error> {
-        self.encapsulate_with_rng(&mut crypto_common::getrandom::SysRng)
+    fn encapsulate(&self) -> (EK, SS) {
+        let Ok(ret) = self.encapsulate_with_rng(&mut SysRng.unwrap_err());
+        ret
     }
 }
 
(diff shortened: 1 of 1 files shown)


rozbb:
This seems reasonable to me. Fwiw, implicit rejection [now seems](https://eprint.iacr.org/2023/1811) to be more of a byproduct of analysis rather than a real security mitigation. Future KEMs may explicitly reject during decapsulation. But we can cross that bridge when we get there.

Do you have opinions on removing `TryCryptoRng` entirely? I know it's a hot take so I wanted to see what your thoughts were. imo its upsides in maintainability win against the downsides of strictness

tarcieri:
> Fwiw, implicit rejection [now seems](https://eprint.iacr.org/2023/1811) to be more of a byproduct of analysis rather than a real security mitigation. Future KEMs may explicitly reject during decapsulation. But we can cross that bridge when we get there.

I'm curious how you avoid adaptive chosen ciphertext attacks if there is a timing sidechannel around rejection. It doesn't seem like that's the sort of thing that paper is concerning itself with though, skimming it the conclusions seem to be mostly in regard to the security proof.

I imagine with future KEMs there's going to be some kind of risk/benefit analysis in regard to implementing explicit rejection which will involve quite a bit of debate, so I agree we can cross that bridge when we get there.

> Do you have opinions on removing TryCryptoRng entirely? I know it's a hot take so I wanted to see what your thoughts were. imo its upsides in maintainability win against the downsides of strictness

We've generally tried to support both `CryptoRng` and `TryCryptoRng` elsewhere, along with `getrandom`-based APIs.

Currently this supports an infallible `getrandom`-based API and `TryCryptoRng`, but not a fallible `getrandom`-based API or `CryptoRng`, so as-is it's somewhat inconsistent with other things.

I'm going to go ahead and merge this and maybe we can circle back on `CryptoRng`, but this should at least make the `TryCryptoRng` error handling easier.

rozbb:
> I'm going to go ahead and merge this and maybe we can circle back on CryptoRng, but this should at least make the TryCryptoRng error handling easier.

Sounds good!

> I'm curious how you avoid adaptive chosen ciphertext attacks if there is a timing sidechannel around rejection.

There's no side-channel. Say a KEM does Fujisaki-Okamoto with rejection, ie, it checks that the ciphertext is well-formed via re-computation, and rejects if not. In this setting, decapsulation fails if and only if the encapsulator deviated from the encapsulation algorithm. The encapsulator already knows they deviated, so that's not an interesting bit of information. What is leaky here?

tarcieri:
Rejecting a malformed ciphertext seems reasonably OK, but perhaps that could be a function of `EK` which happens in advance?

Sidebar: I guess the one remaining gap in the API right now is serializing/deserializing `EK`

rozbb:
> Rejecting a malformed ciphertext seems reasonably OK, but perhaps that could be a function of `EK` which happens in advance?

You can only see if they deviated from the encapsulation algorithm if you have the decapsulation key. So this has to happen after deserialization.

> I guess the one remaining gap in the API right now is serializing/deserializing EK

Yeah, it was tough to get something that encompassed all the aspects of a KEM and also made sense and also wasn't unwieldy. So far I haven't needed (de)serialization routines as a trait, so I don't have intuition for what should be covered.

tarcieri:
@rozbb I see. Well I guess the good news is `Decapuslate` doesn't need to deal with RNG errors, so having an associated `Error` type for that wouldn't be the end of the world

tarcieri:
We can always add a `TryDecapsulate` trait to handle the fallible case.

PR #2220, opened 2026-01-22
kem: use byte arrays for `EK` and `SS`

### LINKED ISSUE #2219: kem: `EK`/`SS` serialization
The one remaining gap in the `kem` API is a way to obtain the bytes serialization of `EK` and `SS` (encapsulated key and shared secret respectively), and for `EK`, a way to deserialize it from bytes before passing it to `Decapsulate::decapsulate`.

I was first wondering if there's actually a use case for overlapping impls with these traits and if not, perhaps these should be associated types instead? That would make it easier to give them bounds. Something like:

```rust
pub trait Encapsulate: TryKeyInit + KeyExport {
    /// Shared secret after being encrypted by this encapsulator
    type EncapsulatedKey: TryKeyInit + KeyExport;

    /// Shared secret value which results after the `EncapsulatedKey` has been decrypted.
    type SharedSecret: AsRef<[u8]>;

    fn encapsulate_with_rng<R>(&self, rng: &mut R) -> Result<(EK, SS), R::Error>
    where
        R: TryCryptoRng + ?Sized;

    /// Encapsulate a fresh shared secret generated using the system's secure RNG.
    #[cfg(feature = "getrandom")]
    fn encapsulate(&self) -> (EK, SS) {
        let Ok(ret) = self.encapsulate_with_rng(&mut SysRng.unwrap_err());
        ret
    }
}

pub trait Decapsulate {
    /// Encapsulator which corresponds to this decapsulator.
    type Encapsulator: Encapsulate<
        EncapsulatedKey = Self::EncapsulatedKey,
        SharedSecret = Self::SharedSecret
    >;

    /// Shared secret after being encrypted by this encapsulator
    type EncapsulatedKey: TryKeyInit + KeyExport;

    /// Shared secret value which results after the `EncapsulatedKey` has been decrypted.
    type SharedSecret: AsRef<[u8]>;

    /// Decapsulates the given encapsulated key
    fn decapsulate(&self, encapsulated_key: &EK) -> SS;

    /// Retrieve the encapsulator associated with this decapsulator.
    fn encapsulator(&self) -> Self::Encapsulator;
}
```

We could also consider a third trait that removes the duplicated associated types, perhaps (possibly `Kem`?)

tarcieri:
Some terminology:
- `EK`: encapsulated key, i.e. ciphertext, not to be confused with "encapsulation key", the public key
- `SS`: shared secret, output of the decapsulator when given a ciphertext a.k.a. encapsulated key

`EK` and `SK` were previously generic parameters on the `Encapsulate` and `Decapsulate` traits, however KEMs don't benefit from overlapping impls and have relatively fixed notions of what these types should be.

This commit replaces them with new type aliases `Ciphertext` and `SharedSecret`:

- `Ciphertext<K>`: type alias for `Array<u8, K::CiphertextSize>`, a.k.a. "encapsulated key", where `Array` is from `hybrid-array`.
- `SharedSecret<K>`: type alias for `Array<u8, K::SharedSecretSize>`

This means consumers of the traits always use bytestrings, which should hopefully make it dramatically simpler to implement things generically across KEMs.

The `K` generic parameter above is for types which impl a new `Kem` trait which defines two associated `ArraySize`s:
- `Kem::CiphertextSize`: size of the ciphertext
- `Kem::SharedSecretSize`: size of the shared secret

This was split out into its own trait so the `Ciphertext<K>` and `SharedSecret<K>` type aliases work with either encapsulators or decapsulators.

Next, `Decapsulate` was split into three(!) traits to handle fallible decapsulation:

- `Decapsulate`: what we had before with the `Decapsulate::Encapsulator` associated type extracted into a supertrait `Decapsulator`, which it now bounds on. It has a provided `decapsulate_slice` method which returns `core::array::TryFromSliceError` in the event the provided slice does not match `CiphertextSize`
- `TryDecapsulate`: fallible equivalent of `Decapsulate`, kind of like what we had prior to #2216, with an associated `Error` type and with a `try_decapsulate` method that returns a result. It also bounds on `Decapsulator` as its supertrait.
- `Decapsulator`: common supertrait of `Decapsulate` and `TryDecapsulate` which defines the associated `Encapsulator` and provides the `Decapsulator::encapsulator` method for retrieving it.

A blanket impl of `TryDecapsulate` is provided for types which impl `Decapsulate` which uses `Infallible` as the error type, so any type which impls `Decapsulate` can be used as a `TryDecapsulate`-bounded argument. Likewise `Decapsulate` carries a `TryDecapsulate<Error = Infallible>` bound which is satisfied by the blanket impl but also enforces this property.

The reason we need to reintroduce fallible decapsulation is `dhkem`: when I was scanning over our KEMs repo looking at the error types, `dhkem` has `Error = Infallible`, but this hides that it was using `elliptic_curve::PublicKey<C>` as its "encapsulated key" / ciphertext type, which is a well-typed wrapper for a valid curve point.

With this type now being a raw byte slice, `dhkem` needs to handle decoding the curve point in the `TryDecapsulate` impl, and if the point fails to decode return an error (there are ways we could pseudorandomly select a different point in constant time to compute a rejection symbol, but having it return an error in this case seems like a straightforward way to start).

Closes #2219

### DIFF
--- a/kem/src/lib.rs
+++ b/kem/src/lib.rs
@@ -22,20 +22,20 @@ use {common::getrandom, rand_core::TryRngCore};
 /// Ciphertext message (a.k.a. "encapsulated key") produced by [`Encapsulate::encapsulate`] which is
 /// an encrypted [`SharedSecret`] that can be decrypted using [`Decapsulate::decapsulate`].
 ///
-/// `K` is expected to be a type that impls [`Kem`], such as an encapsulator or decapsulator.
-pub type Ciphertext<K> = array::Array<u8, <K as Kem>::CiphertextSize>;
+/// `K` is expected to be a type that impls [`KemParams`], such as an encapsulator or decapsulator.
+pub type Ciphertext<K> = array::Array<u8, <K as KemParams>::CiphertextSize>;
 
 /// Shared secret: plaintext produced after decapsulation by [`Decapsulate::decapsulate`] which is
 /// also returned by [`Encapsulate::encapsulate`].
 ///
-/// `K` is expected to be a type that impls [`Kem`], such as an encapsulator or decapsulator.
-pub type SharedSecret<K> = array::Array<u8, <K as Kem>::SharedSecretSize>;
+/// `K` is expected to be a type that impls [`KemParams`], such as an encapsulator or decapsulator.
+pub type SharedSecret<K> = array::Array<u8, <K as KemParams>::SharedSecretSize>;
 
-/// Key encapsulation mechanism.
+/// Key encapsulation mechanism parameters: sizes of the ciphertext and decrypted plaintext.
 ///
 /// This trait is impl'd by types that impl either [`Encapsulate`] or [`Decapsulate`] and defines
 /// the sizes of the encapsulated key and shared secret.
-pub trait Kem {
+pub trait KemParams {
     /// Size of the ciphertext (a.k.a. "encapsulated key") produced by [`Encapsulate::encapsulate`].
     type CiphertextSize: ArraySize;
 
@@ -47,7 +47,7 @@ pub trait Kem {
 ///
 /// Often, this will just be a public key. However, it can also be a bundle of public keys, or it
 /// can include a sender's private key for authenticated encapsulation.
-pub trait Encapsulate: Kem + TryKeyInit + KeyExport {
+pub trait Encapsulate: KemParams + TryKeyInit + KeyExport {
     /// Encapsulates a fresh [`SharedSecret`] generated using the supplied random number
     /// generator `R`.
     fn encapsulate_with_rng<R: TryCryptoRng + ?Sized>(
@@ -67,21 +67,21 @@ pub trait Encapsulate: Kem + TryKeyInit + KeyExport {
 /// Trait for decapsulators, which is a supertrait bound of both [`Decapsulate`] and
 /// [`TryDecapsulate`].
 pub trait Decapsulator:
-    Kem<
-        CiphertextSize = <Self::Encapsulator as Kem>::CiphertextSize,
-        SharedSecretSize = <Self::Encapsulator as Kem>::SharedSecretSize,
+    KemParams<
+        CiphertextSize = <Self::Encapsulator as KemParams>::CiphertextSize,
+        SharedSecretSize = <Self::Encapsulator as KemParams>::SharedSecretSize,
     >
 {
     /// Encapsulator which corresponds to this decapsulator.
-    type Encapsulator: Encapsulate + Clone + Kem;
+    type Encapsulator: Encapsulate + Clone + KemParams;
 
     /// Retrieve the encapsulator associated with this decapsulator.
     fn encapsulator(&self) -> &Self::Encapsulator;
 }
 
-impl<K: Decapsulator> Kem for K {
-    type CiphertextSize = <K::Encapsulator as Kem>::CiphertextSize;
-    type SharedSecretSize = <K::Encapsulator as Kem>::SharedSecretSize;
+impl<K: Decapsulator> KemParams for K {
+    type CiphertextSize = <K::Encapsulator as KemParams>::CiphertextSize;
+    type SharedSecretSize = <K::Encapsulator as KemParams>::SharedSecretSize;
 }
 
 /// Decapsulator for encapsulated keys, with an associated `Encapsulator` bounded by the


kem/src/lib.rs
```diff
@@ -8,45 +8,125 @@
...
+    where
+        Self::Error: From<TryFromSliceError>,
+    {
+        self.try_decapsulate(ct.try_into()?)
+    }
+}
+
+impl<D> TryDecapsulate for D
+where
+    D: Decapsulate,
```
rozbb, line 135: Decapsulate already implies TryDecapsulate with Error = Infallible, no? It’s in the trait def

kem/src/lib.rs
```diff
@@ -8,45 +8,125 @@
...
+/// It could also be a hardware device like an HSM, TPM, or SEP.
 ///
 /// When possible (i.e. for software / non-HSM implementations) types which impl this trait should
 /// also impl the [`Generate`] trait to support key generation.
-pub trait Decapsulate<EK, SS> {
-    /// Encapsulator which corresponds to this decapsulator.
-    type Encapsulator: Encapsulate<EK, SS>;
+pub trait Decapsulate: Decapsulator + TryDecapsulate<Error = Infallible> {
+    /// Decapsulates the given [`Ciphertext`] a.k.a. "encapsulated key".
+    fn decapsulate(&self, ct: &Ciphertext<Self>) -> SharedSecret<Self>;
```
rozbb, line 98: Why not define this to do the try_decapsulate then unwrap?

kem/src/lib.rs
```diff
@@ -8,45 +8,125 @@
...
-/// Decapsulator for an encapsulated keys, with an associated encapsulator.
+/// Trait for decapsulators, which is a supertrait bound of both [`Decapsulate`] and
+/// [`TryDecapsulate`].
+pub trait Decapsulator: Kem {
+    /// Encapsulator which corresponds to this decapsulator.
+    type Encapsulator: Encapsulate
+        + Kem<CiphertextSize = Self::CiphertextSize, SharedSecretSize = Self::SharedSecretSize>;
+
+    /// Retrieve the encapsulator associated with this decapsulator.
+    fn encapsulator(&self) -> Self::Encapsulator;
```
rozbb: Why not return a reference? Pubkeys can get pretty big. If you already have it in the decapsulator, why copy it?

kem/src/lib.rs
tarcieri: This is a blanket impl that always satisfies that bound. It means that any type that impls `Decapsulate` also impls `TryDecapsulate` with `Error = Infallible` automatically.

The explicit bound means we always know `D` can be used as `TryDecapsulate` (we ran into a problem where a blanket impl like this wasn't working in rust-random/rand_core#51 and it was really hard to debug, so this reduces the problem to why `TryDecapsulate` isn't impl'd)

kem/src/lib.rs
tarcieri: Because that would introduce a panic.

The idea is that `Decapsulate` can be used as a `TryDecapsulate` if you want, but not the other way around (unless `Error = Infallible`). So if you have a mixture, `TryDecapsulate` becomes the common abstraction.

Think about it like `From`/`TryFrom`. `TryFrom` has a blanket impl for types which impl `From` where `Error = Infallible`. But you wouldn't want the other way around, for `From` impls to unwrap `TryFrom` meaning using `From` might cause panics (this is giving me `generic-array` flashbacks)

kem/src/lib.rs
tarcieri: Yeah, it probably makes sense to borrow and require decapsulators always store a copy of their encapsulation key. If we do make it borrow, we should probably add a `Clone` bound too.

FWIW in `signature` we have both [`Keypair`](https://docs.rs/signature/latest/signature/trait.Keypair.html) and [`KeypairRef`](https://docs.rs/signature/latest/signature/trait.KeypairRef.html) for this, where the former is blanket impl'd for types that impl the latter.

tarcieri: Changed it to borrow with a mandatory `Clone` bound in [54b5144](https://github.com/RustCrypto/traits/pull/2220/commits/54b5144e6ec497b95b131de68609b8172e8abb46)

tarcieri:
Alright, I have a companion PR that implements this passing: https://github.com/RustCrypto/KEMs/pull/191

A few final thoughts:
- I want to rename what this PR calls `Kem` to `KemParams` so we can perhaps save `Kem` for something that describes the whole type family. Several of the bespoke traits in https://github.com/rustcrypto/kems worked like that, and it reminds me of how `elliptic-curve` works
- I added a blanket impl of `Kem(Params)` for all `Decapsulator`s so they effectively inherit their params from the `Encapsulator` but it still seems a little weird

Other than that I plan to merge this, and then attempt to rip out the bespoke traits from https://github.com/RustCrypto/KEMs so we only have one set of traits (the `kem` crate traits) to work with.

Happy for suggestions to evolve this design though.

tarcieri:
I renamed it to `KemParams`, leaving `Kem` free for future use.

Merging

rozbb:
`KemParams` is a good name. Agree on all points 👍

PR #2237, opened 2026-01-25
Re-export `crypto_common` as `common`

tarcieri:
Or rather, this imports `crypto-common` as `common` using the `package = crypto-common` syntax in Cargo.toml, then re-exports it, which should hopefully make for better diagnostics.

This is shorter and more convenient for imports.

Where crates were previously exporting `crypto_common`, this adds an alias with a deprecation. Adding such an alias to `digest` was necessary to even get anything to compile.

### DIFF
--- a/aead/CHANGELOG.md
+++ b/aead/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Added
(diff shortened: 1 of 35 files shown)


aead/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Added
+- Re-export `crypto_common` as `common` ([#2237])
```
tarcieri, line 10: Note: `aead` didn't previously re-export `crypto-common` itself, only select things from it

PR #2240, opened 2026-01-25
crypto-common: add `UnwrapSysRng` alias

tarcieri:
Though I think the `Generate` trait covers a lot of use cases, for anyone who wants to use `SysRng` with `CryptoRng` this is a little more convenient than importing both `SysRng` and `Unwrap` and assembling it yourself.

### DIFF
--- a/crypto-common/src/lib.rs
+++ b/crypto-common/src/lib.rs
@@ -20,7 +20,7 @@ pub use hybrid_array as array;
 pub use hybrid_array::typenum;
 
 #[cfg(feature = "getrandom")]
-pub use getrandom;
+pub use getrandom::{self, SysRng};
 #[cfg(feature = "rand_core")]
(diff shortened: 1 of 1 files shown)


newpavlov:
I don't think we should implement it ourselves. I tried to add it to `getrandom`, but it was rejected, see: https://github.com/rust-random/getrandom/pull/754

Also note that creating value of the aliased type would be less convenient than with `SysRng`, i.e. you would have to write `UnwrapSysRng::default()`.

cc @dhardy @josephlr

tarcieri:
Oh hey look at that you proposed the same thing. Yeah, the main thing that's nice is it cuts the two imports down to one.

tarcieri:
I'll close this for now and if it gets upstream it gets upstream, I guess

PR #2244, opened 2026-01-30
crypto-common: remove weak key test methods

newpavlov:
As was discussed [here](https://github.com/RustCrypto/block-ciphers/pull/526#issuecomment-3793040411).


newpavlov:
Note that changelog entry is not needed since the items weren't part of v0.1 releases.

PR #2245, opened 2026-01-30
cipher: remove padded methods from BlockCipherEncrypt/Decrypt traits

### LINKED ISSUE #2129: cipher: simplify `BlockCipherEncrypt/Decrypt` traits?
The traits contain a bunch of methods which exist only for the ECB mode, which is "implemented" directly by cipher types.

This arrangement is hard to discover for new users and increases API surface of our traits significantly for a niche (and heavily discouraged) use case. It may be worth to remove the methods in favor of implementing ECB mode in a separate crate. I think we should be able to implement `BlockCipherMode` for `&ecb::Encrypt<C>` to preserve the ability to use the mode with shared references.

I asked the `ecb` crate owner about potential transfer [here]( https://github.com/magic-akari/ecb/issues/15).

newpavlov:
The methods were intended to be used for the ECB mode, but they are hard to discover in practice for inexperienced users, so I plan to introduce a separate crate for the mode.

Closes #2129


newpavlov:
@tarcieri 
WDYT about the remaining methods? While they are convenient and probably expected by users, they are problematic since relying on them results in potential backend branching on each method call, which can be quite sub-optimal. In particular, a number of AEAD crates suffer from that.

I wonder if we should keep only `encrypt/decrypt_with_backend` methods to force users to properly handle block cipher implementations with multiple backends selected at runtime.

tarcieri:
I'll say it's a little strange our block cipher traits don't provide the API typically provided by block ciphers whatsoever, but only somewhat arcane performance-oriented APIs

newpavlov:
Ok, let's keep them. But I think I will add a warning to the method docs about the potential inefficiency around backend selection at runtime.

bryceberger:
The changelog says to use the crate `ecb-mode` as a replacement, but I can't find it. It's not on crates.io, and I don't see it after a cursory search of RustCrypto repositories.

Was the crate ever made?

newpavlov:
There is the [`ecb`](https://docs.rs/ecb) crate, compatible with `cipher` v0.5. Note that currently it's not part of RustCrypto, I [asked](https://github.com/magic-akari/ecb/issues/15) @magic-akari about potential transfer, but got no reply. I am not sure whether it's worth to publish `ecb-mode` as a competing crate and to recommend a third-party crate.

newpavlov:
I replaced `ecb-mode` with `ecb` in the changelog in #2488.

PR #2256, opened 2026-02-02
kdf v0.1.0

tarcieri:
Initial release

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -357,7 +357,7 @@ checksum = "92ecc6618181def0457392ccd0ee51198e065e016d1d527a7ac1b6dc7c1f09d2"
 
 [[package]]
 name = "kdf"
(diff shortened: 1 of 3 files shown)


tarcieri:
This is wired up to `hkdf` and all of the PBKDFs in https://github.com/RustCrypto/password-hashes

It hasn't been implemented for anything other than `hkdf` in https://github.com/rustcrypto/kdfs because we'll need to define per-algorithm ZSTs to do the trait impl on.

There isn't much to it but it's useful functionality, particularly for PAKEs where there isn't otherwise a good trait to hook into the low-level KDF interface.

If there's one thing left to discuss it's probably the `Error` type: the crate defines its own unit struct, but some other options are an associated `Error` type, or an `enum Error` with some variants for common KDF errors.

Edit: let's try the simple error type and see if anyone requests something different.

PR #2258, opened 2026-02-02
Bump `getrandom` to v0.4

tarcieri:
Release PR: rust-random/getrandom#798

### DIFF
--- a/crypto-common/Cargo.toml
+++ b/crypto-common/Cargo.toml
@@ -16,7 +16,7 @@ description = "Common cryptographic traits"
 hybrid-array = "0.4"
 
 # optional dependencies
-getrandom = { version = "0.4.0-rc.1", optional = true, features = ["sys_rng"] }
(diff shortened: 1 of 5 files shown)


crypto-common/Cargo.toml
```diff
@@ -16,7 +16,7 @@ description = "Common cryptographic traits"
 hybrid-array = "0.4"
 
 # optional dependencies
-getrandom = { version = "0.4.0-rc.1", optional = true, features = ["sys_rng"] }
+getrandom = { version = "0.4", optional = true, features = ["sys_rng"] }
```
newpavlov, line 19: Maybe we should gate `sys_rng` on `rand_core`?

I don't mean it should be done as part of this PR, just noticed this while reviewing this PR.

tarcieri: It is:

```toml
[features]
getrandom = ["rand_core", "dep:getrandom"]
```

The only way you can use `getrandom` is `rand_core`+`sys_rng`

PR #2273, opened 2026-02-03
Incorporate README.md into rustdoc where it wasn't already

tarcieri:
The `cipher`, `crypto-common`, and `digest` crates didn't yet do this

### DIFF
--- a/cipher/README.md
+++ b/cipher/README.md
@@ -7,11 +7,10 @@
 [![Project Chat][chat-image]][chat-link]
 [![Build Status][build-image]][build-link]
 
-Traits which define the functionality of [block ciphers], [block modes] and
-[stream ciphers].
+Traits which define the functionality of [block ciphers], [block modes] and [stream ciphers].
 
-See [RustCrypto/block-ciphers] and [RustCrypto/stream-ciphers] for algorithm
-implementations which use these traits.
+See [RustCrypto/block-ciphers] and [RustCrypto/stream-ciphers] for algorithm implementations which
+use these traits.
 
 ## SemVer Policy
 


cipher/README.md
```diff
@@ -7,7 +7,8 @@
 [![Project Chat][chat-image]][chat-link]
 [![Build Status][build-image]][build-link]
 
-Traits which define the functionality of [block ciphers] and [stream ciphers].
+Traits which define the functionality of [block ciphers], [block modes] and
+[stream ciphers].
```
newpavlov: I think this can be kept in one line since total length is just 93 chars.

digest/src/lib.rs
```diff
@@ -1,7 +1,16 @@
...
+#![cfg_attr(docsrs, feature(doc_cfg))]
+#![doc = include_str!("../README.md")]
+#![forbid(unsafe_code)]
+#![doc(
+    html_logo_url = "https://raw.githubusercontent.com/RustCrypto/media/6ee8e381/logo.svg",
+    html_favicon_url = "https://raw.githubusercontent.com/RustCrypto/media/6ee8e381/logo.svg"
+)]
+#![warn(missing_docs, rust_2018_idioms, missing_debug_implementations)]
+
+//! ## Structure
```
newpavlov, line 11: Ideally it would be nice to move this section to the readme as well, though adding the hack for intra-crate links would be a bit annoying. Otherwise docs would look a bit strange since the license section will be before this one.

tarcieri: I think it's fine to have these sections which make heavy use of intra-rustdoc links immediately after the README.md.

It's very tedious to maintain those links otherwise.

cipher/README.md
tarcieri: Fixed in [b91ba27](https://github.com/RustCrypto/traits/pull/2273/commits/b91ba273f686a3acd65622e311990fa605c5dffe)

PR #2274, opened 2026-02-03
crypto-common: tweak `TryKeyInit` docs

newpavlov:


crypto-common/src/lib.rs
```diff
@@ -277,14 +277,13 @@ pub trait TryKeyInit: KeySizeUser + Sized {
...
-    /// - if the key is considered invalid according to rules specific to the implementing type
+    /// If the key is considered invalid according to rules specific to the implementing type.
     fn new(key: &Key<Self>) -> Result<Self, InvalidKey>;
 
     /// Create new value from a variable size key.
     ///
     /// # Errors
-    /// - if the provided slice is the wrong length
-    /// - if the key is considered invalid by [`TryKeyInit::new`]
+    /// If the key is considered invalid according to rules specific to the implementing type.
```
newpavlov, line 286: "Wrong length" is covered by this definition and may depend on implementation of the trait for algorithms which support variable key sizes.

tarcieri (approved):

PR #2277, opened 2026-02-04
Adopt Trusted Publishing

newpavlov:
See https://github.com/RustCrypto/meta/issues/27

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -19,7 +19,7 @@ on:
 jobs:
   publish:
     runs-on: ubuntu-latest
-    environment: release
+    environment: publish
     permissions:
       id-token: write
     steps:


newpavlov:
I tested the config in a separate repository and it appears to be working fine, including cases when several tags are pushed for the same commit. But if several tags are pushed simultaneously you need to look into the actions tab (or the "deployments" page) to see individual runs. By clicking on the commit CI badge you will see `publish` workflow run only for the latest pushed tag.

Note that I intentionally did not set tag/branch protection rules in the release environment since it seems they work on the AND principle and I do not want to duplicate the list of crates. But I made the `master` branch protected and deployment can be done only from it (note that we would need to make backport branches protected as well).  The workflow requires at least one review from me or @tarcieri with forbidden self-review. In the case if one of us will not be available, the setting can be bypassed by administrators (i.e. us) or it can be simply changed.

Finally, I think we should enable the "require trusted publishing for all new versions" checkbox for all our crates to improve security.

newpavlov:
I removed the branch protection rule from the environment since it resulted in the following error: https://github.com/RustCrypto/traits/actions/runs/21659008688/job/62439849297

PR #2083, opened 2025-11-07
cipher: refactor asynchronous stream cipher traits

newpavlov:
This PR moves the `AsyncStreamCipher` trait into separate `async_stream` module and splits into separate encryptor and decryptor traits, making them sub-traits of the counterpart block mode traits.

TODO: add buffering wrapper and traits for buffered async stream ciphers  or do it with inherent methods

Issue #1332, opened 2023-07-03
add trait for length-preserving encryption


ericlagergren:
Sort of similar to https://github.com/RustCrypto/traits/issues/177 and https://github.com/RustCrypto/block-ciphers/issues/48, but for tweakable PRPs like [HCTR2](https://eprint.iacr.org/2021/1441.pdf).

If accepted, I can provide an implementation of HCTR2.

Sample API, trait names are TBD:

```rust
pub trait Lpe {
    /// Encrypts `src` into `dst` using `tweak`
    ///
    /// In is an error if `dst` is not at least as long as `src` or if either are less
    /// than [`BLOCK_SIZE`] bytes long.
    fn seal(&self, dst: &mut [u8], src: &[u8], tweak: &[u8]) -> Result<(), Error> {
        if dst.len() < src.len() { ... }
        dst[..src.len()].copy_from_slice(src);
        match self.seal_in_place(dst, tweak) {
            Ok(()) => Ok(()),
            Err(err) => {
                    dst[..src.len()].zeroize();
                    Err(err)
            }
        }
    }

    fn open(&self, dst: &mut [u8], src: &[u8], tweak: &[u8]) -> Result<(), Error> {
         // etc
    }

    fn seal_in_place(&self, data: &mut [u8], tweak: &[u8]) -> Result<(), Error>;
    fn open_in_place(&self, data: &mut [u8], tweak: &[u8]) -> Result<(), Error>;
}
pub trait LpeMut { ... }
```

Some thoughts:

1. I don't like how the API returns a `Result`, but I'm not sure how else to express that there must be at least one block of plaintext without requiring `Block`, panicking, or a lot of `typenum` gymnastics.
2. I don't know if all LPE algorithms require the length of the plaintext to be at least one block, but I assume most do.
3. `seal` and `open` aren't strictly necessary, but encrypting from `src` to `dst` is very common and it would be a shame if doing that always required copying the entire input.

PR #2280, opened 2026-02-04
cipher: remove `AsyncStreamCipher` trait

newpavlov:
This trait is used only by the CFB mode. It either should be reworked (see #2083) or replaced with inherent methods in the `cfb-mode` crate.

To not block the v0.5 release, it's probably worth to simply remove it for now.


baloo:
Is there a rough idea of what is expected here?

The trait was used in both `cfb-mode` and `cfb8`. I don't believe we can just duplicate in both consumers. Should we revive #2083 and release a 0.5.1 or?

newpavlov:
I haven't decided yet. We have two options:
1) Merge `cfb-mode` and `cfb` (see https://github.com/RustCrypto/block-modes/pull/32) and use inherent methods. In future we may add implementation of the length-preserving encryption traits (see #1332).
2) Add refactored `AsyncStreamCipher` in a later `cipher` release and release CFB mode crates only after that.

PR #2282, opened 2026-02-05
kem: move `Kem` for `*Decapsulate`/`Encapsulate` to associated type

tarcieri:
Also adds (back) a `Decapsulator` trait.

This is effectively a followup to #2243 which added the `Kem` trait but made it a generic argument to `*Decapsulate`/`Encapsulate`, and removed the previous `Decapsulator` trait.

Having it be a generic argument is annoying: these types should simply know what KEM algorithm they implement. The only reason I didn't do it that way at the time was I was having trouble making it work, and that was a fairly large refactor to begin with.

This adds an associated `Kem` to `Encapsulate`, but for `Decapsulate`/`TryDecapsulate` we need a common trait to hang it off of, so this adds back the `Decapsulator` trait.

It's not so bad as the previous hack introduced in #2243 of using `AsRef` to obtain an `EncapsulationKey` from a `Decapsulator` has been replaced with a semantically clear `Decapsulator::encapsulation_key` method.


tarcieri:
This was simple enough to implement in https://github.com/RustCrypto/KEMs/pull/248 and should definitely make it much easier to use, especially in a generic context

PR #2284, opened 2026-02-05
kem: add `FromSeed` trait

tarcieri:
Adds a trait for initializing a KEM keypair deterministically and infallibly from a seed value, as is supported by pretty much all of the NIST PQcrypto KEM submissions.

The trait is blanket impl'd for any type which impls the `Kem` trait and has a `DecapsulationKey` which supports `KeyInit` which should cover most cases.

This is effectively an extraction of a trait from the `ml-kem` crate which was implemented the same way.


tarcieri:
cc @rozbb

PR #2286, opened 2026-02-06
Add lincomb_vartime function to LinearCombination trait

randombit:
https://github.com/RustCrypto/elliptic-curves/issues/1656

### DIFF
--- a/elliptic-curve/src/ops.rs
+++ b/elliptic-curve/src/ops.rs
@@ -179,7 +179,7 @@ where
     /// that it may leak the value of the points and/or scalars due to
     /// variable-time behavior.
     fn lincomb_vartime(points_and_scalars: &PointsAndScalars) -> Self {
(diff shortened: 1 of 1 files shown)


elliptic-curve/src/ops.rs
```diff
@@ -172,6 +172,15 @@ where
...
             .sum()
     }
+
+    /// Calculates `x1 * k1 + ... + xn * kn`.
+    ///
+    /// This is equivalent to [`LinearCombination::lincomb`] except
+    /// that it may leak the value of the points and/or scalars due to
+    /// variable-time behavior.
+    fn lincomb_vartime(points_and_scalars: &PointsAndScalars) -> Self {
+        lincomb(points_and_scalars)
```
tarcieri:
```suggestion
        Self::lincomb(points_and_scalars)
```

PR #2298, opened 2026-02-11
digest: enable and fix workspace-level lints

tarcieri:


digest/src/digest.rs
```diff
@@ -156,11 +157,13 @@ pub trait DynDigest {
 
     /// Write result into provided array and consume the hasher instance.
     ///
+    /// # Errors
     /// Returns error if buffer length is not equal to `output_size`.
```
newpavlov: Remove the "Returns error" part?

newpavlov (approved):

PR #2300, opened 2026-02-12
digest v0.11.0

tarcieri:
## Added
- `CustomizedInit` trait ([#1334])
- `SerializableState` support ([#1369])
- `DynDigestWithOid` wrapper trait ([#1390])
- `VariableOutputCoreCustomized` trait ([#1787], [#2043])
- `buffer_fixed`, `buffer_ct_variable`, `buffer_rt_variable`, and `buffer_xof` macros ([#1799])
- `XofFixedWrapper` ([#1815])
- `CollisionResistance` trait ([#1820])
- `CoreProxy::compose/decompose` methods ([#1898])
- `EagerHash` trait ([#2014])

## Changed
- Replaced `generic-array` with `hybrid-array` ([#1358])
- `crypto-common` dependency bumped to v0.2 ([#1173])
- Edition changed to 2024 and MSRV bumped to 1.85 ([#1759])
- Bump `const-oid` dependency to v0.10 ([#1772])
- `digest::core_api` renamed to `digest::block_api` ([#1799])
- `CtVariableCoreWrapper` renamed to `CtOutWrapper` ([#1799])
- Removed the OID type parameter from `CtOutWrapper` ([#1799])
- Implementations of the `SerializableState` trait ([#1953])
- `new_test!` and `new_mac_test!` macros ([#1958])
- Bump `block-buffer` to v0.11 ([#2082])
- Re-export of `crypto-common` moved to `digest::common` ([#2237])
- Bump `crypto-common` to v0.2 ([#2276])
- Replace `subtle` with `ctutils` ([#2301])

## Removed
- `Mac::new`, `Mac::new_from_slice`, and `Mac::generate_key` methods ([#1173])
- `CoreWrapper`, `RtVariableCoreWrapper`, and `XofReaderCoreWrapper` types ([#1799])
- `io::Write/Read` implementations in favor of the `digest_io::IoWrapper` type ([#1809])
- `VariableOutput` trait ([#2043])
- Implementation of `subtle::ConstantTimeEq` for `CtOutput`. Note that implementation of
  `PartialEq`/`Eq` trait is still const time. ([#2292])

[#1173]: https://github.com/RustCrypto/traits/pull/1173
[#1334]: https://github.com/RustCrypto/traits/pull/1334
[#1358]: https://github.com/RustCrypto/traits/pull/1358
[#1369]: https://github.com/RustCrypto/traits/pull/1369
[#1390]: https://github.com/RustCrypto/traits/pull/1390
[...]

Issue #2275, opened 2026-02-03
Replace `subtle` with `ctutils`?


newpavlov:
See https://github.com/RustCrypto/meta/issues/29

In particular, `subtle` is part of the `digest` public API via the `CtOutput` type. `elliptic-curve` also uses it as a dependency. We probably should migrate before cutting the next breaking release.

tarcieri:
It continues to be a big part of the public API of `elliptic-curve` because `ff`/`group` use it: https://github.com/zkcrypto/ff/issues/148

I was hoping to come to come kind of consensus about what to do there but haven't heard back yet.

newpavlov:
WDYT about `digest`? Should we migrate it before the next release or wait for decision on `elliptic-curve`?

tarcieri:
Yeah, it's part of the `digest` public API due to the trait impls on `CtOutput`. I guess that's where we need to start making calls on this (fortunately we can release `cipher` first). 

Give me a day or so to see if I can get some clarity on what direction others want to go. Ideally I think we'd have one and only one constant-time library like that to deal with.

tarcieri:
Okay, I've gotten in touch but it seems like things might not get sorted out until next week.

I don't think it would be the end of the world to release `digest` with `subtle` support, then later add `ctutils` support if we decide to go that direction. We could simply not use the `subtle` feature ourselves. It's a very small piece of API surface.

Edit: however, where it would be a problem is if I can actually coordinate a `subtle` v3 and we want to upgrade, so let me see if I can get clarity on that.

PR #2301, opened 2026-02-12
digest: replace `subtle` with `ctutils`

tarcieri:
As discussed in #2275 and #2300, we are migrating from `subtle` to our new `ctutils` library:

https://github.com/RustCrypto/utils/tree/master/ctutils

### DIFF
--- a/.github/workflows/elliptic-curve.yml
+++ b/.github/workflows/elliptic-curve.yml
@@ -102,6 +102,8 @@ jobs:
       MIRIFLAGS: "-Zmiri-symbolic-alignment-check -Zmiri-strict-provenance"
     steps:
       - uses: actions/checkout@v6
-      - uses: dtolnay/rust-toolchain@nightly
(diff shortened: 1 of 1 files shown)


digest/Cargo.toml
```diff
@@ -28,7 +28,7 @@ sha2 = "0.11.0-rc.5"
 [features]
 default = ["block-api"]
 block-api = ["block-buffer"] # Enable block API traits
-mac = ["subtle"] # Enable MAC traits
+mac = ["dep:ctutils"] # Enable MAC traits
 rand_core = ["common/rand_core"] # Enable random key generation methods
 getrandom = ["common/getrandom", "rand_core"]
 oid = ["const-oid"]
```
tarcieri, line 34: Sidebar: seems like we should use `dep:block-buffer` and `dep:const-oid`. I'll open a followup PR.

tarcieri:
Ugh miri tests failing everywhere 😢 Need to pin to the previous nightly

Issue #2304, opened 2026-02-16
`BlockSizes` up to only 256 bytes makes it impossible to implement `BlockSizeUser` for wide-block ciphers


NumberFour8:
The `BlockSizes` trait is implemented only up to 256 bytes.

https://github.com/RustCrypto/traits/blob/593a0ea151d49d2c8de4b10be78055f5c5e985d9/crypto-common/src/lib.rs#L89

It therefore makes it impossible for a wide-block cipher (with a fixed block length) to implement `BlockSizeUser`.

Would it perhaps make sense to lift this up (at least 2k ?), or even make it unlimited.

newpavlov:
We were previously relying on this constraint in the [`block-buffer`](https://docs.rs/block-buffer/) crate to store cursor position as `u8`. It looks like we removed it in favor of runtime checks during buffer construction (the main motivation was to remove `crypto-common` dependency from `block-buffer`), but forgot to lift the restriction in `crypto-common`.

I guess we could remove it in a future breaking release, but I am not sure if we want to do it in the near future.

Do you plan to implement an algorithm which requires such block sizes?

NumberFour8:
> We were previously relying on this constraint in the [`block-buffer`](https://docs.rs/block-buffer/) crate to store cursor position as `u8`. It looks like we removed it in favor of runtime checks during buffer construction (the main motivation was to remove `crypto-common` dependency from `block-buffer`), but forgot to lift the restriction in `crypto-common`.
> 
> I guess we could remove it in a future breaking release, but I am not sure if we want to do it in the near future.
> 
> Do you plan to implement an algorithm which requires such block sizes?

I'm currently implementing Lioness wide-block cipher, with block sizes up to 1536 bytes. I guess until it is removed, I would need to stay with the older version of traits, or remove the `BlockSizeUser` there.

NumberFour8:
Would the limitation removal be actually breaking?

newpavlov:
I think yes. Downstream code may rely on the promise in the docs:
>Trait implemented for supported block sizes, i.e. for types from `U1` to `U255`.

We would need to review all our uses of `BlockSize` and probably remove the trait completely in favor of `ArraySize`.

PR #2309, opened 2026-02-18
crypto-common: remove `BlockSizes` trait

newpavlov:
`BlockSizeUser` is no longer limited by block sizes smaller than 256 bytes. `BlockSizes` was moved to `block-buffer` (see https://github.com/RustCrypto/utils/pull/1455) and used in bounds only where it's needed.

Closes: #2304

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -111,7 +111,7 @@ checksum = "9330f8b2ff13f34540b44e946ef35111825727b38d33286ef986142615121801"
 
 [[package]]
 name = "cipher"
-version = "0.5.0"
+version = "0.5.1"
 dependencies = [
  "blobby",
  "block-buffer",
@@ -186,7 +186,7 @@ dependencies = [
 
 [[package]]
 name = "crypto-common"
-version = "0.2.0"
+version = "0.2.1"
 dependencies = [
  "getrandom",
  "hybrid-array",
@@ -216,7 +216,7 @@ dependencies = [
 
 [[package]]
 name = "digest"
-version = "0.11.0"
+version = "0.11.1"
 dependencies = [
  "blobby",
  "block-buffer",
--- a/cipher/CHANGELOG.md
+++ b/cipher/CHANGELOG.md
@@ -6,12 +6,19 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## 0.5.1 (UNRELEASED)
+### Changed
+- Block sizes no longer bounded by `crypto_common::BlockSizes` ([#2309])
+- `StreamCipherCoreWrapper` is now bounded by `block_buffer::BlockSizes`
+  instead of `crypto_common::BlockSizes` ([#2309])
+- Bump `block-buffer` dependency to v0.12 ([#2309])
+
 ### Fixed
 - `Key:` option in the `stream_cipher_bench` macro ([#2290])
 
 [#2290]: https://github.com/RustCrypto/traits/pull/2290
+[#2309]: https://github.com/RustCrypto/traits/pull/2309
 
-## 0.5.0 (2026-02-04)
+## 0.5.0 (2026-02-04) [YANKED]
 ### Added
 - Traits for tweakable block ciphers ([#1721])
 - Methods for writing keystream ([#1907])
--- a/cipher/Cargo.toml
+++ b/cipher/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cipher"
-version = "0.5.0"
+version = "0.5.1"
 authors = ["RustCrypto Developers"]
 edition = "2024"
 rust-version = "1.85"
--- a/crypto-common/CHANGELOG.md
+++ b/crypto-common/CHANGELOG.md
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.2.1 (UNRELEASED)
+### Changed
+- `BlockSizeUser::BlockSize` is no longer bounded by `BlockSizes` ([#2309])
+
+### Removed
+- `BlockSizes` trait ([#2309])
+
+[#2309]: https://github.com/RustCrypto/traits/pull/2309
+
 ## 0.2.0 (2026-02-04)
 ### Added
 - Sealed `BlockSizes` trait implemented for types from `U1` to `U255` ([#1172])
--- a/crypto-common/Cargo.toml
+++ b/crypto-common/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "crypto-common"
-version = "0.2.0"
+version = "0.2.1"
 authors = ["RustCrypto Developers"]
 edition = "2024"
 rust-version = "1.85"
--- a/digest/CHANGELOG.md
+++ b/digest/CHANGELOG.md
@@ -5,7 +5,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## 0.11.0 (2026-02-13)
+## 0.11.1 (UNRELEASED)
+### Added
+- `SmallBlockSizeUser` helper trait. The trait is a sub-trait of `BlockSizeUser`
+  with `BlockSize` bounded by `block_buffer::BlockSizes`. ([#2309])
+
+### Changed
+- `BufferKindUser` is now a sub-trait of `SmallBlockSizeUser` ([#2309])
+
+[#2309]: https://github.com/RustCrypto/traits/pull/2309
+
+## 0.11.0 (2026-02-13) [YANKED]
 ### Added
 - `CustomizedInit` trait ([#1334])
 - `SerializableState` support ([#1369])
--- a/digest/Cargo.toml
+++ b/digest/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "digest"
-version = "0.11.0"
+version = "0.11.1"
 authors = ["RustCrypto Developers"]
 edition = "2024"
 rust-version = "1.85"


digest/src/buffer_macros/fixed.rs
```diff
@@ -163,7 +163,9 @@ macro_rules! buffer_fixed {
         ($core_ty:ty);
         OutputSizeUser $($trait_name:ident)*;
     ) => {
-        impl$(< $( $lt $( : $clt $(+ $dlt )* )? ),+ >)? $crate::OutputSizeUser for $name$(< $( $lt ),+ >)? {
+        impl$(< $( $lt $( : $clt $(+ $dlt )* )? ),+ >)? $crate::OutputSizeUser for $name$(< $( $lt ),+ >)?
+        where <Self as $crate::block_api::BlockSizeUser>::BlockSize: $crate::block_buffer::BlockSizes,
```
newpavlov: These `where` bounds are pretty annoying and I am not sure they would not leak in the wrapped type APIs.

Unfortunately, I couldn't find a good way for removing them without making a duplicate of `BlockSizeUser`.

Simply defining a new sub-trait `SmallBlockSizeUser: BlockSizeUser where Self::BlockSize: block_buffer::BlockSizes` unfortunately does not work since the `where` bound still leaks and we have to explicitly duplicate it in users...

UPD: Yeah, `where` does leak as we can see with `elliptic-curve`...

tarcieri:
Hmm, fun times with `elliptic-curve` and its (transitive) dependency on `hmac` there...

digest/src/buffer_macros/fixed.rs
newpavlov: We are dealing with 10+ year old Rust [issue](https://github.com/rust-lang/rust/issues/20671) here... :(

digest/src/block_api.rs
```diff
@@ -24,31 +24,53 @@ pub trait UpdateCore: BlockSizeUser {
...
+/// Sub-trait of [`BlockSizeUser`] implemented if `BlockSize` is
+/// bigger than `U0` and smaller than `U256`.
+///
+/// This trait relies on the hack suggested [here][0] to work around
+/// the long standing Rust issue regarding non-propagation of `where` bounds.
+///
+/// [0]: https://github.com/rust-lang/rust/issues/20671#issuecomment-1905186183
+pub trait SmallBlockSizeUser:
+    BlockSizeUser<BlockSize = <Self as SmallBlockSizeUser>::_BlockSize>
+{
```
newpavlov, line 36: Welp, this hack does the thing, but it's surely not pretty... I wonder if we should move it to `crypto-common`. It would mean addition of optional `block-buffer` dependency for it which may be fine.

baloo:
This is a breaking change that seems to confuse cargo in downstream crates.
I think the minimum would have been to bump the version in `digest` and `cipher`:
```
diff --git a/cipher/Cargo.toml b/cipher/Cargo.toml
index 4399b4fb88..a6a41be332 100644
--- a/cipher/Cargo.toml
+++ b/cipher/Cargo.toml
@@ -13,7 +13,7 @@
 description = "Traits for describing block ciphers and stream ciphers"

 [dependencies]
-common = { version = "0.2", package = "crypto-common" }
+common = { version = "0.2.1", package = "crypto-common" }
 inout = "0.2.2"

 # optional dependencies
diff --git a/digest/Cargo.toml b/digest/Cargo.toml
index bc08a55c7d..79eb264661 100644
--- a/digest/Cargo.toml
+++ b/digest/Cargo.toml
@@ -13,7 +13,7 @@
 description = "Traits for cryptographic hash functions and message authentication codes"

 [dependencies]
-common = { version = "0.2", package = "crypto-common" }
+common = { version = "0.2.1", package = "crypto-common" }

 # optional dependencies
 block-buffer = { version = "0.12", optional = true }
 ```
 
 My worry is that we'll break chacha20 and rand 0.10 with this late change.
 
 Repro in https://github.com/baloo/blocksizes-repro:
 ```
 $ cargo check
    Checking cipher v0.5.1
error[E0277]: the trait bound `BS: crypto_common::sealed::BlockSizes` is not satisfied
  --> /home/arthur_gautier/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cipher-0.5.1/src/block/ctx.rs:15:22
   |
15 |     type BlockSize = BS;
   |                      ^^ the trait `crypto_common::typenum::NonZero` is not implemented for `BS`
   |
   = note: required for `BS` to implement `crypto_common::sealed::BlockSizes`
   = note: required for `<block::ctx::BlockCtx<'_, '_, BS> as BlockSizeUser>::BlockSize` to implement `crypto_common::BlockSizes`
note: required by a bound in `crypto_common::BlockSizeUser::BlockSize`
  --> /home/arthur_gautier/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/crypto-common-0.2.0/src/lib.rs:61:21
   |
61 |     type BlockSize: BlockSizes;
   |                     ^^^^^^^^^^ required by this bound in `BlockSizeUser::BlockSize`
help: consider further restricting type parameter `BS` with trait `NonZero`
   |
14 | impl<BS: ArraySize + crypto_common::typenum::NonZero> BlockSizeUser for BlockCtx<'_, '_, BS> {
   |                    +++++++++++++++++++++++++++++++++
```

(minimum fix would be to bump crypto-common version in the dependency, re-release digest and cipher, and yank the x.x.1, but that's not going to be enough (from tests with patch.crates-io I'm running)).

tarcieri:
@baloo have you tested it with `chacha20`? I thought this would only impact block ciphers

baloo:
@tarcieri updated the original message with a reproducer.

newpavlov:
I and @tarcieri  discussed it previously in Zulip and decided to try the yanking approach. The impact should be minimal since we did not release any `digest`-dependent crates and the only released `cipher`-dependent crate is `chacha20`. It's used by `rand`, but it does not enable the `cipher` feature, so it should not be affected.

To trigger the breakage you need to use both `crypto-common` v0.2.0 and `cipher` v0.5.1 (or `digest` v0.11.1) simultaneously. I raised this possibility during the discussion, but it's is unlikely to happen in practice. If you pulled `chacha20` before the new releases you will use v0.2.0/v0.5.0, running `cargo update` would update both versions. Adding (or updating from the previous minor version) `chacha20` should pull only v0.2.1/v0.5.1 since v0.2.0/v0.5.0 are yanked.

baloo:
If you run `cargo update` that works, but dependabot would only pull one dependency at a time in its default config.

This is showcased in https://github.com/RustCrypto/elliptic-curves/pull/1669 and there is no dependency I can bump to fix that (I have no direct edge to `cipher` to bump it to 0.5.1 in elliptic-curves workspace).

tarcieri:
In our discussions I said the version requirements should be updated to ensure compatibility.

I’m surprised it didn’t trigger a `minimum-versions` failure?

newpavlov:
@baloo 
Yes, I consider it included into the "minimal impact". IMO we should use Dependabot only for updating Cargo.lock and use deps.rs badges to nudge dep updates in Cargo.toml.

@tarcieri 
>In our discussions I said the version requirements should be updated to ensure compatibility.

You mean we should release `chacha20` v0.11.0 with `cipher` set to v0.5.1 and yank v0.11.0? I don't think it's necessary, but you could do it if you want.

>I’m surprised it didn’t trigger a minimum-versions failure?

IIUC `minimum-versions` do not pull yanked dependencies.

baloo:
There might be a bug in regard to the way we override the dependencies in the workspaces (bug or lack of support for `patch.crates-io`).
https://github.com/RustCrypto/actions/blob/master/.github/workflows/minimal-versions.yml

newpavlov:
Merging https://github.com/RustCrypto/elliptic-curves/pull/1670 should fix the breakage after rebase.

UPD: Ah, it does not, since the crates in the repo use pre-release versions of hash crates. We should release the hash crates and update both hashes and `digest`.

tarcieri:
@newpavlov I’m not a computer now, but it sounds like `cipher` needs a stricter requirement for `crypto-common`, no?

newpavlov:
@tarcieri 
I don't think it does since yanked v0.5.0 should not be pulled without a pre-existing Cargo.lock, so simple `version = "0.5"` should be sufficient in practice.

newpavlov:
The problem is with already existing pre-releases of the `sha3`, `rc5`, `rc6`, and block mode crates crates which rely on the removed `BlockSizes` trait in their source code and thus are effectively broken now.

tarcieri:
> I don't think it does since yanked v0.5.0 should not be pulled without a pre-existing Cargo.lock

But several people will have pre-existing `Cargo.lock`s with the existing crate.

If the versions are specified correctly, there should never be a breakage, even for users who have old yanked versions in their `Cargo.lock`.

newpavlov:
>But several people will have pre-existing Cargo.locks with the existing crate.

Their Cargo.lock should contain `crypto-common` v0.2.0 and `cipher` v0.5.0. Updating dependencies with `cargo update` should bump both to v0.2.1 and v0.5.1. You could trigger the breakage by specifically running `cargo update -p cipher`, but I think we can discard such scenarios as improbable.

The elliptic-curve breakage is caused by an incorrect automatic update from the previous release candidate and it would require a manual intervention either way because of the `sha3` pre-release used in the repository.

Moreover, the only proper implementation crate release we have is `chacha20`. IMO any breakage in pre-release crates is not a big problem.

baloo:
My comment was more: Cargo needs an incentive to update a dependency (currently using a yanked dependency is not enough), otherwise it will continue to use the one version it is using.

I'm not saying `cargo update` won't fix that, but in larger code-bases (where one of your dependency might be lagging behind/looser dependencies than they should/not so good api stability in versions) you may not be able to use that.
I myself struggle sometimes with the dependency graph of rustcrypto and I've spent a fair bit of time working on it.
I think we will open ourselves to people opening issues because of compilation failure, and unless we want to spend time on user-support, we should fix that.

I think the better approach is to yank cipher 0.5.1 and digest 0.5.1, force the crypto-common dependency to 0.2.1 (the incentive for cargo to bump), then re-release cipher and digest to 0.5.2.

tarcieri:
I would agree they should use more specific `crypto-common` dependency that always works. Given the choice between that and a requirement that sometimes causes build failures, we should always pick the former

baloo:
I've tried to reproduce it in minimal-version, but sadly cargo will not pull a yanked version.
https://github.com/RustCrypto/actions/pull/57

newpavlov:
@baloo 
My position is that we can treat the yanked versions as if they did not exist in the first place because we did not cut any proper releases which depend on them apart from `chacha20`. In other words, `crypto-common` v0.2.0 should not be in users' Cargo.lock unless they use release candidates and if they do they should be ready to deal with any potential breakage (which is simple enough in this case).

This is the only reason why releasing breaking changes in 0.x.1 and yanking of 0.x.0 was considered "fine" in this case.

>I think the better approach is to yank cipher 0.5.1 and digest 0.5.1, force the crypto-common dependency to 0.2.1 (the incentive for cargo to bump), then re-release cipher and digest to 0.5.2.

At this point I would prefer to just un-yank 0.x.0, yank the 0.x.1, and cut proper new breaking releases of `crypto-common`, `cipher`, and `digest` (and yank `chacha20` v0.10.0 and release v0.10.1 which would depend on `cipher` v0.6.0). It was initially what I wanted to do, but @tarcieri argued for 0.x.0 yanking during the previously mentioned private discussion.

>I've tried to reproduce it in minimal-version, but sadly cargo will not pull a yanked version.

What do you mean "sadly"? It's by design and why I think it's fine to skip the patch number while specifying dependencies.

PR #2310, opened 2026-02-18
fix: add missing hasher2.reset() before chunks loop in dev tests

radik878:
The test functions fixed_reset_test, variable_reset_test, and xof_reset_test were not resetting hasher2 before the chunks loop, causing state accumulation across different chunk size iterations.


newpavlov:
The state is reset by finalize_reset functions.

PR #1469, opened 2024-01-10
Bump `crypto-common` and `hybrid-array`

tarcieri:
Bumps the following dependencies:

- `block-buffer` => v0.11.0-pre.4
- `block-padding` => v0.4.0-pre.4
- `crypto-common` => v0.2.0-pre.4
- `hybrid-array` => v0.2.0-rc.0
- `inout` => v0.2.0-pre.4

Also prepares the following releases:

- `cipher` v0.5.0-pre.2
- `digest` v0.11.0-pre.4

PR #2317, opened 2026-02-26
Release `universal-hash` v0.6.0

newpavlov:
### Changed
- Edition changed to 2024 and MSRV bumped to 1.85 ([#1759])
- Replaced `generic-array` with `hybrid-array` ([#1385])
- Bump `crypto-common` dependency to v0.2 ([#1385])
- Re-export of `crypto-common` moved to `universal_hash::common` ([#2237])

[#1385]: https://github.com/RustCrypto/traits/pull/1385
[#1759]: https://github.com/RustCrypto/traits/pull/1759
[#2237]: https://github.com/RustCrypto/traits/pull/2237

### DIFF
--- a/universal-hash/CHANGELOG.md
+++ b/universal-hash/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## 0.6.0 (2026-02-26)
+## 0.6.0 (2026-02-27)
 ### Changed
 - Edition changed to 2024 and MSRV bumped to 1.85 ([#1759])
 - Replaced `generic-array` with `hybrid-array` ([#1385])


tarcieri:
> Migrate to doc_auto_cfg (https://github.com/RustCrypto/traits/pull/1370)

That's obsolete? It's `doc_cfg` again now.

I was hoping to do a pass over this to get the new clippy config wired up before a final release.

tarcieri:
@newpavlov can you rebase on #2318?

tarcieri:
> Bump hybrid-array dependency to v0.2 (https://github.com/RustCrypto/traits/pull/1469)

We migrated from `generic-array` to `hybrid-array`, which is what the CHANGELOG should note, as in https://github.com/RustCrypto/traits/blob/master/digest/CHANGELOG.md#changed-1

universal-hash/Cargo.toml
baloo, line 16:
```suggestion
common = { version = "0.2.1", package = "crypto-common" }
```

universal-hash 0.6.0 will fail to build with crypto-common 0.2.0, it needs 0.2.1.

(context: https://github.com/RustCrypto/traits/pull/2309#issuecomment-3969263180)

newpavlov: As I commented [here](https://github.com/RustCrypto/traits/pull/2309#issuecomment-3972716629), I don't think we need the patch number. And I certainly do not want to bother with patch numbers in hash and cipher implementation crates. Additionally, the current version is consistent with the already released `cipher` and `digest`.

tarcieri: I think it's easy enough to do here (and in `cipher` and `digest`) and would prevent some breakages, but if you really want to avoid it whatever

tarcieri: Like, it would be nice if these Dependabot upgrades Just Worked: https://github.com/RustCrypto/signatures/actions/runs/22490337172/job/65150558373?pr=1234

newpavlov: Quoting my previous comment:
>IMO we should use Dependabot only for updating Cargo.lock and use deps.rs badges to nudge dep updates in Cargo.toml.

Potentially breaking dependency updates should be done manually. Updating from 0.x.a to 0.y.b or from 0.a.b-rc.x to 0.a.b-rc.y can not "just work" for all updates.

tarcieri:
I just realized `universal-hash` v0.6.0 still hash `subtle` as a dependency. Should we migrate it to `ctutils`? (which would be yet another breaking change, but...)

newpavlov:
`subtle` is a private dependency, so I think we can migrate to `ctutils` in a patch release.

Issue #2316, opened 2026-02-26
universal-hash-0.6.0-rc.10 not updated to remove common::BlockSizes


Harrygiel:
Following #2309 , universal-hash-0.6.0-rc.10 can't compile anymore:

error[E0432]: unresolved import `common::BlockSizes`
  --> /home//.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/universal-hash-0.6.0-rc.10/src/lib.rs:16:29
   |
16 | use common::{BlockSizeUser, BlockSizes, ParBlocksSizeUser, array::Array};
   |                             ^^^^^^^^^^
   |                             |
   |                             no `BlockSizes` in the root
   |                             help: a similar name exists in the module: `BlockSizeUser`

In the same way, ctr-0.10.0-rc.3 can't compile:
error[E0432]: unresolved import `cipher::common::BlockSizes`
 --> /home/<user>/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/ctr-0.10.0-rc.3/src/ctr_core.rs:7:14
  |
7 |     common::{BlockSizes, InnerUser, IvSizeUser},
  |              ^^^^^^^^^^
  |              |
  |              no `BlockSizes` in the root
  |              help: a similar name exists in the module: `BlockSizeUser`

Yet [universal-hash src](https://github.com/RustCrypto/traits/blob/master/universal-hash/src/lib.rs) was updated 17 hours ago with a fix, but without a release tag.

For example, the following tree is then now broken:
```
crypto-common v0.2.1
└── universal-hash v0.6.0-rc.10
    └── polyval v0.7.0-rc.9
        └── ghash v0.6.0-rc.6
            └── aes-gcm v0.11.0-rc.3 (*)
```

Is it possible to tag a new version of **universal-hash** and  **ctr** ?
Respectfully,

newpavlov:
My intent is to cut proper (i.e. non-RC) releases of the `universal-hash`/`cipher`/`digest` crates in the near future. AEAD crates will stay in the RC status for the time being, since `aead` is likely to change a bit.

Harrygiel:
@newpavlov I understand. Would you please be able to give an approximate time of release ?
It would help me measure how long will my project be down:
1) I can't rollback crypto-common as it allow the compilation of universal-hash, but break the compilation of rand => chacha20 => cipher =>  crypto-common (cipher 0.5 was yanked)
[...]

PR #2321, opened 2026-02-27
crypto-common: note that 0.2.0 was [YANKED]

### LINKED ISSUE #2319: Please document why a crate was yanked
https://crates.io/crates/crypto-common/0.2.0 has been yanked, but I can't find the reason anywhere. The place I'd expect that to be would be in the CHANGELOG; is there somewhere else I should look?

dishmaker:
Resolves #2319

### DIFF
--- a/crypto-common/CHANGELOG.md
+++ b/crypto-common/CHANGELOG.md
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 [#2309]: https://github.com/RustCrypto/traits/pull/2309
 
-## 0.2.0 (2026-02-04)
+## 0.2.0 (2026-02-04) [YANKED]
 ### Added
 - Sealed `BlockSizes` trait implemented for types from `U1` to `U255` ([#1172])
 - `SerializableState` trait under `hazmat` module ([#1369])


shepmaster:
Note that I had asked *why* this crate was yanked, which this commit does not explain. As someone that is using crypto-common 0.2.0, I don't have any context to understand if I *need* to upgrade.

newpavlov:
As mentioned in the changelog, v0.2.1 has introduced a breaking change. Technically we should've released v0.3.0, but since we thought that the impact will be minimal, we decided to sneak it in a patch release (see discussion in #2309).

>if I need to upgrade

I think yanking of the version is a pretty clear signal that you should. But you may want to postpone the upgrade if you use existing pre-releases (e.g. see #2316) since the upgrade may break your builds.

shepmaster:
> As mentioned in the changelog, v0.2.1 has introduced a breaking change

I did not look at the rest of the changelog, only at this commit's diff, so I apologize. However, now that I do look at it, I don't think that the text in the changelog explains why 0.2.0 was yanked or clearly indicates what the breaking change in 0.2.1 would have been (or that it *is* a breaking change).

I already knew that the version was yanked as my builds / crates.io told me that already, so marking that version as "yanked" doesn't provide any information, other than it was an intentional yank and not a typo. What is missing is the why the yank happened. Having a dedicated sentence in the changelog along the lines of "this version was yanked because we forgot to make an important breaking change and decided to include that in 0.2.1" (or whatever phrasing is true and accurate), potentially with a link to the discussion you have above, would go a long way.

For what it's worth, in my changelogs, I include this with each relevant bullet point:

```markdown
This is a **breaking change**.
```

> I think yanking of the version is a pretty clear signal that you should.

Maybe for most people, but my interest arises because this crate has stopped the [nightly playground builds from occurring](https://github.com/rust-lang/rust-playground/actions/runs/22471033089/job/65087786770). In the vast majority of cases that someone has yanked a version, it doesn't actually matter to the playground, but I can't tell that without some sort of documentation / explanation.

newpavlov:
I will add a note for the yanked versions.

>Maybe for most people, but my interest arises because this crate has stopped the nightly playground builds from occurring. 

I wonder which crate is responsible for this:

>error: failed to select a version for the requirement `crypto-common = "=0.2.0"`

Such bounds are notoriously fragile.

newpavlov:
See #2325

shepmaster:
> I wonder which crate is responsible for this:

Oh, that's [something I do specifically](https://github.com/rust-lang/rust-playground/blob/main/compiler/base/Cargo.toml). I want to know when there are issues, but I also don't want to rebuild every crate every night.

PR #2341, opened 2026-03-10
digest: remove `Clone` from `MacTraits`

### LINKED ISSUE #2329: digest: remove `Clone` from `MacTraits` in `buffer_fixed!`
See https://github.com/RustCrypto/MACs/pull/250

This is technically a breaking change. We could release it in `digest` v0.11.2 after yanking v0.11.1 or leave it for a future breaking release since the workaround is fine for now.

tarcieri:
NOTE: this is a breaking change! We are doing it in spite of `digest` having a stable release as the downstream crates that depend on this functionality have not yet had stable releases, so it should only impact people who are currently relying on prereleases

Closes #2329

### DIFF
--- a/digest/CHANGELOG.md
+++ b/digest/CHANGELOG.md
@@ -8,9 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## 0.11.2 (UNRELEASED)
 ### Changed
 - Do not implement `Clone` as part of `(Reset)MacTraits` in the `buffer_fixed!` macro ([#2341])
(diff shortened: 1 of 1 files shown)


tarcieri:
Looks like this will need a companion PR in https://github.com/RustCrypto/MACs

newpavlov:
Huh, the fun of circular dependencies... I can prepare `patch`-based PRs tomorrow.

newpavlov:
I merged the `hmac` fix. I think we can patch it as part of this PR.

tarcieri:
I’m busy for the next few days if you want to edit or recreate this PR

PR #2351, opened 2026-03-31
[WIP] elliptic-curve: transition from `subtle` to `ctutils`

tarcieri:
We can't completely remove `subtle` because it's used by the `ff`/`group` traits, however as much s possible this migrates from `subtle` to `ctutils`.

It's a bit annoying because we end up with a mixed API of `ctutils` and `subtle`, but perhaps we can get `ff`/`group` to migrate upstream as well: RustCrypto/ff#148.

See also: #2275

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -180,7 +180,6 @@ dependencies = [
  "hybrid-array",
  "num-traits",
  "rand_core",
- "subtle",
  "zeroize",
 ]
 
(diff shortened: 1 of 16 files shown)


tarcieri:
After discussing this with the upstream devs, they've effectively accepted the RFC but deferred the change until `ff`/`group` v0.15.

So we'll get to this in `elliptic-curve` v0.15. I definitely plan on making sure it happens in that release.

PR #2387, opened 2026-04-16
Implement SecretKey::from_pem to decode variety of keys

DoumanAsh:
As suggested in https://github.com/RustCrypto/signatures/pull/1305 created API to decode SecretKey from variety of PEM encodings

### DIFF
--- a/elliptic-curve/src/secret_key.rs
+++ b/elliptic-curve/src/secret_key.rs
@@ -297,7 +297,7 @@ where
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
 pub enum PemParseError {
     ///Indicates invalid PEM string
-    Pem(pem::Error),
+    Pem(der::Error),
     ///Indicates invalid pkcs8 EC key
     Pkcs8(::pkcs8::Error),
     ///Indicates invalid Sec1 EC key
@@ -307,9 +307,9 @@ pub enum PemParseError {
 }
 
 #[cfg(feature = "pem")]
-impl From<pem::Error> for PemParseError {
+impl From<der::Error> for PemParseError {
     #[inline(always)]
-    fn from(error: pem::Error) -> Self {
+    fn from(error: der::Error) -> Self {
         Self::Pem(error)
     }
 }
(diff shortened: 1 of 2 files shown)


elliptic-curve/src/secret_key.rs
```diff
@@ -293,6 +293,90 @@ where
...
+    /// - `SEC1` - requires feature `sec1`
+    /// - `PKCS #8` - requires feature `pkcs8`
+    ///
+    /// # Errors
+    /// - If `pem` is not valid PEM encoded private key
+    /// - If label within `pem` is not known valid label
+    /// - If label is valid, but unable to decode DER content of the PEM file
+    #[cfg(feature = "pem")]
+    pub fn from_pem(pem: &str) -> ::core::result::Result<Self, PemParseError> {
+        let (label, der_bytes) = pem::decode_vec(pem.as_bytes()).map_err(PemParseError::Pem)?;
```
tarcieri: This is making an unnecessary copy of the secret key material.

Can you implement it like you had it before, where it first checks for a PKCS#8 label with `validate_pem_label`, and if so passes the full PEM to the PKCS#8 decoder, and if not checks for a SEC1 label and the same with the SEC1 decoder?

Those implementations support on-the-fly PEM decoding which does not need to first decode the PEM to a `Vec`.

DoumanAsh: Do you mean using `SecretDocument`?
I did the same as in my signatures PR now so it validates PEM label then passes reference to SecretDocument's bytes

Do you really want to pass full PEM string to be decoded or?

tarcieri: Oh, I missed you were doing that. That will also make a copy of the key.

Can you use `pem_rfc7468::decode_label` to decode the label first, then dispatch on that?

DoumanAsh: Ok, but it returns only label, so it would mean that it checks label twice

tarcieri: That's certainly preferable to making a copy of the secret key data

elliptic-curve/src/secret_key.rs
```diff
@@ -293,6 +293,88 @@ where
...
+        match self {
+            Self::Pem(error) => fmt.write_fmt(format_args!("Failed to parse PEM: {error}")),
+            Self::UnknownLabel => fmt.write_str("Unrecognized key label"),
+            Self::Pkcs8(error) => fmt.write_fmt(format_args!("Faoled to parse Pkcs8 key: {error}")),
+            Self::Sec1(error) => fmt.write_fmt(format_args!("Faoled to parse SEC1 key: {error}")),
+        }
+    }
+}
+
+#[cfg(all(feature = "std", feature = "pem"))]
```
tarcieri:
```suggestion
#[cfg(feature = "pem")]
```

tarcieri (approved):

elliptic-curve/src/secret_key.rs
```diff
@@ -293,6 +293,88 @@ where
...
+}
+
+#[cfg(feature = "pem")]
+impl fmt::Display for PemParseError {
+    fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Pem(error) => fmt.write_fmt(format_args!("Failed to parse PEM: {error}")),
+            Self::UnknownLabel => fmt.write_str("Unrecognized key label"),
+            Self::Pkcs8(error) => fmt.write_fmt(format_args!("Faoled to parse Pkcs8 key: {error}")),
+            Self::Sec1(error) => fmt.write_fmt(format_args!("Faoled to parse SEC1 key: {error}")),
```
dishmaker:
```suggestion
            Self::Sec1(error) => fmt.write_fmt(format_args!("Failed to parse SEC1 key: {error}")),
```

PR #2395, opened 2026-04-27
digest: add `TryCustomizedInit` trait

newpavlov:
Some algorithms (e.g. Ascon-CXOF128 and `bash-prg-hash`) place restrictions on customization strings, so with the current version of `digest` we either have to panic on invalid strings or use inherent methods. This PR amends this by introducing a fallible variant of `CustomizedInit` with the blanket impl to act as a bridge between the traits.

### DIFF
--- a/digest/src/lib.rs
+++ b/digest/src/lib.rs
@@ -218,6 +218,14 @@ pub trait TryCustomizedInit: Sized {
     fn try_new_customized(customization: &[u8]) -> Result<Self, Self::Error>;
 }
 
+impl<T: CustomizedInit> TryCustomizedInit for T {
+    type Error = core::convert::Infallible;
+
+    fn try_new_customized(customization: &[u8]) -> Result<Self, Self::Error> {
+        Ok(Self::new_customized(customization))
+    }
+}
+
 /// Types with a certain collision resistance.
 pub trait CollisionResistance {
     /// Collision resistance in bytes.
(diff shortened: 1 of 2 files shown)


digest/src/lib.rs
```diff
@@ -205,6 +205,27 @@ pub trait CustomizedInit: Sized {
...
+    type Error;
+
+    /// Create new hasher instance with the given customization string.
+    ///
+    /// # Errors
+    /// If the provided customization string is not valid for the hash function.
+    fn try_new_customized(customization: &[u8]) -> Result<Self, Self::Error>;
+}
+
+impl<T: CustomizedInit> TryCustomizedInit for T {
```
newpavlov, line 221: Usually I would prefer to do this instead:
```rust
pub trait TryCustomizedInit: Sized {
    type Error;
    fn try_new_customized(customization: &[u8]) -> Result<Self, Self::Error>;
}

pub trait CustomizedInit: TryCustomizedInit<Error = Infallible> {
    fn new_customized(customization: &[u8]) -> Self {
        let Ok(res) = self.try_new_customized(customization);
        res
    }
}

impl<T: TryCustomizedInit<Error = Infallible> CustomizedInit for T {}
```

But it would be a breaking change, so it has to be deferred until a future breaking release.

tarcieri (approved):

PR #2406, opened 2026-05-09
crypto-common: add `MaybeBox`

tarcieri:
Adds a type which is always available, but when the newly added `alloc` feature is enabled, provides opportunistic heap allocation with `Box`, falling back on stack allocation if it is not.

Originally added to the `ml-dsa` crate in RustCrypto/signatures#1320 to address the large size of post-quantum keys and signatures while still retaining `no_alloc` support.

However, it is generally useful anywhere we work with secret values to e.g. prevent moves from making copies of them on the stack, even if they aren't excessively large.

### DIFF
--- a/.github/workflows/crypto-common.yml
+++ b/.github/workflows/crypto-common.yml
@@ -41,6 +41,10 @@ jobs:
           toolchain: ${{ matrix.rust }}
           targets: ${{ matrix.target }}
       - run: cargo build --target ${{ matrix.target }}
+      - run: cargo build --target ${{ matrix.target }} --features alloc
+      - run: cargo build --target ${{ matrix.target }} --features rand_core
+      - run: cargo build --target ${{ matrix.target }} --features zeroize
+      - run: cargo build --target ${{ matrix.target }} --features alloc,rand_core,zeroize
 
   minimal-versions:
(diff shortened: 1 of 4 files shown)


tarcieri:
Note: trying to find somewhere to extract this. If `crypto-common` doesn't make sense I can start with `module-lattice` I guess, as my main goal is using it with `ml-kem` in addition to `ml-dsa`.

Also not wild about the name, but I can't think of a better one.

newpavlov (commented):
The type itself looks fine to me, but I am not sure it's worth to keep it as part of `crypto-common`. It would've been a bit better if we already had the `alloc` feature. So I would prefer to keep it elsewhere for now.

>Also not wild about the name, but I can't think of a better one.

How about `CondBox` (i.e. conditionally `Box`)?

tarcieri:
I can start with `module-lattice` for now, though there are definitely many more places I think this could be used. Pretty much anywhere we have secret key material, I think it would make sense.

I do agree adding an `alloc` feature to `crypto-common` just for this is a little annoying, though since it's off-by-default I thought that also minimized the impact of the change.

newpavlov:
We also could put it into a separate utils crate.

Issue #2413, opened 2026-05-12
elliptic-curve: fix minimal versions


newpavlov:
See #2412

newpavlov:
Fixed by #2413.

PR #2414, opened 2026-05-12
elliptic-curve: fix minimal-versions check

tarcieri:
The issue as reported in #2413 is buried deep in the transitive dependencies of `bitvec` which is in turn a transitive dependency of `ff`, so it's a bit annoying to debug.

It turns out to be `wyz` does not correctly specify the version of `tap` it needs: it needs 1.0.1 but asks for 1 (I feel like I've probably encountered this before)

Sidebar: alternatively we could drop the `bits` feature. I'm not sure if anyone's actually using it and I suppose I could look into that, but `bitvec` development seems stalled at this point and asking about it on Zulip's #t-opsem channel did not inspire confidence for cryptographic applications.


newpavlov:
IMO it's probably worth to trim the `elliptic-curve`'s dependency tree in the next-next breaking release (i.e. v0.15).

PR #2417, opened 2026-05-12
elliptic-curve: remove `bits` feature

tarcieri:
I was not able to find any users of it searching GitHub.

This feature is backed by `bitvec` which both has a high number of transitive dependencies which have been minorly problematic (#2414), but also my understanding based on Zulip discussions on #t-opsem is that `bitvec` is not sound under either stacked or tree borrows, which makes it a bit worrisome for use in cryptographic applications.

We can always add the feature back if it turns out there are users, but if not it seems like it's not worth the maintenance burden.

### DIFF
--- a/.github/workflows/elliptic-curve.yml
+++ b/.github/workflows/elliptic-curve.yml
@@ -46,7 +46,6 @@ jobs:
       - run: cargo build --target ${{ matrix.target }} --release --no-default-features
       - run: cargo build --target ${{ matrix.target }} --release --no-default-features --features alloc
       - run: cargo build --target ${{ matrix.target }} --release --no-default-features --features arithmetic
-      - run: cargo build --target ${{ matrix.target }} --release --no-default-features --features bits
       - run: cargo build --target ${{ matrix.target }} --release --no-default-features --features critical-section
       - run: cargo build --target ${{ matrix.target }} --release --no-default-features --features dev
       - run: cargo build --target ${{ matrix.target }} --release --no-default-features --features digest
@@ -104,6 +103,6 @@ jobs:
       - uses: actions/checkout@v6
       - uses: dtolnay/rust-toolchain@stable
         with:
-          toolchain: nightly-2026-02-11 # pinned due to rust-lang/miri#4855
+          toolchain: nightly-2026-05-01 # pinned due to ensure determinism
       - run: rustup component add miri && cargo miri setup
       - run: cargo miri test --all-features
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,17 +3,6 @@
 version = 4
 
 [[package]]
-name = "addchain"
-version = "0.2.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2e33f6a175ec6a9e0aca777567f9ff7c3deefc255660df887e7fa3585e9801d8"
-dependencies = [
- "num-bigint",
- "num-integer",
- "num-traits",
-]
-
-[[package]]
 name = "aead"
 version = "0.6.0-rc.10"
 dependencies = [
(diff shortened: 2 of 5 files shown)


kayabaNerve:
> I was not able to find any users of it searching GitHub.

I don't believe I used `elliptic-curve/bits` but effectively all of my work used `k256/bits` because I generally assume `F: PrimeFieldBits` as a precondition, to enable using my [`multiexp` crate](https://crates.io/crates/multiexp). This is a precursor to _most_ of my cryptography regarding proofs due to how multiexponentations are almost always a good idea, and it avoids the issue with `group::WnafScalar` assuming LE representations.

I don't love the `PrimeFieldBits` trait, especially because it returns a `bool` and not a `Choice` when iterated over, and I don't need/care for `bitvec`. Some solution for `to_le_bits` is used and necessary though.

I hope the RFC to `ff` is merged, implemented, and released soon...

tarcieri:
@kayabaNerve you just want an iterator over `Choice`?

I did write `bitref` as an alternative to `bitvec`, which passes under Tree Borrows, but still has the same soundness issue under Stacked Borrows: https://github.com/tarcieri/bitref

I could potentially consider a `ctutils` feature that enables an `Iterator<Item=Choice>`

tarcieri:
Also `elliptic-curve` natively supports multiscalar multiplication with Straus now. Pippenger would be nice to have if someone wants to contribute it either here or to `group`.

I am trying to upstream it: https://github.com/zkcrypto/group/pull/85

kayabaNerve:
For `multiexp`, I'd like a `to_le_bits` which yields `Choice` as the representation of the bits.

I do have uses beyond `multiexp` though. I have a project which maps scalars to integers and vice-versa, and that requires endian-aware encoding (such as `to_le_bits` offered if you then map it to bytes yourself).

tarcieri:
FYI, I extracted a `wnaf` crate (which implements Straus) if you'd like to contribute Pippenger (if not I may attempt to adapt the implementation from `curve25519-dalek`):

https://github.com/RustCrypto/elliptic-curves/tree/master/wnaf/

The request for bits makes me want to add a `ctutils` integration with `bitref` as I mentioned earlier. Though there is the unfortunate Stacked Borrows soundness problem, heh.

kayabaNerve:
I'd be quite interested in doing so. My `multiexp` crate does have an implementation of Pippenger in it, and I'd be happy to have less under our umbrella and more under the ideal generic group which is RustCrypto. I also have a variety of misc contributions to `crypto-bigint` in my backlog...

But for `multiexp`, one needs `to_le_bits` or similar (actually `to_be_bits` is arguably optimal), and in other work of mine, I need to map between integers and scalars in a prime field (so $\mathbb{F}_p$ -> $\mathbb{Z}$ and vice-versa). All of this basically resolves as needing endian-aware encoding, whether on a bit level or a byte level.

For _how_ that encoding occurs, I'd prefer a sound solution. `to_le_repr` where `Repr` is left as-is would still allow iterating over the bits with just a bit of indexing, and any reliance on `bitvec` or similar is completely outside my interest.

As for the migration moving forward, I do need _some_ universal trait I can rely on, or I'll have to define my own and begin fragmenting the ecosystem. `PrimeFieldBits` _was_ that trait. Reviewing how `wnaf` handles it, in case RustCrypto has already begun such a trait, I see `wnaf` currently just takes the flip side of the same flawed coin `group` itself has: https://docs.rs/wnaf/0.14.0-pre.0/src/wnaf/lib.rs.html#356 :/ (in the pre-release, to be fair)

tarcieri:
@kayabaNerve I've been trying really hard to solve that particular problem, just haven't had a lot of uptake on the @zkcrypto side: https://github.com/zkcrypto/rfcs/pull/4

The `LimbBuffer` type in `wnaf` (that came via `group`) could potentially be made into a better bit-based iterator and even exposed in the public API.

PR #2418, opened 2026-05-18
[WIP] aead: add `PrefixNonceAead`

tarcieri:
This is another attempt at #1666 (see also #1818).

Where #1818 tried to add some additional trait methods, this instead opts for a generic newtype which provides type safety around the use of prefix messages (i.e. avoids confusion around accidentally mixing up prefixed and non-prefixed messages by calling the wrong methods) and in general keeps the method space unambiguous so we don't need to worry about qualifying every method name with something to indicate it's working on nonce-prefixed messages.

TODO: tests, docs, but I'm posting this early to get feedback on the general direction

### DIFF
--- a/aead/Cargo.toml
+++ b/aead/Cargo.toml
@@ -28,7 +28,7 @@ bytes = { version = "1.11.1", optional = true, default-features = false }
 default = ["rand_core"]
 alloc = []
 dev = ["blobby", "alloc"]
-getrandom = ["common/getrandom"]
+getrandom = ["common/getrandom", "rand_core"]
 rand_core = ["common/rand_core"]
 
 [lints]
--- a/aead/src/lib.rs
+++ b/aead/src/lib.rs
@@ -8,15 +8,18 @@
 #![forbid(unsafe_code)]
 
 #[cfg(feature = "alloc")]
+#[macro_use]
 extern crate alloc;
 
 #[cfg(feature = "dev")]
 pub mod dev;
+mod prefix_nonce;
 
 pub use common::{
     self, Key, KeyInit, KeySizeUser,
     array::{self, typenum::consts},
 };
+pub use prefix_nonce::PrefixNonceAead;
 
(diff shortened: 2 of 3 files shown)


tarcieri:
I suppose this is always assuming a postfix nonce, I will need to further compose it to abstract over prefix vs postfix nonces

newpavlov:
It's an interesting approach, but I am not sure about its consistence in the context of the overall API. I would expect to also have `PrefixTagAead` or such for tag-only ciphertexts. I guess one advantage of this method is that it allow to introduce `encrypt_with_random_nonce` methods on such struct, but I think we could do something similar with generic parameters as well.

As I wrote previously, I prefer to introduce a trait (e.g. `AeadFormat`) and implement different format descriptor types like `NonceTagCt`, `TagCt`, `CtTag`, etc. For the default tag-only format we could introduce a special type with behavior dependent on the `IS_PREFIX` constant. I will try to prepare a draft for it later.

tarcieri:
> I am not sure about its consistence in the context of the overall API

I will say implementing this feels harder than it should be.

>  I guess one advantage of this method is that it allow to introduce encrypt_with_random_nonce methods on such struct

That's the whole point: abstracting away the nonce management. That's why all the methods that generate a random nonce are front-loaded in the API.

Though #1666 suggested having a marker trait for AEADs considered safe to use with random nonces, which this PR is currently lacking.

This sort of API is fundamentally different in that all the methods except the one that accepts an explicit nonce don't have a nonce because it's randomly generated instead.

> but I think we could do something similar with generic parameters as well.

How? There's a pretty fundamental disparity between e.g. using the system's ambient RNG as the `getrandom`-based methods in this PR are doing, randomly generating a nonce with an RNG param, and accepting an explicit nonce value, all of which are supported by this PR.

newpavlov:
>How?

By using an additional marker trait (or an associated constant set to `true`) for formats which include nonce in generated ciphertexts.

PR #2420, opened 2026-05-19
crypto-common: add `SetIvState` trait

newpavlov:
See discussion in https://github.com/RustCrypto/block-modes/pull/91


newpavlov:
Note that I plan to re-export it from `cipher` in a follow-up PR after `crypto-common` v0.2.2 is released.

tarcieri (approved):

Issue #2425, opened 2026-05-23
aead v0.6.0 release tracking issue


tarcieri:
This is an issue for tracking the next stable release of the `aead` crate, v0.6.0

newpavlov:
One open question is whether we want to support nonce sizes variable at runtime.  Some algorithms (e.g. GCM) support variable nonce sizes and currently we expose it at compile time using generic parameters. We could add methods which accept nonces as byte slices, but it's unclear whether it's worth to do.

kkoyung:
One possible downstream benefit of runtime-sized nonces is standards compatibility for browser engines.

Servo uses the `aes-gcm` crate for its WebCrypto AES-GCM implementation, and the WebCrypto specification allows callers to provide nonces of varying lengths. Under the current `aead` trait design, Servo only supports a few fixed, commonly used nonce lengths.

Disclaimer: I'm one of the maintainers of Servo. Just wanted to share this as a downstream use case.

rozbb:
Sounds like a valuable feature, though maybe hazmat-gated since short nonces can easily cause a catastrophic failure

tarcieri:
It seems like there are more or less two ways to implement dynamically-sized nonces: a generic parameter that can abstract over `Array<u8, _>` vs `[u8]`, or a parallel trait with `[u8]` in the place of the current `Array`.

We can debate the virtues of one over the other, however I'll note the latter unblocks this feature from preventing a release, since it can be done in a purely additive manner, and I think it's a bit late for major breaking changes at this point.

I say we try the parallel trait approach, and if it seems like a generic approach would work out better, we can revisit it in the next breaking release. Like @rozbb said perhaps it should be hazmat, and I think it can present a fairly low-level API aimed for "power user" use cases like Servo, as opposed to having to duplicate every single method of e.g. `AeadInOut`.

--

I'll also say I otherwise don't see any other release blockers. The other features I'm interested in pursuing can also be done in an additive manner.

newpavlov:
>I'll also say I otherwise don't see any other release blockers.]

[...]

PR #2427, opened 2026-05-25
aead: rework traits API

newpavlov:
TODO:
- Update changelog.
- Update docs.
- Test new methods.
- Migrate implementation crates.

Unresolved questions:
- Solution for potential incorrect reordering of arguments (`nonce`, `aad`, `plaintext`/`ciphertext`) with the same type (`&[u8]`).
- Should we handle potential fallibility in `allocate`, `extend`, and `truncate` closures?


aead/src/lib.rs
```diff
@@ -409,96 +150,35 @@ impl<T: AeadInOut> AeadInPlace for T {
...
 }
 
-impl<'msg> From<&'msg [u8]> for Payload<'msg, '_> {
-    fn from(msg: &'msg [u8]) -> Self {
-        Self { msg, aad: b"" }
-    }
+/// Trait which carries the default tag position.
+pub trait AeadTagPosition: AeadCore {
+    /// The AEAD tag position.
+    const TAG_POSITION: TagPosition;
```
newpavlov: `TAG_POSITION` is moved into a separate trait to make `AeadCore` `dyn`-compatible (after pinning `Tag/NonceSize`). Plus it has nothing to do with `AeadCore` methods, so it probably makes sense to keep it separately.

aead/src/high_aead.rs
```diff
@@ -0,0 +1,151 @@
+use crate::{
+    AeadCore, AeadTagPosition, Error, InOutBuf, Nonce, Result,
+    TagPosition::{Postfix, Prefix},
+};
+use alloc::vec::Vec;
+use common::array::typenum::Unsigned;
+
+/// High-level functionality of Authenticated Encryption with Associated Data (AEAD) algorithms.
+pub trait Aead: AeadCore + AeadTagPosition {
```
newpavlov: Note that `Aead` in its current form is not `dyn`-compatible. See this [comment](https://github.com/RustCrypto/traits/issues/2425#issuecomment-4535584073) for a potential alternative.

newpavlov:
To help with migration from `Buffer` I consider adding something like this to `AeadCore`:
```rust
fn encrypt_into<B: AsMut<[u8]>>(
    &self,
    nonce: &[u8]/Nonce<Self>, 
    aad: &[u8],
    pt: &[u8],
    // Allocate new buffer with the given size for writing ciphertext and tag
    allocate: impl FnOnce(usize) -> B,
) -> B { ... }

// will use `copy_within` for prefix modes
fn encrypt_within<B: AsMut<[u8]>>(
    &self,
    nonce: &[u8]/Nonce<Self>,
    aad: &[u8],
    buf: &mut B,
    // Extend `buf` to the given size
    extend: impl FnOnce(&mut B, usize),
) { ... }    
```
They also could be added to `Aead` as well with `where Self: Sized` bound to preserve `dyn`-compatibility.

aead/src/tag_pos.rs
```diff
@@ -0,0 +1,226 @@
...
+            self.decrypt_inout_detached(nonce, aad, buf, tag)
+        })
+    }
+
+    #[inline]
+    fn decrypt_into_with_var_nonce<B: AsMut<[u8]>>(
+        &self,
+        nonce: &[u8],
+        aad: &[u8],
+        ciphertext: &[u8],
```
newpavlov: Using 3 arguments with the same type may cause bugs, so it may be worth to use `Payload` here as well.

aead/src/high_aead.rs
```diff
@@ -0,0 +1,98 @@
+use crate::{AeadCore, AeadWithTag, Result};
```
tarcieri: I'm not sure the `high_aead` module name really makes sense. "High" is not a part of the trait name.

aead/src/lib.rs
```diff
@@ -245,260 +121,132 @@ pub trait AeadInOut: AeadCore {
...
+    /// returning the authentication tag.
     ///
-    /// The buffer must have sufficient capacity to store the ciphertext
-    /// message, which will always be larger than the original plaintext.
-    /// The exact size needed is cipher-dependent, but generally includes
-    /// the size of an authentication tag.
+    /// # Warning
+    /// Some algorithms support very short nonces. Users should exercise extreme caution
+    /// while using this method since incorrect handling of nonces may defeat security
+    /// provided by the algorithm.
```
tarcieri: This really feels like hazmat functionality...

aead/src/tag_pos.rs
```diff
@@ -0,0 +1,226 @@
...
+        $truncate($buf, ct_len);
+        assert_eq!($buf.as_mut().len(), ct_len);
+
+        Ok(())
+    }};
+}
+
+/// Trait implemented for AEAD modes which specify tag position in ciphertext.
+#[allow(missing_docs, clippy::missing_errors_doc)] // TODO: fix
+pub trait AeadWithTag: AeadCore {
```
tarcieri: AEADs definitionally always have a tag, so `AeadWithTag` seems like an odd choice of name. The previous `AeadTagPosition` name made more sense as a trait to associate a tag position with an `Aead` algorithm, though I see you also put the methods here too, which again also seems weird as all AEADs have tags

tarcieri:
Overall this doesn't feel like an improvement to me. Several things have changed but I don't think they've changed in ways that make more sense.

aead/src/lib.rs
newpavlov: Sure, but we already gate short nonce support in the implementation crates, so gating the methods on top of it looks redundant. Plus it could be used as a convenience API to eliminate redundant `TryInto` checks, while the intentionally long name plays role of "salt" to discourage their use.

aead/src/high_aead.rs
newpavlov: I also do not like this name, but couldn't think of anything better considering that `aead` would clash with the crate name. I guess we could just move it to `lib.rs`.

aead/src/tag_pos.rs
newpavlov: As I wrote in the OP, we could merge it to `AeadCore` if we are fine with giving up `dyn`-compatibility.

tarcieri: I would prefer to have a non-`dyn`-compatible trait that knows all of this information at the type level (`AeadCore` is fine for that), and a separate trait that is `dyn`-compatible, as you originally proposed

newpavlov:
IMO variable nonce support, merging `AeadCore` and `AeadInOut`, significantly reducing number of crate features, and making `Aead` properly `dyn`-compatible (and potentially implementable by IO-based tokens) make a fair amount of sense.

aead/src/lib.rs
tarcieri: Those `TryInto` checks ensure a user is using a sufficiently-sized nonce. If we have APIs accept variable-sized nonces, we lose that defense against misuse. It's really power user functionality for use cases like Servo.

I'll also note that generally for AES-GCM, we want to steer users to using 96-bit uses which the current API enforces at the type level, as any other size including 128-bits involves a secondary GHASH invocation which introduces nonce collision potential where none need exist, and abandoning type safety here loses that, introducing misuse risk.

newpavlov: In implementation crates I consider introducing separate "variable" types which would be gated on `hazmat`, so for the "default" types shorter nonces will be simply rejected.

newpavlov: We also could introduce a totally separate trait for variable methods, but I am not sure about implementation relations between the resulting traits.

tarcieri: Yes, I think something like that would be safer, either a separate `hazmat` trait or failing that a `DynAead` trait that carries a large warning about potential nonce size misuses and to read the crate documentation carefully when selecting a nonce size. Perhaps such a trait could also dynamically generate a random nonce with recommended size, particularly if it were gated on `alloc`, although that's perhaps a separate discussion.

A separate type for dynamic use cases as you proposed would also be cleaner for something like `aes-gcm` where the current types carry nonce size as a generic parameter, but the typed wrapper can be composed in terms of a dynamic core.

aead/src/high_aead.rs
newpavlov: Renamed the module to `aead`. It seems to work fine.

aead/src/utils.rs
```diff
@@ -0,0 +1,130 @@
...
+    let (ct_dst, tag_dst) = match C::TAG_POSITION {
+        Postfix => ct_tag.as_mut().split_at_mut(plaintext.len()),
+        Prefix => {
+            let (tag_dst, ct_dst) = ct_tag.as_mut().split_at_mut(tag_len);
+            (ct_dst, tag_dst)
+        }
+    };
+
+    let buf = InOutBuf::new(plaintext, ct_dst)
+        .expect("`plaintext` and `ct_dst` always have the same length");
```
newpavlov: Replace it with `.map_err(|_| Error)` to remove `#[allow(clippy::unwrap_in_result)]`?

aead/src/aead.rs
```diff
@@ -0,0 +1,59 @@
...
+use alloc::vec::Vec;
+
+/// High-level functionality of Authenticated Encryption with Associated Data (AEAD) algorithms.
+pub trait Aead {
+    /// Encrypt the given plaintext payload, and return the resulting
+    /// ciphertext as a vector of bytes.
+    ///
+    /// # Errors
+    /// AEAD algorithm implementations may return an error if the plaintext or AAD are too long.
+    fn encrypt_into_vec(&self, nonce: &[u8], aad: &[u8], plaintext: &[u8]) -> Result<Vec<u8>>;
```
newpavlov, line 11: I removed the `Payload` type for now. I think we either need a more general solution which would help with other methods as well  (e.g. `Nonce`, `Aad`, `Plaintext`, etc. wrappers), or we could just tolerate potential incorrect reorderings of the arguments, which are hopefully somewhat unlikely in modern conditions (with IDEs and stuff).

aead/src/lib.rs
```diff
@@ -105,400 +86,286 @@ pub type Nonce<A> = Array<u8, <A as AeadCore>::NonceSize>;
...
+    #[inline]
+    fn encrypt_into<B: AsMut<[u8]>>(
         &self,
         nonce: &Nonce<Self>,
-        associated_data: &[u8],
-        buffer: &mut dyn Buffer,
-    ) -> Result<()>;
+        aad: &[u8],
+        plaintext: &[u8],
+        allocate: impl FnOnce(usize) -> B,
```
newpavlov, line 182: Should we handle potential fallibility here and in `extend`/`truncate`?

aead/src/lib.rs
newpavlov: Unfortunately, it looks like moving the variable methods into a separate trait can not be easily done without specialization.

newpavlov: Or we could do something like this:
```rust
pub trait AeadCore { ... }

// without blanket impl for T: AeadCore
pub trait VariableAead: AeadCore {
   // methods with blanket impls in terms of `AeadCore` methods
}

pub trait Aead { ... }

impl<T: VariableAead> Aead for T { ... }

// implementation crate
impl AeadCore for SomeAead { ... }
impl VariableAead for SomeAead {}
```
In other words, we would need to include somewhat annoying empty impl blocks of `VariableAead` for types which do not support variable nonces.

aead/src/lib.rs
```diff
@@ -105,400 +89,289 @@ pub type Nonce<A> = Array<u8, <A as AeadCore>::NonceSize>;
...
+    /// If `extend` does not extend the buffer to the specified length in bytes.
+    #[inline]
+    fn encrypt_within<B: AsMut<[u8]>>(
         &self,
         nonce: &Nonce<Self>,
-        associated_data: &[u8],
-        buffer: &mut dyn Buffer,
+        aad: &[u8],
+        buf: &mut B,
+        extend: impl FnOnce(&mut B, usize),
```
newpavlov, line 268: I thought about using a pass-through API:
```rust
fn encrypt_within<B: AsMut<[u8]>>(
    &self,
    nonce: &Nonce<Self>,
    aad: &[u8],
    buf: B,
    extend: impl FnOnce(&mut B, usize),
) -> Result<B> { ... }
```
But I decided against it since it may be inefficient for some buffer types (e.g. `ArrayVec`).

aead/src/variable.rs
```diff
@@ -0,0 +1,261 @@
...
+
+/// Functionality of Authenticated Encryption with Associated Data (AEAD) algorithms
+/// with variable nonce and tag size support.
+///
+/// <div class="warning">
+/// Some algorithms support very short nonce and tag sizes. Users should exercise extreme caution
+/// while using this trait since incorrect handling of nonces and tags may defeat security
+/// provided by the algorithm.
+/// </div>
+pub trait VariableAead: AeadCore {
```
tarcieri: This seems like a purely additive change which could be added and discussed in its own PR

tarcieri: I also don't understand the supertrait bounds here. It seems like it should be the most flexible, fundamental trait which isn't constrained by knowledge of particular sizes, and support different sizes at runtime.

tarcieri:
I really think, at the very least, this should be broken down into multiple, smaller PRs.

You are trying to rush through a large number of unrelated changes in a single PR shortly before a release.

I would like to see at least two PRs:

1. A purely additive PR for `VariableAead`
2. At least one separate PR for all of the assorted other unrelated changes, which can build on top of the `VariableAead` PR

Trying to make this many changes so very late in the crate's development lifecycle right before a stable release is highly problematic, IMO, and having them all in a single PR means discussion is unfocused as we're jumping around between all of the different changes rather than focusing on a set of them in isolation.

newpavlov:
Ok, I will remove `VariableAead` and will add it in a separate PR. But the `AeadCore` and `Aead` changes are interconnected and I don't think it makes sense to split them, especially considering that `Aead` is used by the `dev` module. The resulting code is pretty small (just 12 methods), so splitting it further would be just a confusing busywork.

tarcieri:
Can we start with `VariableAead`? It was the whole impetus for making any last minute changes whatsoever.

aead/src/variable.rs
newpavlov: As I wrote above, I am not sure about what would be the best implementation relationships between the traits, so I am open to alternatives. I got to the current solution from the following factors:

1) `Aead` should support variable nonces, so it's blanket implementation should be based on `VariableAead`.
2) We want to automatically (or at least easily) implement `VariableAead` if type implements `AeadCore`.
3) We do not have specialization in stable Rust.
4) We want to guide users towards "default" sizes even if they rely on `VariableAead` and, as discussed above, we are likely to have separate "variable" types in implementation crates.

We could make `VariableAead`  more "fundamental" than `AeadCore`, but it would look strange from the naming perspective ("core" becomes a mid-level layer) and it would mean that implementing both `VariableAead` and `AeadCore` would be slightly more annoying.

tarcieri: `VariableAead` is more fundamental in that it can do things an `AeadCore` cannot, namely implement multiple sizes simultaneously.

An `AeadCore` bound means a type that impls `VariableAead` will have an associated set of fixed sizes for things like nonces and tags, which isn't a property a type which supports multiple sizes at runtime can have.

newpavlov:
I would **strongly** prefer to start with the `AeadCore` and `Aead` changes and to build `VaraibleAead` on top of it.

I will ignore your "last minute" remarks, otherwise the discussion could get pointlessly heated _again_ and we will get nowhere.

tarcieri:
> I would strongly prefer to start with the AeadCore and Aead changes and to build VaraibleAead on top of it.

Where are the actual interdependencies which make that preferable?

aead/src/variable.rs
newpavlov: We also could view `VariableAead` as an "extension" of `AeadCore` for additional support of nonce sizes at runtime, but I understand your position. I will try to play with it later.

Am I correct, that you prefer for `AeadCore` implementers to not implement `VariableAead`? And that `Aead` should be implemented in terms of `AeadCore` and not `VariableAead`? Note that we can not have two blanket impls without support of mutually exclusive traits in the language.

newpavlov:
The structure of methods. `VariableAead` effectively mirrors `AeadCore` by replacing compile-time sizes with run-time equivalents, so it helps to finalize it first without going through a bunch of annoying iterations.

aead/src/variable.rs
tarcieri: I think in cases where there is actually a legitimate need to support multiple nonces and tag sizes at runtime, as with AES-GCM, that it can have a `VariableAead` core which is untyped in regard to nonce and tag sizes, and separately have well-typed wrappers which are newtypes composed in terms of the `VariableAead` core and delegate to that implementation, rather than using a blanket impl.

The only way such a blanket impl could even work, IMO, is if you kept `AeadCore` as-is where it only provides the size information, and the blanket impl could use that size information to provide a well-typed API. But I think that makes less sense than just having those types impl the relevant traits directly with a blanket impl that goes the other direction, providing `VariableAead` for `AeadCore` implementations.

tarcieri:
> I will ignore your "last minute" remarks, otherwise the discussion could get pointlessly heated again and we will get nowhere.

I'm just going to close this PR. This is absolutely the last minute and it's ridiculous to claim it's not. I will contact you on PM. 

There are [many people](https://github.com/RustCrypto/traits/issues/2425) including myself who have been waiting on this release. The last stable release was over three years ago.

I do not think it makes sense to make massive numbers of changes in an unfocused PR called "rework traits" at the last minute. That is not a responsible way to develop software.

If you want to make such last minute cross-cutting breaking changes, please open a *focused* PR that can explain what was wrong with the old API, proposes a change, and can explain how that change is an improvement over the original.

Please do it quickly, people are waiting.

newpavlov:
>If you want to make such last minute cross-cutting breaking changes, please open a focused PR that can explain what was wrong with the old API, proposes a change, and can explain how that change is an improvement over the original.

IMO calling this PR "unfocused" is frankly ridiculous. I already [wrote](https://github.com/RustCrypto/traits/pull/2427#issuecomment-4587226253) why I consider this PR improvement. And stop already with the "last minute" card, I tried to introduce changes to the crate a number of times, but you stubbornly refused them using the flimsy "late stage of development" justification despite me incorporating almost all changes requested by you. Sorry, but I do not want to jump through any more unnecessary hoops. I already wasted a lot of time working on the PRs.

As I wrote in PM, feel free to release the crate in any form you deem appropriate. But I will be introducing this and other changes (e.g. support for AE and incremental AEAD) in a future release (probably shortly after) and I will not wait on your feedback with them.

PR #2438, opened 2026-06-06
elliptic-curve: add `WnafScalar::from_le_bytes`

42Pupusas:
Adds a constructor that builds a `WnafScalar` directly from raw little-endian scalar bytes, bypassing the full-width field representation used by `WnafScalar::new`.

This is useful when the scalar has already been decomposed into a value smaller than the field modulus — e.g. the ~128-bit half-scalars produced by a GLV endomorphism decomposition. Passing the shorter byte slice yields a shorter w-NAF, which halves the number of doublings in the evaluation loop relative to feeding the full-width scalar.

`bytes` is interpreted as a little-endian unsigned integer (trailing zero bytes may be omitted); over-length input returns `None`. The existing `wnaf_form` already zero-extends short inputs and emits a trailing carry digit, so values near the top of the byte range remain exact.

### DIFF
--- a/elliptic-curve/src/arithmetic/wnaf.rs
+++ b/elliptic-curve/src/arithmetic/wnaf.rs
@@ -458,6 +458,38 @@ impl<F: PrimeField, const WINDOW_SIZE: usize> WnafScalar<F, WINDOW_SIZE> {
             field: PhantomData,
         }
     }
+
+    /// Computes the w-NAF representation directly from raw little-endian bytes.
+    ///
+    /// `bytes` is interpreted as a little-endian unsigned integer (trailing zero bytes may be
+    /// omitted), and the resulting [`WnafScalar`] evaluates to that integer times the base.
+    /// Because the number of w-NAF digits — and therefore the number of doublings in the
+    /// evaluation loop — is proportional to `bytes.len() * 8`, passing a slice shorter than the
+    /// field's canonical representation produces a faster scalar.
+    ///
+    /// This is intended for callers that have already decomposed a scalar into a value provably
+    /// smaller than the field modulus, e.g. the ~128-bit half-scalars produced by a GLV
+    /// endomorphism decomposition.
+    ///
+    /// Returns `None` if `bytes` is longer than the field's canonical representation. Note that
+    /// no reduction modulo the field order is performed: the caller is responsible for ensuring
+    /// the encoded integer is the intended representative.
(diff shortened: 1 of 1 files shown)


elliptic-curve/src/arithmetic/wnaf.rs
```diff
@@ -458,6 +458,38 @@ impl<F: PrimeField, const WINDOW_SIZE: usize> WnafScalar<F, WINDOW_SIZE> {
...
+    /// evaluation loop — is proportional to `bytes.len() * 8`, passing a slice shorter than the
+    /// field's canonical representation produces a faster scalar.
+    ///
+    /// This is intended for callers that have already decomposed a scalar into a value provably
+    /// smaller than the field modulus, e.g. the ~128-bit half-scalars produced by a GLV
+    /// endomorphism decomposition.
+    ///
+    /// Returns `None` if `bytes` is longer than the field's canonical representation. Note that
+    /// no reduction modulo the field order is performed: the caller is responsible for ensuring
+    /// the encoded integer is the intended representative.
```
tarcieri: Hmm, that seems a little suspect re: scalars overflowing the order, or at least like it could use a more prominent warning, possibly a `<div class="warning">`

tarcieri: It also seems a little odd the function is fallible but doesn't fail on the most important case: overflowing the order.

Maybe make the function infallible and have it truncate the input to the byte length of the modulus if necessary, but still include a prominent warning.

42Pupusas: If this misuse of the API is enough of a worry, maybe the method should return a `Result` instead of `Option`, with each failure mode properly identified?

Would that be preferable to the warning tag?

42Pupusas: Ok, just answered my question, will rework to return Result

tarcieri: I'd be fine with either an infallible function that truncates but doesn't reduce or check for a canonical representative with a warning, intended as a low-level primitive, or one with a `Result` that checks the scalar is canonical

42Pupusas:
@tarcieri the `Result` returning path seems saner and with less footguns, and the extra checks had no measurable impact on the perf optimizations.

tarcieri (approved):
Nice! That works

PR #2450, opened 2026-06-17
Make get_random feature dependent on rand_core

### LINKED ISSUE #755: Doctests no longer build with default features
This was somehow introduced in #752 which bumped `aead` to v0.6.0-rc.4, but I am baffled as to how or what's even happening.

## Example reproduction

```console
$ cd aes-gcm
$ cargo test --doc
[...]
---- aes-gcm/src/lib.rs - (line 150) stdout ----
error[E0432]: unresolved import `aes_gcm::aead::Generate`
  --> aes-gcm/src/lib.rs:156:28
   |
8  |     aead::{Aead, AeadCore, Generate, Key, KeyInit},
   |                            ^^^^^^^^ no `Generate` in the root
   |
note: found an item that was configured out
```

## Notes

This occurs in several crates, any of which use this style of gating:

```rust
#![cfg_attr(feature = "getrandom", doc = "```")]
#![cfg_attr(not(feature = "getrandom"), doc = "```ignore")]
```

Curiously, if the first line is changed to:

```rust
#![cfg_attr(feature = "getrandom", doc = "```ignore")]
```

Then the test is successfully ignored, suggesting this gating somehow isn't working correctly:

```
running 2 tests
test aes-gcm/src/lib.rs - (line 150) ... ignored
test aes-gcm/src/lib.rs - (line 189) ... ignored
```

Note the other example, which depends on both the `arrayvec` and `getrandom` features, *is* successfully being ignored with logic like:

```rust
#![cfg_attr(all(feature = "getrandom", feature = "arrayvec"), doc = "```")]
#![cfg_attr(
    not(all(feature = "getrandom", feature = "arrayvec")),
    doc = "```ignore"
)]
```

Perhaps the weirdest part is nothing in #752 actually changed how this gating worked, it just started behaving differently when the `aead` crate was upgraded from v0.6.0-rc.3 to v0.6.0-rc.4.

eligrubb:
Fixes [AEADs Issue#755](https://github.com/RustCrypto/AEADs/issues/755). See also alternative solution, [AEADs PR#833](https://github.com/RustCrypto/AEADs/pull/833).

The `Generate` trait is used by rustdoc code demonstrating various AEADs `getrandom` feature, but the `Generate` trait is gated by the `rand_core` feature, which is not always enabled when `getrandom` is enabled. This is causing compilation errors for rustdoc tests that use `Generate` for Nonce generation.

### DIFF
--- a/aead/Cargo.toml
+++ b/aead/Cargo.toml
@@ -28,7 +28,7 @@ bytes = { version = "1.11.1", optional = true, default-features = false }
 default = ["rand_core"]
 alloc = []
 dev = ["blobby", "alloc"]
-getrandom = ["common/getrandom"]
+getrandom = ["common/getrandom, rand_core"]
 rand_core = ["common/rand_core"]
 
 [lints]


tarcieri:
Looks like you have a syntax error. I went ahead and took care of it in #2452

PR #2452, opened 2026-06-17
aead: have `getrandom` enable `rand_core`

tarcieri:
Closes #2450

### DIFF
--- a/aead/Cargo.toml
+++ b/aead/Cargo.toml
@@ -28,7 +28,7 @@ bytes = { version = "1.11.1", optional = true, default-features = false }
 default = ["rand_core"]
 alloc = []
 dev = ["blobby", "alloc"]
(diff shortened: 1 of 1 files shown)


newpavlov:
FYI this change was part of #2427.

tarcieri:
...which is a change unrelated to the other changes #2427 is making, which is exactly the problem with #2427.

Perhaps there are other additive changes in there which can be extracted into small, well-scoped PRs.

newpavlov:
And perhaps you could've noticed this and "other additive changes" by properly reviewing the PR, instead of being so dismissive. As I wrote many times, I consider the PR small enough and do not like the busy work of bikeshedding micro changes. IMO it only makes the history of changes unnecessarily confusing.

But alas, what's done is done.

PR #2457, opened 2026-06-21
elliptic-curve: replace `FieldBytesEncoding` with `FIELD_ENDIANNESS`

tarcieri:
Removes the `FieldBytesEncoding` trait and replaces it with:
- `Curve::FIELD_ENDIANNESS` which defaults to `ByteOrder::BigEndian`
- `field::{bytes_to_uint, uint_to_bytes}` generic free functions

This constant annoyingly duplicates the ones we have elsewhere, but without getting it upstream into `ff` (zkcrypto/ff#158) there is no single other convenient place to hang it but here.

Ideally this functionality could be phased out completely, but there are still places that need it for now (e.g. legacy `rfc6979`, ECDSA recovery).

In a future breaking release after the imminent one, we need to refactor everything so the base and scalar fields are treated completely separately. At that point, hopefully this all can and will need to go away, and we will be able to leverage an upstream endianness constant instead of having to stick (another) one in `elliptic-curve`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -141,9 +141,8 @@ dependencies = [
 
 [[package]]
 name = "crypto-bigint"
-version = "0.7.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "42a0d26b245348befa0c121944541476763dcc46ede886c88f9d12e1697d27c3"
+version = "0.7.4"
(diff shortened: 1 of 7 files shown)


tarcieri:
Have the curves migrated locally so I'll go ahead and merge

PR #2474, opened 2026-07-14
digest: Hash the length of the prefix in `new_with_prefix()`

fjarri:
It seems that it is a good idea to hash the prefix length in addition to prefix itself in `new_with_prefix()` to avoid collisions. Was that an omission, or is there something I'm not considering?

The length can be cast to `u128` instead of `u64` to make it extremely future-proof, adding a small performance cost. 

Not sure what is the policy regarding the output consistency, this change may be considered breaking.

### DIFF
--- a/digest/src/digest.rs
+++ b/digest/src/digest.rs
@@ -68,6 +68,7 @@ impl<D: FixedOutput + Default + Update + HashMarker> Digest for D {
         Self: Default + Sized,
     {
         let mut h = Self::default();
+        h.update(&(data.as_ref().len() as u64).to_be_bytes());
         h.update(data.as_ref());
         h
     }


newpavlov:
We clearly document what `new_with_prefix` does. If you want to include prefix length, then it's your responsibility.

fjarri:
The docs say

> Create new hasher instance which has processed the provided data.

I don't think it is in any way clear. I certainly thought it does hash the tag length until I actually looked in the code. Otherwise, what's the point of this function when one can just write `new().chain_update()` which is only longer by a few characters? If one implements `Digest` manually, in which cases wouldn't they implement it this way? What is the contract here?

tarcieri:
@fjarri I think you want a construction like TupleHash or SequenceHash: https://github.com/C2SP/C2SP/pull/246

The `Digest` trait is just a facade over a raw IUF API. As soon as you talk about it injecting things like length prefixes you're talking about a protocol, not just a raw digest function. At that point you have to decide how big the length headers are, whether they're big endian, little endian, or some textual encoding, or something else entirely etc.

It arguably doesn't wind up saving that many characters over `new().chain_update()` but actually winds up being used quite a bit any time there's some sort of fixed prefix used in protocols.

Issue #2478, opened 2026-07-31
Proposal: a multi-buffer (multi-message) hashing trait for parallel independent digests


npmccallum:
## The core primitive

We'd like to add a first-class abstraction for **hashing N independent messages at once**, so an implementation can place one message per SIMD lane and run the compression function across all lanes simultaneously:

> Given N independent byte messages, produce their N digests in a single call.

This is the classic "multi-buffer" SHA construction (Gueron–Krasnov, Intel isa-l `sha512_mb`/`sha256_mb`). It exploits the fact that the round function has no cross-*message* dependency, even though it's fully serial within one message — so it's a different axis of parallelism from everything RustCrypto has today.

## Why it's missing today (prior art)

As far as we can tell this capability doesn't exist anywhere in RustCrypto, and the existing parallelism is all a different axis:

- **`ParBlocksSizeUser` / `ParBlocks`** (`crypto-common`, used by `cipher`) is the closest *structural* match — "process LANES units at once + scalar tail," via `encrypt_par_blocks`/`encrypt_tail_blocks` over `inout::InOutBuf`. But it means *multiple blocks of one stream*, not independent messages.
- **BLAKE2bp/sp, BLAKE3 tree mode, the sha2 AVX2 backend** (#327) are all *single-message* internal SIMD.
- The hashing core API (`digest/src/block_api.rs`) has **no** par-blocks concept; `update_blocks` is serial over one message.

So this would be a genuinely new capability on the hashing side, ideally borrowing the cipher-side conventions.

## Motivating use case

Parallel Merkle-tree **leaf** hashing: the leaves are independent, equal-length messages — the ideal multi-buffer workload. (Our concrete driver is a parallel, order-independent launch-measurement for AMD SEV-SNP confidential VMs, but the primitive is general: tree/dedup/storage/packet hashing all want it.) On AMD in particular, SHA-384/512 has no hardware engine, so multi-buffer is the only way to use the SIMD datapath — and it's substantial on the Zen 5 / Turin native 512-bit path.

## Design options

**A — High-level one-shot trait on top of `Digest`, runtime dispatch internal (sha2-style):**
```rust
pub trait MultiDigest: Digest {
    fn digest_many(msgs: &[&[u8]], out: &mut [Output<Self>]); // impl detects avx2/avx512, chunks internally
}
```

**B — Backend + driver split mirroring `cipher`'s `ParBlocks` (compile-time width):**
```rust
[...]

PR #2479, opened 2026-08-03
Add traits for hashing many messages at once

npmccallum:
Strawman trait API for #2478. Traits only; no algorithm implementation is included here.

These traits are backed by a complete multi-message implementation of SHA-224/256/384/512 (AVX2 4- and 8-lane, AVX-512 8- and 16-lane, and a portable fallback), tested byte-for-byte against scalar `Digest` and benchmarked.

On Zen 4c it reaches 4.2× single-stream throughput for SHA-512 and 1.6× for SHA-256, with throughput flat from 16 to 512 messages per batch. That implementation is not yet public and will follow separately; I'm glad to answer questions about it here in the meantime.

## What this adds

A hash reads one message and produces one digest. Inside, each block depends on the one before it, so a single message cannot be spread across SIMD lanes. Two *different* messages have no such dependency, so several can be hashed side by side — one per lane, the same instructions driving all of them. That is the capability this adds.

It is a different axis of parallelism from `ParBlocks`, which processes several blocks of one stream.

New feature-gated module `digest::multi` (`multi = []`, off by default), additive.

## Layering

Two unrelated things must be kept apart, and each layer hides one of them from the layer above:

1. **`MultiUpdateBackend` hides the implementation.** One backend is one concrete way of doing the work — an AVX2 routine, an AVX-512 routine, a portable one, or a future instruction set. Each has a lane count fixed by the hardware it targets. An algorithm supplies as many backends as it has implementations.

2. **`MultiUpdateCore` / `MultiFixedOutputCore` hide the lane count.** A caller asks to hash *N* messages, where *N* is a property of the workload rather than of the machine. The core selects a backend the current CPU supports, splits the batch across it, and handles whatever does not fill a complete set of lanes. Above this layer lane counts do not appear.

This mirrors the layering already used for ordinary hashing — a backend, then `UpdateCore` / `FixedOutputCore`. The backend layer follows the conventions used by `cipher`: `LanesSizeUser` is the analogue of `ParBlocksSizeUser`, and `LaneBlocks` of `ParBlocks`.

## Blocks in, a final tail at the end

`update_blocks` accepts whole blocks and nothing else; only `finalize_fixed_core` takes a sub-block tail. Absorbing whole blocks is what keeps the bulk of a message copy-free: blocks are read from wherever the caller already keeps them, and lanes are borrowed individually, so messages need not be contiguous with one another. Because finalize is the last call, it can take the loose tail and pad it without compromising that.

This matters most for structured messages — `tag ‖ payload`, as in Merkle leaves or any tagged record. A one-shot `multi_digest(msgs, out)` forces the caller to materialize the concatenation first, copying the payload. Here a shared prefix is absorbed once for all lanes with `update_blocks_shared`, and each payload is absorbed where it already lives. The only bytes copied are the one or two final padded blocks, which do not exist in the message and must be built regardless.

## The equal-length invariant is in the types

Every method takes fixed-size arrays: `MSGS` lanes of `BLOCKS` blocks or a `TAIL`-byte tail. So all lanes advance in step by construction — unequal lengths fail to compile rather than panicking, and there is no run-time length check anywhere. The API is infallible, like `Digest`.

```rust
fn update_blocks<const BLOCKS: usize>(&mut self, msgs: &[&[Block<Self>; BLOCKS]; MSGS]);
fn update_blocks_shared<const BLOCKS: usize>(&mut self, blocks: &[Block<Self>; BLOCKS]);
fn finalize_fixed_core<const TAIL: usize>(&mut self, tails: &[&[u8; TAIL]; MSGS], out: &mut [Output<Self>; MSGS]);
```

Output is written into caller-owned storage at the core layer, matching `FixedOutputCore`. The API uses plain slices and hybrid arrays, without `inout`.

## Dispatch

Backends are static types; the core selects among them at run time. A single binary stays correct on any CPU while using the widest available implementation, and adding an instruction set later requires a new backend and dispatch arm rather than a change to these traits.

## Open questions

- **Trait count.** `LanesSizeUser` exists to mirror `ParBlocksSizeUser` and has a single implementor-facing use; it could be folded into `MultiUpdateBackend`. Mirror the convention, or collapse it?
- **Splitting update from finalize.** `MultiUpdateCore` and `MultiFixedOutputCore` are split as `UpdateCore` and `FixedOutputCore` are, leaving room for a future `finalize_xof_core`. Worth it, or merge for now?
- **Naming.** Const parameters are `MSGS`, `BLOCKS`, and `TAIL`; same-input-to-every-lane variants use a `_shared` suffix. Placeholders for review.
- **A byte-buffering convenience.** Accepting arbitrary byte lengths means buffering a partial block whose length is only known at run time, which cannot be reconciled with the compile-time equal-length guarantee without either a runtime check (making the API fallible) or per-length monomorphization. It seems better to keep these traits infallible and leave a byte-oriented wrapper to an implementation crate. Agree, or should a wrapper live here?

### DIFF
--- a/digest/Cargo.toml
+++ b/digest/Cargo.toml
@@ -32,6 +32,7 @@ block-api = ["dep:block-buffer"] # Enable block API traits
 dev = ["blobby"]
 getrandom = ["common/getrandom", "rand_core"]
 mac = ["dep:ctutils"] # Enable MAC traits
+multi = [] # Enable multi-buffer (multi-message) hashing traits
 rand_core = ["common/rand_core"] # Enable random key generation methods
 oid = ["dep:const-oid"]
 zeroize = ["dep:zeroize", "block-buffer?/zeroize"]
--- a/digest/src/lib.rs
+++ b/digest/src/lib.rs
@@ -55,6 +55,8 @@ mod buffer_macros;
 mod digest;
 #[cfg(feature = "mac")]
 mod mac;
+#[cfg(feature = "multi")]
+pub mod multi;
 mod xof_fixed;
 
 #[cfg(feature = "block-api")]
--- a/digest/src/multi.rs
+++ b/digest/src/multi.rs
@@ -1,52 +1,192 @@
-//! Multi-buffer (multi-message) hashing.
+//! Hashing many independent messages at once.
 //!
-//! [`MultiDigest`] hashes many **independent messages of one compile-time length `N`** at
-//! once — one per SIMD lane — running the compression across all lanes simultaneously.
-//! This is a different axis of parallelism from block-level `ParBlocks` (which processes
-//! multiple blocks of a *single* stream): here each lane is a distinct message.
+//! An ordinary hash reads one message and produces one digest. The compression function
+//! inside it is serial — block *n* depends on block *n−1* — so a single message cannot be
+//! spread across SIMD lanes. But *different* messages are completely independent of one
+//! another, so several of them can be hashed side by side, one per lane, with the same
+//! instructions driving all lanes at once. That is what this module exposes.
 //!
-//! The message length `N` and batch size `B` are const generics, so equal length and the
-//! one-output-per-message count are enforced by the type system rather than checked at
-//! run time — `&[&[u8; N]; B] -> [Output; B]` cannot mismatch. Messages are passed *by
-//! reference* (`&[u8; N]`), so they need not be contiguous in memory. The lane count is
-//! deliberately *not* exposed: it is the hardware detail (AVX2 vs AVX-512 width) that an
-//! implementation abstracts over via its own runtime dispatch.
+//! This is a different kind of parallelism from
+//! [`ParBlocks`](crate::common::ParBlocks), which processes several blocks of a *single*
+//! stream. Here each lane is a separate message.
+//!
+//! # Two layers, each hiding one detail
+//!
+//! Two unrelated things must be kept apart, and each layer hides one of them from the layer
+//! above:
+//!
+//! 1. **[`MultiUpdateBackend`] hides the implementation.** One backend is one concrete way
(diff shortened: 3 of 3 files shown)


npmccallum:
Thanks for the review.

On one-shot vs incremental: our use case needs incremental. The consumer hashes domain-separated Merkle leaves (`tag ‖ page`, 4 KiB payload). A one-shot multi-message method would force us to concatenate `tag ‖ page` per lane — a 4 KiB copy per message that otherwise never exists — which cancels most of the multi-buffer speedup. Incremental lets us absorb the shared tag block and then borrow the page blocks in place. So this isn't just an early trait for its own sake; it's driven by a zero-copy requirement one-shot can't meet.

On postponing the traits: happy to land this behind an `unstable`-gated feature so the surface stays revisable until more crates adopt it.

On the backend trick: agreed, and I'll align with `cipher`'s closure-based dispatch (`fn call<B: Backend>` + a `with_backend`-style entry point) rather than the inline `cfg_if` selection currently in the sha2 side, so backend selection lives in the trait crate. Note the current design already carries the backend trait and runs a monomorphized generic driver after the runtime check — the closure formalizes and centralizes that.

On `&[&[u8; TAIL]; MSGS]` for finalize: this is deliberate, not incidental. Multi-buffer hashing requires every lane to absorb the same number of bytes, so that equal-length invariant has to be enforced somewhere. The two options are a compile-time guarantee via const generics, or a runtime check — and a runtime check is either a bare panic (low value) or forces every method to become fallible, which discards the infallibility that mirrors `Digest`. The multi-buffer use cases I'm aware of all have message sizes fixed at compile time, so encoding the invariant in the types costs callers nothing in practice while keeping the API zero-copy and infallible.

npmccallum:
@tarcieri your review would be appreciated

tarcieri:
Sorry, have a huge backlog of PRs I'm trying to get through, but I'll take a look at this soon

PR #2483, opened 2026-08-10
digest: erase truncated output temporary

Abdullah1738:
## Summary

- wrap the full variable-output result in a scoped zeroizing buffer when the zeroize feature is enabled
- preserve the existing non-zeroize behavior
- cover ordinary return and unwinding with a serialized erasure-path test

## Rationale

CtOutWrapper copies a truncated view into the caller output, but its full intermediate result previously used ordinary drop behavior. This change gives that intermediate the same zeroization guarantee as the surrounding zeroize-enabled digest state.

## Verification

- cargo test -p digest --all-features
- cargo test -p digest --no-default-features --features block-api,zeroize
- cargo test -p digest --no-default-features --features block-api
- cargo clippy -p digest --all-targets --all-features -- -D warnings
- cargo fmt --all -- --check

### DIFF
--- a/digest/src/block_api/ct_variable.rs
+++ b/digest/src/block_api/ct_variable.rs
@@ -156,14 +156,11 @@ mod tests {
     extern crate std;
 
     #[test]
-    fn scoped_full_result_zeroizes_on_return() {
+    fn scoped_full_result_zeroizes_on_return_and_unwind() {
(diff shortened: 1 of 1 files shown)


newpavlov:
See this [comment](https://github.com/RustCrypto/KEMs/pull/366#issuecomment-5034435499).

newpavlov:
Relevant discussion: https://github.com/RustCrypto/meta/issues/38

PR #2492, opened 2026-09-14
Update reference to `input` (now called `update`) in digest README.md

Fjalar:
confused me when reading on crates.io until I realized it was referencing an older name for the function


newpavlov (approved):
Thanks!

